PR #3, opened 2013-04-18
ui fixes

abenari:
Note that for this patch to work we also need to apply a patch to foreman

``` diff
--- a/app/controllers/hosts_controller.rb
+++ b/app/controllers/hosts_controller.rb
@@ -511,7 +511,7 @@ class HostsController < ApplicationController
   def find_multiple
   # Lets search by name or id and make sure one of them exists first
     if params[:host_names].present? or params[:host_ids].present?
-      @hosts = Host.where("id IN (?) or name IN (?)", params[:host_ids], params[:host_names] )
+      @hosts = Host::Base.where("id IN (?) or name IN (?)", params[:host_ids], params[:host_names] )
       if @hosts.empty?
         error 'No hosts were found with that id or name'
         redirect_to(hosts_path) and return false
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 *.gem
 *.rbc
 .bundle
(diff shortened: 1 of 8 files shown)


GregSutcliffe:
Thanks @abenari !

PR #5, opened 2013-04-30
Add some tests and some final fixes/updates

GregSutcliffe:
More tests to come ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ right now ;)
 Require the gem in Foreman.
 
     bundler.d/Gemfile.local.rb:
+    gem 'foreman_discovery'
+
+If you want to use the very latest code, use a git gem:
+
     gem 'foreman_discovery', :git => "https://github.com/theforeman/foreman_discovery.git"
 
 Update & Restart Foreman:
--- a/README.md
+++ b/README.md
@@ -68,16 +68,8 @@ options) it will look for a DNS record of `foreman`
 
 No configuration of the Foreman UI is required. If you are using Locations and/or Organisations,
 Foreman will default to using the first Location and first Organisation for Discovered
-hosts. If you wish to place them in some other Location/Organization, you can (for now)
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ right now ;)
 Require the gem in Foreman.
 
     bundler.d/Gemfile.local.rb:
-    gem 'foreman_discovery', :git => "https://github.com/GregSutcliffe/foreman_discovery.git"
+    gem 'foreman_discovery', :git => "https://github.com/theforeman/foreman_discovery.git"
```
ohadlevy, line 19: I would prefer the default location would be an official Gem?

app/controllers/discovers_controller.rb
```diff
@@ -13,11 +13,12 @@ class DiscoversController < ::ApplicationController
...
-    # sometimes loses the `where` clause, leading to all Hosts on the Discovery index
-    hosts.reject! {|h| h.type != "Host::Discovered" }
     respond_to do |format|
-      format.html { @hosts = hosts.includes(:model, :location, :organization).paginate :page => params[:page] }
+      format.html do
+        # This `where` shouldn't be necessary, but randomly the sql query on the line above
+        # sometimes loses the `where` clause, leading to all Hosts on the Discovery index
+        @hosts = hosts.where(:type => "Host::Discovered").includes(:model, :location, :organization).paginate :page => params[:page]
+      end
       format.json { render :json => hosts }
```
ohadlevy, line 21: either make sure the json response include the same type, or remove json all together imho

README.md
GregSutcliffe, line 19: I agree, we can release one later today. This is correct for running the 'latest' version, so we'll extend it when the gem goes out.

app/controllers/discovers_controller.rb
GregSutcliffe, line 21: Fix incoming

PR #7, opened 2013-05-01
architecture detection

abenari:


ohadlevy:
I'm not sure if this makes sense, as the arch is always going to be i386 (as thats what the image builds).

ohadlevy:
is there a fact instead that we can use?

abenari:
hardwaremodel

app/models/host/discovered.rb
```diff
@@ -81,6 +83,7 @@ def attributes_to_import_from_facts
   def populateFieldsFromFacts facts = self.facts_hash
     importer = super
     self.subnet = Subnet.subnet_for(importer.ip)
+    self.architecture = Architecture.find_by_name(facts['architecture'])
```
ohadlevy: if at all, you want to use importer.architecture instead of facts[]

abenari: importer.architecture has the following code 

``` ruby
# On solaris architecture fact is harwareisa
name = case os_name
          when /(sunos|solaris)/i
            facts[:hardwareisa]
          else
            facts[:architecture] || facts[:hardwareisa]
          end
 # ensure that we convert debian legacy to standard
 name = "x86_64" if name == "amd64"
 Architecture.find_or_create_by_name name unless name.blank?
```

since facter is running in a tiny-core context, os is known, fact architecture is not relevant and hardwareisa is unknown.
This code is not useful in this case.

ohadlevy:
we might need to introduce a new fact that check if the machine is x86_64 capable, currently facter would report i686 on x86_64 machines regardless (assuming we decide to keep the i686 kernel).

GregSutcliffe:
Sounds good.

PR #9, opened 2013-05-01
moved discovered hosts to the provisioning section of the more menu.

abenari:


GregSutcliffe:
merged in 899424a, thanks @abenari !

PR #11, opened 2013-05-01
replaced multiple buttons to select action style menu.

abenari:
same style as current foreman hosts page has.


GregSutcliffe:
+1 for consistent style - merged as 2f2b89b58bdeb86dbbcf40d380ed8432b5b56c70. thanks!

PR #12, opened 2013-05-01
fixed select/unselect all message.

abenari:


GregSutcliffe:
tested, works, merged. Thanks @abenari !

PR #13, opened 2013-05-02
REVIEW ONLY: Create new Discovered Hosts using a new route

GregSutcliffe:
Don't merge this yet - the change to extra/discover_host will affect current RC users. We'll do a new release of rc3 when we merge this.

### DIFF
--- a/README.build_iso.md
+++ b/README.build_iso.md
@@ -1,0 +1,32 @@
+# Building an ISO
+
+This document assumes you have already run the rake task for building the discovery
+boot image (see main README). As such, you should already have:
+
+    ./discovery_image/vmlinuz
+    ./discovery_image/initrd.gz
+    /tmp/tcl.iso
+
+Now you wish to build an ISO of these files.
+
+# Dependencies
+
+You'll need `genisoimage`
+
+# Procedure
+
+    mkdir newiso
+    sudo mount /tmp/tcl.iso /mnt -oloop
+    cp -r /mnt/boot newiso
+    sudo umount /mnt
+    sudo mv discovery_image/vmlinuz newiso/boot/vmlinuz
+    sudo mv discovery_image/initrd.gz newiso/boot/core.gz
+    sudo genisoimage -quiet -l -J -R \
+        -no-emul-boot -boot-load-size 4 -boot-info-table \
+        -b boot/isolinux/isolinux.bin \
+        -c boot/isolinux/boot.cat \
+        -o "discovery.iso" newiso/
+
+You should now have discovery.iso and can boot it. As there is no way to pass any boot-time
+information to the image, you will need to ensure there is a DNS record for "foreman" that
+the image can resolve.
--- a/app/controllers/discovers_controller.rb
+++ b/app/controllers/discovers_controller.rb
@@ -3,6 +3,9 @@ class DiscoversController < ::ApplicationController
   include Foreman::Controller::TaxonomyMultiple
   unloadable
 
+  # Avoid auth for discovered host creation
+  skip_before_filter :require_login, :require_ssl, :authorize, :verify_authenticity_token, :set_taxonomy, :session_expiry, :update_activity_time, :only => :create
+
   before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert]
   before_filter :find_multiple, :only => [:multiple_destroy, :submit_multiple_destroy]
 
@@ -16,6 +19,26 @@ class DiscoversController < ::ApplicationController
       format.html { @hosts = hosts.list.paginate :page => params[:page] }
       format.json { render :json => hosts }
     end
+  end
+
+  # Importing yaml is restricted to puppetmasters, so instead we take the ip
+  # as a parameter and use refresh_facts to get the rest
+  def create
+    Taxonomy.no_taxonomy_scope do
+      imported = Host::Discovered.new(:ip => get_ip_from_env).refresh_facts
+      respond_to do |format|
+        format.yml {
+          if imported
+            render :text => _("Imported facts"), :status => 200 and return
+          else
+            render :text => _("Failed to import facts"), :status => 400
+          end
+        }
+      end
+    end
+  rescue Exception => e
+    logger.warn "Failed to import facts: #{e}"
+    render :text => _("Failed to import facts: %s") % (e), :status => 400
   end
 
   def show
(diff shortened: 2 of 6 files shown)


app/controllers/discovers_controller.rb
```diff
@@ -3,6 +3,14 @@ class DiscoversController < ::ApplicationController
   include Foreman::Controller::TaxonomyMultiple
   unloadable
 
+  # Avoid auth for discovered host creation
+  skip_before_filter :require_login, :only => :create
+  skip_before_filter :require_ssl, :only => :create
+  skip_before_filter :authorize, :only => :create
+  skip_before_filter :verify_authenticity_token, :only => :create
+  skip_before_filter :set_taxonomy, :only => :create
+  skip_before_filter :session_expiry, :update_activity_time, :only => :create
```
domcleal: You can put the whole lot on one line I think, e.g.

```
skip_before_filter :require_login, :require_ssl, :authorize, :verify_authenticity_token, :set_taxonomy, :session_expiry, :update_activity_time, :only => :create 
```

app/models/host/discovered.rb
```diff
@@ -84,11 +84,11 @@ def populateFieldsFromFacts facts = self.facts_hash
     self.save
   end
 
-  def refresh_facts
+  def refresh_facts ip = self.ip
```
domcleal: This should be a class method I think, rather than creating a new instance first in the controller.

app/controllers/discovers_controller.rb
```diff
@@ -18,6 +26,26 @@ def index
     end
   end
 
+  # Importing yaml is restricted to puppetmasters, so instead we take the ip
+  # as a parameter and use refresh_facts to get the rest
+  def create
+    Taxonomy.no_taxonomy_scope do
+      imported = Host::Discovered.new.refresh_facts ( params.delete('ip') || request.env['REMOTE_ADDR'] )
```
domcleal: You might want to trust X-Forwarded-For too, as the unattended controller does iirc.

app/controllers/discovers_controller.rb
```diff
@@ -18,6 +26,26 @@ def index
     end
   end
 
+  # Importing yaml is restricted to puppetmasters, so instead we take the ip
+  # as a parameter and use refresh_facts to get the rest
```
domcleal, line 25: This seems insecure to me still, as you're querying an arbitrary IP address and then loading the YAML returned.  I think refreshFacts needs securing first (i.e. it should only handle JSON via JSON.load) for this work.

app/models/host/discovered.rb
domcleal: Hmm, maybe instead in the controller do...

```
Host::Discovered.new(:ip => params.delete('ip')).refresh_facts
```

GregSutcliffe:
@domcleal thanks for the ideas. I've done most of them (kept the ip debug ip in refresh_facts, seems useful). As to the security, the facts api from the proxy returns JSON already, and a quick test shows that a phony api returning, say, YAML will cause Foreman to throw an error. Seems ok to you as is?

domcleal:
Agreed, looks good to me security-wise.

ohadlevy:
stupid question, why not fix the image to upload json and then use the same existing route?

GregSutcliffe:
The existing route is covered by the before_filter :require_puppetmaster_or_login, which means foreman has to have the proxy on the discovered host in it's db - which since it's just being discovered, it isn't. This leads to the classic error:

```
No smart proxy server found on [] and is not in trusted_puppetmaster_hosts
```

By using a new route that doesn't use YAML at all, we can relax this filter and allow unknown proxies to tell foreman to pull in the facts.

GregSutcliffe:
Unless anyone objects, I'll merge this on Monday and release rc3 at the same time. The next rc has a fix for build_iso.sh that pulls files from the local gem instead of going to Github, so we won't be blocked merging things to master again, hopefully.

README.build_iso.md
```diff
@@ -0,0 +1,32 @@
+# Building an ISO
```
ohadlevy, line 1: TODO rake task?

extra/build_iso.sh
```diff
@@ -97,7 +100,12 @@ for n in $GEMS ; do gem fetch $n ; done
 # Build the fallback script
 echo "#!/bin/sh" >> ../discovery_init.sh
 for n in $GEMS ; do echo "gem install -l --no-ri --no-rdoc /opt/gems/`ls *$n-[0-9]*.gem`" >> ../discovery_init.sh ; done
-echo "" >> ../bootlocal.sh
+echo "" >> ../discovery_init.sh
+echo 'FACTER_LIB=`gem which facter`' >> ../discovery_init.sh
+echo 'FACTER_PATH=`dirname $FACTER_LIB`' >> ../discovery_init.sh
+echo 'cp /additional_build_files/ipmi.rb $FACTER_PATH/facter/ipmi.rb' >> ../discovery_init.sh
+echo 'cp /additional_build_files/lldp.rb $FACTER_PATH/facter/lldp.rb' >> ../discovery_init.sh
+echo 'cp /additional_build_files/netinfo.rb $FACTER_PATH/facter/netinfo.rb' >> ../discovery_init.sh
```
ohadlevy, line 108: extra line? can we do it in a loop some how?

extra/discover_host
```diff
@@ -45,35 +45,47 @@ def discover_by_dns
   end
 end
 
+def upload
+  puts "#{`date`.chomp}: Triggering import of facts from Foreman"
```
ohadlevy: Time.now ?

extra/discover_host
```diff
@@ -45,35 +45,47 @@ def discover_by_dns
...
+  begin
+    uri = URI.parse(discover_server)
+    http = Net::HTTP.new(uri.host, uri.port)
+    if uri.scheme == 'https' then
+      http.use_ssl = true
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+    end
+    req = Net::HTTP::Post.new("/discovers")
```
ohadlevy, line 57: I never liked that route... i think we should change it.

README.build_iso.md
GregSutcliffe, line 1: Yeah that would be nice. Especially as the script is called build_iso.sh ;)

extra/build_iso.sh
GregSutcliffe, line 108: We definitely need a better way to add custom facts to an image. Adding it to the TODO pile...

extra/discover_host
GregSutcliffe: Doh! Fixing...

extra/discover_host
GregSutcliffe, line 57: As discussed by phone, this only exists to avoid hitting :require_puppetmaster_or_login - once we rework foreman-core to not apply this filter on JSON uploads, we can drop this route (or better, redirect it to /fact_values)

GregSutcliffe:
Ok, merging this monster ahead of rc3 :)

PR #14, opened 2013-05-02
Add discovered host summary to dashboard

GregSutcliffe:
@abenari, how does this look to you?

### DIFF
--- a/app/overrides/add_dashboard_table.rb
+++ b/app/overrides/add_dashboard_table.rb
@@ -1,0 +1,11 @@
+Deface::Override.new(:virtual_path => "dashboard/_status_table",
+                     :name => "add_dashboard_table",
+                     :insert_after => "ul:last",
+                   :text => '
+<h4 class="header"><%= "Host Pool" -%></h4>
+<ul>
+  <li>
+    <%= link_to("Unassigned Hosts", discovers_path, :class=>"dashboard-links" ) %>
+    <h4><%= Host::Discovered.count %></h4>
+  </li>
+</ul>')
--- a/foreman_discovery.gemspec
+++ b/foreman_discovery.gemspec
@@ -23,5 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rails", "~> 3.2.13"
   s.add_dependency "open4"
   s.add_dependency "ftools"
+  s.add_dependency "deface"
+
 
 end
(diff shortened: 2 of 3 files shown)


app/overrides/add_dashboard_table.rb
```diff
@@ -0,0 +1,11 @@
+Deface::Override.new(:virtual_path => "dashboard/_status_table",
+                     :name => "add_dashboard_table",
+                     :insert_after => "ul:last",
+                   :text => '
+<h4 class="header"><%= "Discovery Status" -%></h4>
+<ul>
+  <li>
+    <%= link_to("Discovered Hosts", discovers_path, :class=>"dashboard-links" ) %>
+    <h4><%= Host::Discovered.all.count %></h4>
```
ohadlevy: all.count is not required, just count would perform better (no need to convert to array first)

GregSutcliffe:
@ohadlevy good catch, updated to just use .count

GregSutcliffe:
@abenari ping? Does this seem ok to you?

app/overrides/add_dashboard_table.rb
```diff
@@ -0,0 +1,11 @@
+Deface::Override.new(:virtual_path => "dashboard/_status_table",
+                     :name => "add_dashboard_table",
+                     :insert_after => "ul:last",
+                   :text => '
+<h4 class="header"><%= "Discovery Status" -%></h4>
```
abenari: Can we translate the header?  _("Discovery Status")

app/overrides/add_dashboard_table.rb
```diff
@@ -0,0 +1,11 @@
...
+<ul>
+  <li>
+    <%= link_to("Discovered Hosts", discovers_path, :class=>"dashboard-links" ) %>
```
abenari: Same as above translation.

abenari:
Looks good to me, fits well in the page.

I was thinking of the discovered hosts as a host pool, with unassigned hosts. We could have 'host pool' in the title and 'unassigned hosts' in the link, what do you think?

app/overrides/add_dashboard_table.rb
GregSutcliffe: I don't think translations work in plugins yet - we need to figure out how to get the strings into Transifex. See https://trello.com/card/as-a-foreman-user-i-would-like-plugins-to-be-localized/5162af6bb868365e0c007135/68

app/overrides/add_dashboard_table.rb
GregSutcliffe: See above comment. I agree, I'd like it to be localized ;)

GregSutcliffe:
Merged as 890ef2a. Thanks for the review, guys.

PR #15, opened 2013-05-06
Cache facts and upload to foreman when they change

GregSutcliffe:
Tests ok here.

### DIFF
--- a/extra/discover_host
+++ b/extra/discover_host
@@ -6,8 +6,10 @@ require 'net/https'
 require 'uri'
 require 'yaml'
 
-# Would like to not need this...
+# For comparison
 require 'rubygems'
+require 'facter'
+require 'yaml'
 require 'json'
 
 def discover_server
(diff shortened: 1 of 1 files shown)


extra/discover_host
```diff
@@ -45,35 +47,47 @@ def discover_by_dns
...
-puts "Importing facts"
-begin
-  uri = URI.parse(discover_server)
-  http = Net::HTTP.new(uri.host, uri.port)
-  if uri.scheme == 'https' then
-    http.use_ssl = true
-    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+# loop, but only upload on changes
+while true do
+  uninteresting_facts=/kernel|operatingsystem|osfamily|ruby|path|time|swap|free|filesystem|version/i
```
ohadlevy, line 85: i would also look at selinux etc

extra/discover_host
```diff
@@ -45,35 +47,47 @@ def discover_by_dns
...
+  facts = Facter.to_hash.reject! {|k,v| k =~ uninteresting_facts }
+  unless YAML.load(read_cache) == facts
```
ohadlevy, line 87: hashes are not always sorted, you might need to do full compare or sort it etc

extra/discover_host
GregSutcliffe, line 85: Will add selinux to the main PR in a sec. From my testing, this cache didn't change in ~20 mins.

extra/discover_host
GregSutcliffe, line 87: I tested it, it compares properly when unsorted. From the Hash docs:

Equality—Two hashes are equal if they each contain the same number of keys and if each key-value pair is equal to (according to Object#==) the corresponding elements in the other hash.

GregSutcliffe:
Merging into #13

PR #16, opened 2013-06-26
fixes #2716 error in discovery plugin menu and user without permissions

abenari:
this pull request depend on  https://github.com/theforeman/foreman/pull/750

### DIFF
--- a/lib/discovery_home_helper_patch.rb
+++ b/lib/discovery_home_helper_patch.rb
@@ -3,16 +3,16 @@ module DiscoveryHomeHelperPatch
     base.send(:include, InstanceMethods)
 
     base.class_eval do
-      alias_method_chain :setting_options, :discovers_link
+      alias_method_chain :settings_menu_items, :discovers_link
     end
   end
 
   module InstanceMethods
     # Adds a discovers link to the More menu
-    def setting_options_with_discovers_link
-      choices = setting_options_without_discovers_link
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
Looks good, happy to merge once foreman#750 is in

domcleal:
Should we do this by matching text instead?

We could change Foreman to use N_("Provisioning") and translate the string closer to the page render, then search for the string "Provisioning" from the plugin (ditto for position inside the menu, which [I want to sort](http://projects.theforeman.org/issues/2688) closer to rendering anyway).  It'd be more robust then if we changed the menu between versions.

GregSutcliffe:
@domcleal I agree, but that's a big change, and if we release 1.2 in this state then it will break the UI for non-admin users of a discovery-enabled system. I vote we merge this and review how plugins hook into the menu as soon as possible in the 1.3 cycle.

abenari:
@domcleal I was looking at changing the menu implementation from a structure to class, and maybe using the same code as katello project does and use method  such as insert_after instead of having the plugin know about the internal data structure.
The a larger change might take time until we agree on a common solution with katello project.I don't think we need to postpone this simple fix.

domcleal:
Ok, thanks for the explanation @abenari, that works for me.

GregSutcliffe:
Merged as eab5fa3, thanks @abenari !

PR #18, opened 2013-06-27
Add wget for netinfo.rb 

mattj-io:
Netinfo.rb is referred to in the fallback discovery_init script so needs to be pulled into the base initrd image. New to github so hopefully this is the right way to do this :)

### DIFF
--- a/extra/build_iso.sh
+++ b/extra/build_iso.sh
@@ -46,6 +46,7 @@ ln -s /usr/local/sbin/dmidecode usr/sbin/dmidecode
 mkdir ./additional_build_files
 wget -O ./additional_build_files/ipmi.rb https://github.com/zoide/puppet-ipmi/raw/master/lib/facter/ipmi.rb
 wget -O ./additional_build_files/lldp.rb https://github.com/mfournier/puppet-lldp/raw/master/lib/facter/lldp.rb
+wget -O ./additional_build_files/netinfo.rb https://github.com/ohadlevy/puppet-network/raw/master/lib/facter/netinfo.rb
 if [ -d $LAUNCH_DIR/additional_build_files ]; then
   echo "including files from $LAUNCH_DIR/additional_build_files"
   cp -r $LAUNCH_DIR/additional_build_files .


GregSutcliffe:
Looks great. I might add a "Known Issues" section of the readme, but otherwise perfect. I'll merge this tomorrow, thanks!

README.md
```diff
@@ -75,6 +75,16 @@ specify a DNS record (`foreman.server=myforemanhost`) but in this case the port
 be assumed to be http (80). If all else fails (say, USB boot where we can't provide
 options) it will look for a DNS record of `foreman`
 
+On some multi-core systems the TinyCore kernel may segfault on boot with an error message similar to :
+        
+    "Fixing recursive fault but reboot is needed!"
+
+If this happens, it can be fixed by changing the KERNEL line in the PXE config above to read :
+
+    KERNEL /boot/TinyCore-vmlinuz maxcpus=1
```
ohadlevy: will that affect facter output? it might lead to incorrect inventory value?

extra/build_iso.sh
```diff
@@ -44,6 +44,7 @@ ln -s /usr/local/sbin/dmidecode usr/sbin/dmidecode
 mkdir ./additional_build_files
 wget -O ./additional_build_files/ipmi.rb https://github.com/zoide/puppet-ipmi/raw/master/lib/facter/ipmi.rb
 wget -O ./additional_build_files/lldp.rb https://github.com/mfournier/puppet-lldp/raw/master/lib/facter/lldp.rb
+wget -O ./additional_build_files/netinfo.rb https://github.com/ohadlevy/puppet-network/raw/master/lib/facter/netinfo.rb
```
ohadlevy, line 49: afair, cdpr is not included in the image, so it would just add the gateway ip and gateway interface

mattj-io:
That occurred to me last night and having checked this morning it does affect the initial facts that are returned during discovery, both processorcount and physicalprocesscount are set to 1, which is obviously an issue if you're making provisioning decisions based on processors. Razor have apparently done the same thing, although I'm not sure how/if they work around incorrect facts being returned - https://github.com/puppetlabs/Razor/pull/82. 
Looks like a better fix would be to fix the TC kernel problem to start with.

GregSutcliffe:
I suspect fixing the TC kernel is a much harder problem, and so long as we document the side effects of the fix, I'm happy - it means people with this hardware can use our tools. The longer plan is to move to an RPM-based image anyway, so I'm trying to avoid spending too much time on this image - minor fixes (and any Pull Requests!) are fine, of course ;)

extra/build_iso.sh
GregSutcliffe, line 49: Agreed, but even that's useful info, so I vote we keep it in.

GregSutcliffe:
@dc-mattj if you agree, could you extend the docs to explain the side-effect?

GregSutcliffe:
@dc-mattj It's probably worth opening new Pull Requests rather than adding more things to this one - it'll never get merged at this rate :) Take a look at https://help.github.com/articles/using-pull-requests for tips. Essenitally you want to make one branch per change on your side, and issue a PR from each branch.

Secondly you've managed to catch some of my commits in there, so try doing a rebase followed by a "push -f" to update the PR without the already-committed stuff.

Thanks for the contributions though!

mattj-io:
Sorry, just realised I messed up here. Was just trying to work out how to sort that out !

GregSutcliffe:
@dc-mattj no worries. Feel free to ping 'gwmngilfen' on IRC if you need a hand, that's me ;)

GregSutcliffe:
@dc-mattj ah, you did learn about "push -f", I hadn't noticed this got updated ;)

Merged as ea651dcbaf84ab2be65a57c6635d6a44a06d8eef, thanks! Was there another commit you wanted to get in?

PR #19, opened 2013-06-28
Detect script location and ruby environment better

GregSutcliffe:
Hopefully a fix for #17 - @dc-mattj could you test and confirm? I built the rake task on a 12.04 host with this script and hosts booted with the resulting image seemed to check-in ok, but I'd like confirmation before I merge ;)


mattj-io:
Works for me.

GregSutcliffe:
Great, thanks!

PR #22, opened 2013-07-04
Set local system IP in POST data to Foreman

GregSutcliffe:
This should allow the discovered system to present it's real IP to Foreman when behind a NAT firewall.

@wnkz would you like to test this? You'll need to rebuild the discovery image (or I can build one for you if you like).


wnkz:
Fix works for me ;)

GregSutcliffe:
@wnkz thanks for testing!

PR #28, opened 2013-07-22
Remove explicit Rails dependency, update homepage

domcleal:
This was causing problems for bundler_ext users, since rubygems then tried to install a second copy of rails.


GregSutcliffe:
Looks good, merging

PR #34, opened 2013-10-17
Update to handle the new JSON response from the smart proxy in foreman 1...

asergeant01:
Update to handle the new JSON response from the smart proxy in foreman 1.3

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -22,21 +22,9 @@ class Host::Discovered < ::Host::Base
   scope :list, lambda { where(:type => "Host::Discovered").includes(:model, :location, :organization) }
 
   def self.importHostAndFacts data
-    # data might already be a hash, from refresh_facts
-    facts = data.is_a?(Hash) ? data : YAML::load(data)
-    case facts
-      # Discovered hosts don't have a name yet - use MAC for now
-      # TODO: make this more intelligent, we might not have an eth0...
-      when Puppet::Node::Facts
-        name   = facts.values["macaddress_eth0"].downcase.gsub(/:/,'')
-        values = facts.values
-      when Hash
-        name   = facts["macaddress_eth0"].downcase.gsub(/:/,'')
-        values = facts
-        return raise("invalid facts hash") unless name and values
-      else
-        return raise("Invalid Facts, much be a Puppet::Node::Facts or a Hash")
-    end
+  	facts = data
+	name   = facts["macaddress_eth0"].downcase.gsub(/:/,'')
+	values = facts
 
     # filter facts
     values.reject!{|k,v| k =~ /kernel|operatingsystem|osfamily|ruby|path|time|swap|free|filesystem|version/i }
@@ -64,7 +52,7 @@ class Host::Discovered < ::Host::Base
     end
 
     h.save if h.new_record?
-    h.importFacts(name, values)
+    h.importFacts(values)
   end
 
   def importFacts name, facts


app/models/host/discovered.rb
```diff
@@ -22,21 +22,9 @@ class Host::Discovered < ::Host::Base
   scope :list, lambda { where(:type => "Host::Discovered").includes(:model, :location, :organization) }
 
   def self.importHostAndFacts data
```
GregSutcliffe, line 24: We should probably accept the same input data as the Host::Managed method here (see https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L356) even if we don't actually use it.

app/models/host/discovered.rb
```diff
@@ -64,7 +52,7 @@ def self.importHostAndFacts data
     end
 
     h.save if h.new_record?
-    h.importFacts(name, values)
+    h.importFacts(values)
```
GregSutcliffe, line 55: I don't see how this change can work - the importFacts method 3 lines below requires 2 parameters to be passed to it.

GregSutcliffe:
Thanks! I've put some comments inline, and we should also match the new return state of the core method - as you can see here (https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L373) it now returns both the created host and the save result.

I'll try to test this shortly.

GregSutcliffe:
@asergeant01 Try this commit, it seems to be all I needed to make hosts import again - https://github.com/GregSutcliffe/foreman_discovery/commit/15807f95c24b190966808330682475564c448ddf

asergeant01:
Sorry, I was copying from a virtual machine and copied the wrong version, hence the h.importfacts typo. Will try commit shortly and get back.

app/models/host/discovered.rb
```diff
@@ -22,21 +22,9 @@ class Host::Discovered < ::Host::Base
   scope :list, lambda { where(:type => "Host::Discovered").includes(:model, :location, :organization) }
 
   def self.importHostAndFacts data
-    # data might already be a hash, from refresh_facts
-    facts = data.is_a?(Hash) ? data : YAML::load(data)
-    case facts
-      # Discovered hosts don't have a name yet - use MAC for now
-      # TODO: make this more intelligent, we might not have an eth0...
-      when Puppet::Node::Facts
```
asergeant01, line 30: @GregSutcliffe - This line causes an error "Failed to import facts: uninitialized constant Host::Discovered::Puppet" if left in. This is because Puppet has been removed from facts - https://github.com/theforeman/foreman/commit/01055e77a79e3f0b229a9955ff943514c82e7ca0

asergeant01:
It didn't work for me because the Puppet::Node::Facts was still in your commit. The data being received from the smart-proxy is json.

GregSutcliffe:
It's got nothing to do with the data coming from the proxy, otherwise it wouldn't work for me either :)

I think I know what's going on - without the Puppet libraries it can't figure out what Puppet::Node::Facts object is, so therefore it cannot possibly compare the incoming data to one. My local Foreman instance must still be loading Puppet somehow, otherwise I'd be seeing this too.

I'll try to replicate and ensure it's definitely fixed today.

GregSutcliffe:
Ok, replicated and fixed, I think. Try this one: https://github.com/GregSutcliffe/foreman_discovery/commit/7ca1cb56c80d4816f362108112db0ee4a7572f4f

I've updated the method to be a little clearer in the case of errors, and by slightly more fault tolerant too.

asergeant01:
Works! :+1: When will this be merged and available to get via the bundle update method?

GregSutcliffe:
I'll try and get a new gem out today. Thanks for testing!

GregSutcliffe:
@asergeant01 1.1.0 released - let me know how you get on, and thanks for the pointers :P

asergeant01:
1.1.0 appears to have been yanked from rubygems - https://rubygems.org/gems/foreman_discovery/versions

GregSutcliffe:
Apologies for that, I built a bad gem and then got distracted :)

PR #38, opened 2013-11-07
Do not filter out discovered_version fact

lzap:
plus one small improvement


lzap:
Ok fixed, fyi the fact I want to be there is:

```
  discovery_version 0.1.0
```

It pulls facter version as well, which is hopefully not an issue. Thanks.

GregSutcliffe:
Merged as 45fd6ea thanks lzap!

PR #44, opened 2013-12-14
Update README with ISC DHCP configuration

antaflos:
Describe ISC DHCP server configuration needed for making discovery work.
Also re-add section on PXE default configuration which seems to have
gotten lost some time in the past two months.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,13 @@
 # foreman\_discovery
 
-This plugin enables MaaS hardware discovery in Foreman.
+This plugin enables MaaS hardware discovery in Foreman. The machine boots from
+the network via PXE and loads a special image based on [Tiny Core Linux](http://distro.ibiblio.org/tinycorelinux/) that
+talks to Foreman and registers the machine as a Discovered Host. Once the
+machine is discovered it can be set up and provisioned just like regular hosts
+managed by Foreman.
+
+Please see the [Foreman Discovery Plugin Walkthrough video](http://www.youtube.com/watch?v=52K7gYAR84k)
+on YouTube for an introduction and installation walkthrough. 
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,13 @@
 # foreman\_discovery
 
-This plugin enables MaaS hardware discovery in Foreman.
+This plugin enables MaaS hardware discovery in Foreman. The machine boots from
+the network via PXE and loads a special image based on [Tiny Core Linux](http://distro.ibiblio.org/tinycorelinux/) that
```
GregSutcliffe, line 4: While this is absolutely correct, we also have an RPM image availabel too, might be worth mentioning

README.md
```diff
@@ -36,21 +43,85 @@ There are two options at the moment:
 The TCL image has the same version as the Foreman Discovery plugin, the oVirt
 Node image has different versioning scheme.
 
+# Preparing the DHCP server 
+
+A prerequisite is that the network's DHCP server can serve unknown clients and
+that it points those clients to the TFTP server from which they will load the
+discovery boot image. This is different from regular Foreman operation, where
```
GregSutcliffe, line 50: Not strictly true. 1) Foreman still creates leases for known hosts, and 2) there might well be unknown clients, but Foreman won't interact of with them ofc :)

I'd suggest something like "This is different from regular Foreman operation, where clients unkown to Foreman are unaffected"

README.md
```diff
@@ -36,21 +43,85 @@ There are two options at the moment:
...
 
 ## UI config
 
 No configuration of the Foreman UI is required. If you are using Locations and/or Organisations,
 Foreman will default to using the first Location and first Organisation for Discovered
-hosts. If you wish to place them in some other Location/Organization, you can alter the
+Hosts. If you wish to place them in some other Location/Organization, you can alter the
 default Loc/Org in `More->Settings->Discovery Settings`
 
+## PXE configuation (PXE default file)
```
GregSutcliffe, line 86: This doesn't need to be here - PXE config is detailed in the TCL page (and the RPM/ovirt page) since they have different boot instructions. You could add a short sentence pointing pople to the relevant document above though (i.e https://github.com/theforeman/foreman_discovery#building-or-downloading-an-image)

GregSutcliffe:
Sorry for the delay getting back to you - thanks for the contribution! Just a couple of small points, but it's mostly awesome :)

antaflos:
So sorry, completely forgot about this. Will review if my changes are still relevant and update according to your comments, @GregSutcliffe.

lzap:
Hello, we have updated our README in the meantime. This needs rebasing. Also note that TCL image is no longer supported, our new image lives here: https://github.com/theforeman/foreman-discovery-image

I am closing this PR now, feel free to retest on our new image and apply changes. Reopen any time. Thank you.

PR #46, opened 2013-12-20
Fixes #3921 - Make choice of discovery fact name a setting

GregSutcliffe:
@lzap thoughts? @mmol care to test drive? :)

### DIFF
--- a/extra/ovirt-node-plugin-foreman/README.md
+++ b/extra/ovirt-node-plugin-foreman/README.md
@@ -24,14 +24,21 @@ page](https://github.com/theforeman/foreman_discovery).
 Injecting into existing base image
 ----------------------------------
 
-At the moment, it is not possible to use existing oVirt Node ISO base files
-and insert discovery plugin RPM into them as version 3.0.1 (current stable) is
-not compatible.
+Warning: This process will set SELinux to permissive mode during the build of
+new iso image. Since this can compromise security, it is highly recommended to
+use dedicated VM guest for building images.
 
-Once oVirt Node 3.1.0 is out, we will provide discovery node plugin RPM and
-image can be easily modified using ovirt-node tools. We are not there yet, but
-the images will be published on the [oVirt Node download
-page](http://www.ovirt.org/Node).
+Although it is not possible to use stable version (3.0) for a base image,
+there are couple of [nightly
+builds](http://resources.ovirt.org/releases/node-base/nightly/iso/) that can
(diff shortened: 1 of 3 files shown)


mmoll:
works for me :)

app/models/setting/discovered.rb
```diff
@@ -6,6 +6,12 @@ def self.load_defaults
     # Check the table exists
     return unless super
 
+    Setting.transaction do
+      [
+        self.set('discovery_fact', _("The default fact name to use for the MAC of the system"), "macaddress"),
```
lzap, line 11: So we are changing from macaddress_eth0 to macaddress? Just to confirm...

lzap:
WFM

app/models/setting/discovered.rb
GregSutcliffe, line 11: We're changing the default to "macaddress", to fit with the new RPM images. I'll merge this and do a new gem release once we feel the RPM's are ready for prime-time :)

lzap, line 11: Ok, except the extra 500 error when rebooting (fix is going soon) I think we are fine. We need more testing from the community tho.

GregSutcliffe, line 11: Great, we can move the RPM image to the primary method of using Discovery for 1.4.

lzap, line 11: Right, I was thinking about versioning scheme. Do we want to follow
Foreman core, or keep very own version (different to TCL)? Looks like a
mess now.

LZ

On Tue, Jan 07, 2014 at 03:19:17AM -0800, Greg Sutcliffe wrote:

> > @@ -6,6 +6,12 @@ def self.load_defaults
> >      # Check the table exists
> >      return unless super
> > -    Setting.transaction do
> > -      [
> > -        self.set('discovery_fact', _("The default fact name to use for the MAC of the system"), "macaddress"),
> 
> Great, we can move the RPM image to the primary method of using Discovery for 1.4.
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/theforeman/foreman_discovery/pull/46/files#r8690706

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

GregSutcliffe, line 11: I tend to release a new version of the plugin with each core release, as usually things have changed. I think we can use our own version numbers in the plugin though.

GregSutcliffe:
Merged, thanks for the reviews guys

PR #50, opened 2014-01-24
fixes #4174 - fixed locale distribution and jenkins rake task

lzap:
Two issues from RC1 resolved plus i18n support is fully working now. Although
i18n plugin support does not support javascript and model localization, we do
not use these in Discovery.

Also, there is a PR coming to Foreman with minor changes in the Makefile script
to better support plugin extraction.

Packagers note: Please do this command during package build to prepare MO
files:

```
make -C locale all-mo
```

That should create locale/*/LC_MESSAGES/discovery.mo files.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,0 +1,8 @@
+[main]
+host = https://www.transifex.com
+
+[foreman.discovery]
+file_filter = locale/<lang>/discovery.po
+source_file = locale/discovery.pot
+source_lang = en
+type = PO
--- a/foreman_discovery.gemspec
+++ b/foreman_discovery.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
     "LICENSE",
     "README.md"
   ]
-  s.files = Dir["{app,extra,config,db,lib}/**/*"] + ["LICENSE", "Rakefile", "README.md"]
+  s.files = Dir["{app,extra,config,db,lib,locale}/**/*"] + ["LICENSE", "Rakefile", "README.md"]
   s.test_files = Dir["test/**/*"]
   s.homepage = %q{http://github.com/theforeman/foreman_discovery}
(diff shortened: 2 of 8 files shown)


GregSutcliffe:
@lzap so that needs to go in the debian package for the discovery plugin?

@domcleal looks ok to me but I know little about i18n, could you ack?

domcleal:
:+1: all present and correct.

I guess you'd need to run the make before building the gem and distribute .mo files with it?

lzap:
Exactly, just run the above command in your "SPEC" file, I guess it's "rules" file.

I'll go ahead and modify our RPM file for rc2 once it is out. @domcleal not releasing workarounded RPM as Greg plans to release rc2 quite soon I guess.

domcleal:
Debian uses the packaged gems, and surely other users who apply just the rubygem will need the .mo files?  So it seems to me they should packaged in the .gem, not built in the RPM or Debian package - because the .gem is also an end user binary package.

GregSutcliffe:
@domcleal I think I agree, it needs to be in the gem. @lzap care to add that to this PR? Or shall I figure out where that needs to go? :)

lzap:
Hmmm I say yes to MO files in the gem file, but NO to MO files in the
git :-)

Can you just create a ./release.sh with the make file from above and
"gem release discovery.gemspec"?

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

GregSutcliffe:
I agree, we  don't want them in git, just in the gem.

lzap:
FYI its already in gitignore, so you really only need to generate them
and include them. I guess it will be included automatically as the whole
locale/*_/_ is there.

Really just the two commands, go ahead!

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #51, opened 2014-01-28
Some fixes for the Tiny Core Image

eesprit:
Hi,

2 little patches to :
- Replace the sleep (for network) by a script which actually check the network (20 seconds was not enough in my case)
- Explicitly fetch the 1.3.4 version of the tilt gem because the one that was fetched by default (2.0.0) was breaking the sinatra gem.

The network check script will try to ping the foreman server (based on the kernel boot command line) every second for 3 minutes, perhaps not the better way to handle that (a better failure system if network is still down ?) but still better than the basic "sleep 20".

Regards,

### DIFF
--- a/extra/build_iso.sh
+++ b/extra/build_iso.sh
@@ -88,7 +88,7 @@ fi
 
 # Build the init script
 echo "" >> opt/bootlocal.sh
-echo "sleep 20" >> opt/bootlocal.sh # network can be slow to come up, and we're not in a rush
+echo "/opt/wait_for_network.sh" >> opt/bootlocal.sh # network can be slow to come up
 echo "/opt/foreman_startup.rb" >> opt/bootlocal.sh
 echo "/opt/discovery_init.sh" >> opt/bootlocal.sh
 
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
This looks ok to me, I'll try to test it next week. @lzap thoughts?

lzap:
Hmmm this will break I guess when no parameter is passed (and "foreman" default value is used). If you can just set server to "foreman" by default.

I'd like to see some maximum limit of retries, because what happens when foreman does not respond to pings? I know this is against rules, but I've seen many enterprises not doing that because their security dept said not to do it :-)

Minor issues tho +1

eesprit:
You're right for the first point, should have added a server="foreman" to have a default value.

For the second point, there is already a maximum limit of retries (180 => 3 min) or I misunderstood your comment.
Perhaps a better way to handle the ICMP blocking problem is to try (for a maximum number of retries) to resolve the DNS name, and when its done (or if we already had an IP as parameter) just checking the availabilty of port 80/443 (for a given time) ?

lzap:
> For the second point, there is already a maximum limit of retries (180
> => 3 min) or I misunderstood your comment.

Oh right I see it now.

> Perhaps a better way to handle the ICMP blocking problem is to try
> (for a maximum number of retries) to resolve the DNS name, and when
> its done (or if we already had an IP as parameter) just checking the
> availabilty of port 80/443 (for a given time) ?

I think this is okay, after maximum time we will continue so even if
ICMP is blocked, this will work, but slower. I am fine with this.

+1

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

GregSutcliffe:
Do we want that wait-for-network script in the RPM too @lzap?

lzap:
Oh no, we have systemd there :-) (And sysvinit in RHEL6 image that has dependencies - this has not been implemented yet).

GregSutcliffe:
Ok, awesome, I'll get this merged later today then, unless you have other concerns?

GregSutcliffe:
Squashed and merged as ce0e93621c3ee488fcb9643b8e0112c565747cda, thanks @eesprit! Sorry for the delay.

PR #52, opened 2014-01-30
longer fact values are truncated in host detail

lzap:
This was bugging me all the time, fixing this for my talk. Long fact values
(like ssh public keys) are long and ruining the screen. I am putting 100 as a
sane default there.


GregSutcliffe:
Thanks @lzap works nicely. Merged into the 1.2.0 release.

PR #53, opened 2014-02-10
discovery taxonomies and interactive hostname

tomnowak:
patch to allow taxonomies to be assigned in PXE menu or predetermined in discovery_init.sh and to allow interactive host name to be provided for discovered hosts

### DIFF
--- a/extra/discovery_init.sh.example
+++ b/extra/discovery_init.sh.example
@@ -5,6 +5,27 @@ echo "Starting custom discovery script"
 GEMS="facter json_pure rack rack-protection tilt sinatra"
 for n in $GEMS ; do gem install --no-ri --no-rdoc -l `ls /opt/gems/*$n-[0-9]*.gem` ; done
 
+set -- `cat /proc/cmdline`
+for I in $*; do case "$I" in *=*) eval $I;; esac; done
+
+mkdir -p /etc/facter/facts.d
+if [ -n "$discovery_organization" ]; then
+  echo discovery_organization=$discovery_organization >/etc/facter/facts.d/discovery.txt
+fi
+
+if [ "$discovery_mode" == 'interactive' ]; then
+  chvt 2
+  (
+    echo Foreman interactive mode activated...
+    echo
+    read -p "Hostname: " discovery_hostname
+    if [ -n "$discovery_hostname" ]; then
+      echo discovery_hostname=$discovery_hostname >>/etc/facter/facts.d/discovery.txt
+    fi
+  )  </dev/tty2 >/dev/tty2 2>&1
+  chvt 1
+fi
+
 /usr/share/foreman-proxy/bin/smart-proxy
 /usr/share/foreman-proxy/bin/discover_host
 
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -24,8 +24,9 @@ class Host::Discovered < ::Host::Base
   def self.importHostAndFacts facts
     raise(::Foreman::Exception.new("Invalid Facts, must be a Hash")) unless facts.is_a?(Hash)
     fact_name = Setting[:discovery_fact] || 'macaddress'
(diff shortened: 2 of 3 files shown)


extra/discovery_init.sh.example
```diff
@@ -5,6 +5,27 @@ echo "Starting custom discovery script"
 GEMS="facter json_pure rack rack-protection tilt sinatra"
 for n in $GEMS ; do gem install --no-ri --no-rdoc -l `ls /opt/gems/*$n-[0-9]*.gem` ; done
 
+set -- `cat /proc/cmdline`
```
GregSutcliffe, line 8: Whilst I see where you're going with this, we already have stuff that parses /proc/cmdline in discover_host.rb - it would probably make more sense to add this code there, so that we can ensure it's done before starting to upload data to Foreman.

extra/discovery_init.sh.example
```diff
@@ -5,6 +5,27 @@ echo "Starting custom discovery script"
...
+for I in $*; do case "$I" in *=*) eval $I;; esac; done
+
+mkdir -p /etc/facter/facts.d
+if [ -n "$discovery_organization" ]; then
+  echo discovery_organization=$discovery_organization >/etc/facter/facts.d/discovery.txt
+fi
+
+if [ "$discovery_mode" == 'interactive' ]; then
```
GregSutcliffe, line 16: I'm not sure I want to encourage interactive elements in the discovery images - they're not meant for that. Do you have a particular need for this functionality?

GregSutcliffe:
sorry for the delay @tomnowak - just catching up on my reviews. Looks fairly sensible, just a few comments.

@lzap any thought from your side?

lzap:
Yeah, I was waiting for your opinion. While I think setting up fact name, org and location is great addition, I am also not sure about the interactivity. Discovery is not meant for this, but on the other hand if the use case will be described, then I have no problem with that - if this is optional and turned off by default.

One important note - this will only work in TCL image, we have also oVirt Fedora RHEL based images which reside in extra/ovirt-node-foreman-plugin directory. The oVirt images are likely those who will be maintained by us in the future. If you only want the change in TCL image, I am fine with that. If you are willing to make the changes for oVirt image, I am happy to test it.

tomnowak:
Thank you for your feedback.
As for the interactive mode. I'm happy to not include that in main repo, as the discovery_init.sh script I use on site will have more customization anyway and it is kept outside discovery plugin (public folder).
I however use that functionality on site for some specific builds. Some types of machines I build have host name written as a label attached to then . It is much more convenient for me to provide host name when building machine via discovery interactive dialog, so I don't have to know the MAC address of network interface (some laptops have MAC labels, some desktops don't so I have to go to BIOS first). So I have to write MAC address down on the paper along with the host name that is attached to the machine. I also build multiple of these machines in parallel, so it is much easier to assign a proper host groups on web interface based on the name of the machine provided in interactive discovery mode.

lzap:
This looks like a valid case for this feature.

If Greg approves, I'd like to ask you for TCL documentation (wiki)
enhancement as well.

Thanks

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lzap:
Ping, Greg what is the status now?

GregSutcliffe:
@lzap sorry, this slipped the net. The PR needs a rebase at the very least, and as we're about to drop the TCL image, we need to move the image-side code to the new repo.

lzap:
Tom, we are dropping TCL-based image in favor of Red Hat based image:

Are you interested in trying the new one out and maybe rebasing your change there? It has slightly different architecture tho.

https://github.com/theforeman/ovirt-node-plugin-foreman

lzap:
Tom, TCL image is history. I am closing this PR now, feel free to rebase on top of our new image which is here: https://github.com/theforeman/foreman-discovery-image

PR #61, opened 2014-04-07
Api

GregSutcliffe:
Adds a basic api. Needs a bit of tweaking to get the root node right in all cases, and you can't refresh_facts yet, but CRUD works (use PUT/update for provisioning a node).

Side note, also handles the renamed methods in core.

### DIFF
--- a/app/controllers/api/discovers_controller.rb
+++ b/app/controllers/api/discovers_controller.rb
@@ -1,0 +1,114 @@
+module Api
+  class DiscoversController < ::Api::V2::BaseController
+
+    before_filter :find_resource, :except => %w{index create facts}
+    skip_before_filter :authorize, :only => :facts
+
+    api :GET, "/discovers/", "List all discovered hosts"
+    param :search, String, :desc => "Filter results"
+    param :order, String, :desc => "Sort results"
+    param :page, String, :desc => "paginate results"
+    param :per_page, String, :desc => "number of entries per request"
+
+    def index
+      @discovers = resource_scope.search_for(*search_options).paginate(paginate_options)
+    end
+
+    api :GET, "/discovers/:id/", "Show a discovered host"
+    param :id, :identifier_dottable, :required => true
+
+    def show
+    end
+
+    def_param_group :discovered_host do
+      param :discovered_host, Hash, :action_aware => true do
+        param :name, String, :required => true
+        param :ip, String, :required => true
+        param :mac, String, :required => true
+      end
+    end
+
+    api :POST, "/discovers/", "Create a discovered host"
+    param_group :discovered_host, :as => :create
+
+    def create
+      @discovered_host = Host::Discovered.new(params[:discovered_host])
+      process_response @discovered_host.save
+    end
+
+    api :PUT, "/discovers/:id/", "Provision a discovered host"
+    param :id, :identifier, :required => true
+    param :discovered_host, Hash, :action_aware => true do
+      param :name, String
+      param :environment_id, String
+      param :ip, String, :desc => "not required if using a subnet with dhcp proxy"
+      param :mac, String, :desc => "not required if its a virtual machine"
+      param :architecture_id, :number
+      param :domain_id, :number
+      param :puppet_proxy_id, :number
+      param :puppet_class_ids, Array
+      param :operatingsystem_id, String
+      param :medium_id, :number
+      param :ptable_id, :number
+      param :subnet_id, :number
+      param :compute_resource_id, :number
+      param :sp_subnet_id, :number
+      param :model_id, :number
+      param :hostgroup_id, :number
+      param :owner_id, :number
+      param :puppet_ca_proxy_id, :number
+      param :image_id, :number
+      param :host_parameters_attributes, Array
+      param :build, :bool
+      param :enabled, :bool
+      param :provision_method, String
+      param :managed, :bool
+      param :progress_report_id, String, :desc => 'UUID to track orchestration tasks status, GET /api/orchestration/:UUID/tasks'
+      param :capabilities, String
+      param :compute_attributes, Hash do
+      end
+    end
+
+    def update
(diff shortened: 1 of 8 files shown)


app/controllers/api/discovers_controller.rb
```diff
@@ -0,0 +1,114 @@
+module Api
+  class DiscoversController < ::Api::V2::BaseController
+
+    before_filter :find_resource, :except => %w{index create facts}
+    skip_before_filter :authorize, :only => :facts
+
+    api :GET, "/discovers/", "List all discovered hosts"
```
ohadlevy, line 7: does /discovery (singular) makes more sense? I never liked discovers, as it doesn't sound like a tech term ;)

app/controllers/api/discovers_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    api :GET, "/discovers/:id/", "Show a discovered host"
+    param :id, :identifier_dottable, :required => true
+
+    def show
+    end
+
+    def_param_group :discovered_host do
+      param :discovered_host, Hash, :action_aware => true do
+        param :name, String, :required => true
```
ohadlevy, line 25: do we require a name?

app/controllers/api/discovers_controller.rb
```diff
@@ -0,0 +1,114 @@
...
+    def create
+      @discovered_host = Host::Discovered.new(params[:discovered_host])
+      process_response @discovered_host.save
+    end
+
+    api :PUT, "/discovers/:id/", "Provision a discovered host"
+    param :id, :identifier, :required => true
+    param :discovered_host, Hash, :action_aware => true do
+      param :name, String
+      param :environment_id, String
```
ohadlevy, line 43: most of these attributes are not required? do we really support PUT/update for discovered host?

ohadlevy, line 43: ah, I see what you mean, scratch that.

app/views/api/discovers/main.json.rabl
```diff
@@ -0,0 +1,20 @@
+object @discovered_host
+
+extends "api/discovers/base"
+
+attributes :ip, :environment_id, :environment_name, :last_report, :mac,
+           :sp_mac, :sp_ip, :sp_name, :domain_id, :domain_name, :architecture_id, :architecture_name, :operatingsystem_id, :operatingsystem_name,
```
ohadlevy, line 6: I dont think we should display those?

config/routes.rb
```diff
@@ -23,4 +23,15 @@
     end
   end
 
+  ## API
+  namespace :api, :defaults => {:format => 'json'} do
+    scope "(:apiv)", :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2) do
+      resources :discovers do
```
ohadlevy, line 29: I would change it to a nested route under hosts, e.g.

```
/api/hosts/discovery
```

app/controllers/api/discovers_controller.rb
GregSutcliffe, line 7: I had issues with using a singluar as an STI of a plural back when I first wrote discovery, hence why we have "discovers" today. If I'd have understood Rails better back then I'd have used "discoveries" which is the correct plural.

I'm happy with a singular instead of a plural, but I'd want to change everything to be consistent (in the UI, etc) so that should be a second PR after this one, I think.

app/controllers/api/discovers_controller.rb
GregSutcliffe, line 25: Good catch, we don't.

app/views/api/discovers/main.json.rabl
GregSutcliffe, line 6: I was in two minds about it myself. I guess it's not required for anything other than PUT/update (which will create a real host).

Ideally I'd like to extend the hosts rabl so that we're always consistent, but I had issues with that (either failing to find the extensions or having incorrect attributes). I'll take these out for now and we can revisit the hosts inheritance later.

pitr-ch:
@GregSutcliffe hi when do you think this could be merged? I need this for staypuft.

GregSutcliffe:
@pitr-ch I've merged the fix for 4373 separately and dropped it from this PR - thanks for the ack.

pitr-ch:
Thanks for quick fix.

lzap:
Greg, how far are you with Ohad's remarks? I am ready to test the PR now.

config/routes.rb
GregSutcliffe, line 29: This can't be done, as far as I can see. The existing route in core `GET /api/{v1|v2}/hosts/:id` will intercept this first, and attempt to look up a Host::Managed called "discovers" and fail:

```
Host::Managed Load (0.2ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."type" IN ('Host::Managed') AND "hosts"."name" = 'discovers' LIMIT 1
```

We can't change or override a core route from a plugin, routes are read from the top down. So I don't see how to do this - it'll have to stay as /api/v2/discovers for now.

app/views/api/discovers/main.json.rabl
GregSutcliffe, line 6: Fixed

GregSutcliffe:
@lzap updated.

lzap:
Ok I am ready to test this one out with patched discovery. on that.

lzap:
I am able to confirm this works fine. I have build new disovery nightly image 0.5.0 that works _both_ with V2 and the legacy API. We need to hold it for couple of releases because of the OFI tiger project (they have some builds with legacy api). There is a FIXME line there so we can remove that code later.

+1 if there are no objections I am going to merge that @ohadlevy ?

GregSutcliffe:
Merged as f1b03f6 - thanks for testing @lzap

PR #68, opened 2014-04-25
Fixes #5459 - Add zanata.xml file for RH translations

bkearney:


lzap:
As always, let's ask our translation team to poke us with updating translations from upstream just before they start working on it. +1

bkearney:
I do not have commit.. would you mind merging?

lzap:
Sure thing, merged as 7742b6a thanks!

PR #69, opened 2014-05-13
Removing legacy foreman plugin, updating readme

lzap:
Do you want me to remove ISO building bits?


README.md
```diff
@@ -35,13 +36,12 @@ github.com issues.
...
+[ovirt-node-plugin-foreman](https://github.com/theforeman/ovirt-node-plugin-foreman)
+page with further information.
 
-* [Build or download Tiny Core Linux based image](README.tcl_image.md)
-* [Build or download oVirt Node based image](README.ovirt_image.md)
-
-The TCL image has the same version as the Foreman Discovery plugin, the oVirt
-Node image has different versioning scheme.
+On the page, you will find more info how to configure Foreman templates as
+well.
```
GregSutcliffe, line 44: Can we keep the template instructions in this document? There's no need to force people using our images to visit the ovirt repo _just_ to configure their templates

GregSutcliffe:
@lzap sorry for the delay - one comment, but otherwise looks good.

README.md
lzap, line 44: > Can we keep the template instructions in this document? There's no need to force people using our images to visit the ovirt repo _just_ to configure their templates

I understand, but isn't hyperlinks the way world wide web works? :-)

I mean, then we need to make sure both instructions are same. And I
usually forget. Let's have it in the single place.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

GregSutcliffe, line 44: > I understand, but isn't hyperlinks the way world wide web works? :-)

Maybe, but it's not how humans work. I guarantee people will miss this step in the configuration if you leave it the way it is now.

> I mean, then we need to make sure both instructions are same. And I usually forget. Let's 
> have it in the single place.

+1 for a single place, so long as it's in this repo :)

GregSutcliffe:
So after seeing a user yesterday who had missed all the docs except the toplevel one, I think we needs to:

1) Have a single readme, with a link to the download location of the recommended image
2) The single readme should contain the PXE setup instructions
3) The single readme should contain a link to the ovirt repo for instructions on building the image (this is image specific and an advanced topic, I'm fine with that in another place)

The PXE instructions are unlikely to change significantly, and are critical to the functioning of the Discovery plugin. It's good user experience to have those instructions up front, and not much overhead for us.

GregSutcliffe:
Merged as dfaf6fe, many thanks for the initial draft @lzap

PR #70, opened 2014-05-15
API improvements

GregSutcliffe:
Still need to re-write the main controller name though :P


lzap:
Taking that for review. I need to solve out the wrong form first. My priority next week.

README.api.md
```diff
@@ -0,0 +1,84 @@
+# foreman\_discovery api
+
+The most up-to-date API information should be documented in the api cache on your Foreman server. However, this may server as a useful reference.
```
bronhaim: "may serve"

README.api.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+{
+  "discover": {
+    "name": "m52540007aa4d",
+    "mac": "11:22:33:44:55:66",
+    "ip": "192.168.124.50",
+    "hostgroup_id": "6",
+    "root_pass": "12345678",
+    "subnet_id": "2"
```
bronhaim, line 77: I would put in the example also how to pass host_parameters_attributes

app/controllers/api/v2/discovers_controller.rb
```diff
@@ -0,0 +1,115 @@
...
+        param :mac, String, :desc => "not required if its a virtual machine"
+        param :architecture_id, :number
+        param :domain_id, :number
+        param :puppet_proxy_id, :number
+        param :puppet_class_ids, Array
+        param :operatingsystem_id, String
+        param :medium_id, :number
+        param :ptable_id, :number
+        param :subnet_id, :number
+        param :compute_resource_id, :number
```
bronhaim: why doesn't compute resource relation exist in the UI when picking provision parameters but here it does?

app/controllers/api/discovers_controller.rb
```diff
@@ -1,114 +0,0 @@
...
-      param :model_id, :number
-      param :hostgroup_id, :number
-      param :owner_id, :number
-      param :puppet_ca_proxy_id, :number
-      param :image_id, :number
-      param :host_parameters_attributes, Array
-      param :build, :bool
-      param :enabled, :bool
-      param :provision_method, String
-      param :managed, :bool
```
bronhaim, line 65: can you add desc on "managed" "build" "enabled" "provision_method" "model_id" ?

app/controllers/api/v2/discovers_controller.rb
lzap: Not relevant here for bare metal, it should be hidden, yeah.

README.api.md
```diff
@@ -0,0 +1,84 @@
...
+a 'discover' root node, such as:
+
+```
+{
+  "discover": {
+    "name": "m52540007aa4d",
+    "mac": "11:22:33:44:55:66",
+    "ip": "192.168.124.50"
+  }
+}
```
lzap, line 61: We should mention that hostname must not start with digits here. Maybe more than that, we should use a different default for newly provisioned discovered hosts.

bronhaim:
adding issue url for followup: http://projects.theforeman.org/issues/5781

lzap:
I am able to confirm the patch works fine here, using hg, no hg, mixed hg + own params.

bronhaim:
works for me. BUT, each wrong parameter returns same error - "Name must not include periods" which is wrong and hard to find the real issue with the request

app/controllers/api/v2/discovers_controller.rb
GregSutcliffe: Agreed. `compute_resource_id` and `compute_attributes` will be removed.

README.api.md
GregSutcliffe, line 61: Agreed, i'll alter the default and add a description update.

GregSutcliffe:
@lzap i finally got around to renaming the controller, so this PR just got a lot bigger - worth the effort though. Could you review?

@bronhaim could you retest using 'discovered_host' as your root node? I'm not getting any duplicates now. The error you're seeing on wrong parameters is coming from Foreman I think, so I don't think we can fix that in this PR. I'll try to test further.

GregSutcliffe:
@lzap @bronhaim note https://github.com/theforeman/ovirt-node-plugin-foreman/issues/3 is required for host discovery to work, otherwise the image will post to the wrong api url. I'll build a test image now and post the URL to it here.

bronhaim:
@GregSutcliffe Why is ovirt-node-plugin-foreman needed here? I didn't bundle it and it works.  You're right about foreman's output, I'll check that more deeply and maybe I'll post fix for that later.
I'll test with discovered_host and update. thanks

GregSutcliffe:
Update: http://yum.theforeman.org/discovery/gwmngilfen/ - these images work for me with this PR. Host registers via the new api, and can be provisioned.

GregSutcliffe:
@bronhaim ovirt-node-plugin-foreman is the repo used to build the discovery image itself. That needs an update for the new api endpoint - use the images I linked above.

bronhaim:
@GregSutcliffe provision works well with api/v2/discovered_hosts/:id as described in README.api.md. thanks. still I don't understand the meaning and use of some optional fields in the request, such as "build", "enabled", "managed", "capabilities", "provision_method", "image_id", "model_id". why not adding desc for those?

GregSutcliffe:
They come from here: https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L31-L60

Bear in mind we're converting a discovered host to a real host, so we need to support pretty much everything a real host supports. The likelihood of you wanting to set any of the things you've listed is small, but possible. The vast majority of users can ignore them and the defaults will be fine.

For the record, I only PUT name, mac, ip, hostgroup, subnet, and root password. Given all the other 'defaults' set in my hostgroup (OS, environment, etc) that provisions just fine.

I can add some description anyway, but you're unlikely to want to set them.

lzap:
Pushed as
- e5d0735 Fixes #5589 - Add a prefix to hostnames to avoid hostnames beginning with a digit <Greg Sutcliffe>
- 9a827b4 Fixes #5781 - Change controller name to DiscoveredHostsController <Greg Sutcliffe>
- d7a6507 Correctly use V2 API namespace <Greg Sutcliffe>

Thanks!

PR #71, opened 2014-05-21
dashboard widget sample

abenari:
This is not a real widget just a wrapper for a future widget that you might want to write.


GregSutcliffe:
Awesome, I'll play with this and use it instead of Deface in 1.6 :)

GregSutcliffe:
Merged as 7b11c447c3fac173ea157cd7336c826bd3391f8c, thanks @abenari!

PR #72, opened 2014-06-04
Updated image names on lzap.fedorapeople.org

ibravo:


lzap:
Hello,

I apologize, these are pretty old. We are currently working on the new README that points to the correct directory which is:

http://downloads.theforeman.org/discovery/

The current version is 0.5 RC3. Final release pretty soon (no changes).

Thanks

PR #73, opened 2014-06-05
Added root pass inheritence from hostgroup/global setting.

radioactiv:


derektamsen:
I can also confirm this is an issue. When trying to provision a host from the discovered hosts report screen foreman will not let you continue unless a root password is entered directly for the host. I tested with the global setting `root_pass` set. I also attempted to provision the system with a host group that had a root password set. However, in both cases the provision host form from the discovered hosts report would not let me continue with an inherited root password.

Attempting to isolate the issue I successfully built a host manually using the "new host" form without supplying a root password and relying on inheritance.

This was running foreman 1.5 for el6 installed with the foreman installer. I was using ruby193-rubygem-foreman_discovery 1.3.0.rc1 (http://yum.theforeman.org/plugins/1.5/el6/x86_64/ruby193-rubygem-foreman_discovery-1.3.0-0.1.rc1.el6.noarch.rpm) of the foreman_discovery plugin supplied by the foreman plugins repo.

After applying this pull request manually on my server I am able to provision discovered hosts without specifying root passwords for each server.

lzap:
Oh right, I was wondering why this does not work. I can confirm this weird behavior too :-) Thanks.

GregSutcliffe:
Feels a bit odd to associate it to a hostgroup for just this, but I can confirm it fixes the problem.

GregSutcliffe:
Merged as fd96dc5, just in the nick of time for Discovery 1.3.0 - thanks @radioactiv!

PR #74, opened 2014-06-12
Installer image download support is in 1.6+

domcleal:


lzap:
Oh right, thanks. And I need to fix it (we have the new URL).

PR #78, opened 2014-06-25
fixes #6382 - use correct 48bit mac address regex

jlsherrill:
Opened http://projects.theforeman.org/issues/6388 for proper 64bit mac address support.


jlsherrill:
closing in favor of https://github.com/theforeman/foreman/pull/1541

PR #79, opened 2014-07-24
Refs #6560 - parsing method interface change

ares:
see https://github.com/theforeman/foreman/pull/1578 for more details, should be merge at the same time


GregSutcliffe:
Merged as cff5a66, thanks!

PR #82, opened 2014-08-07
Add nice dashboard widget

GregSutcliffe:
Also drops the last use of Deface! :clap: 

There's a minor issue with Taxonomies. If I have a non-admin user with :view_discovered_hosts, and 8 discovered hosts, 2 of which are in the user's taxonomies, then the widget displays correctly, but the discovered_hosts index does not. The log shows:

```
  Filter Load (0.3ms)  SELECT DISTINCT "filters".* FROM "filters" INNER JOIN "filterings" ON "filterings"."filter_id" = "filters"."id" INNER JOIN "permissions" ON "permissions"."id" = "filterings"."permission_id" INNER JOIN "roles" ON "filters"."role_id" = "roles"."id" INNER JOIN "cached_user_roles" ON "roles"."id" = "cached_user_roles"."role_id" LEFT JOIN taxable_taxonomies ON (filters.id = taxable_taxonomies.taxable_id AND taxable_type = 'Filter') LEFT JOIN taxonomies ON (taxonomies.id = taxable_taxonomies.taxonomy_id) WHERE "cached_user_roles"."user_id" = 3 AND (permissions.resource_type = 'Host') AND (permissions.name = 'view_discovered_hosts') AND (taxable_taxonomies.id IS NULL OR (taxonomies.type = 'Organization') OR (taxonomies.type = 'Location')) ORDER BY role_id, filters.id
  CACHE (0.0ms)  SELECT id FROM "taxonomies" WHERE (taxonomies.id = 4 or taxonomies.ancestry like '4/%' or taxonomies.ancestry = '4') ORDER BY title
  CACHE (0.0ms)  SELECT id FROM "taxonomies" WHERE (taxonomies.id = 3 or taxonomies.ancestry like '3/%' or taxonomies.ancestry = '3') ORDER BY title
  Host::Discovered Load (0.3ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."organization_id" IN (4) AND "hosts"."location_id" IN (3) AND "hosts"."type" IN ('Host::Discovered') AND (1=0) ORDER BY "hosts"."name" ASC LIMIT 20 OFFSET 0
```

Seems like the resource_type is wrong in the first line? @ares @elobato any ideas? I'm debating whether to care, discovered hosts are usually global anyway...


lib/foreman_discovery/engine.rb
```diff
@@ -59,6 +58,10 @@ class Engine < ::Rails::Engine
           :caption=> N_('Discovered hosts'),
           :parent => :hosts_menu,
           :after=>:hosts
+
+      # add dashboard widget
+      widget 'discovery_widget', :name=>N_('Discovery widget'), :sizex => 4, :sizey =>1
```
domcleal: indentation's off, and you need min foreman version of 1.6

app/views/dashboard/_discovery_widget.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% else %>
+  <table class="table table-striped ellipsis">
+    <% hosts.each do |host| %>
+      <tr>
+        <td><%= link_to h(host.name), discovered_host_path(host) %></td>
+        <td><%= host.ip  %></td>
+      </tr>
+    <% end %>
+  </table>
+  <p class="ca"><%= link_to _("Total pool size"), discovered_hosts_path %> - <%= all_hosts.size %> </p>
```
domcleal: I think n_() would be preferred here

app/models/host/discovered.rb
```diff
@@ -20,10 +20,19 @@ class Host::Discovered < ::Host::Base
...
+    conditions[:organization_id] = org.subtree_ids if org
+    conditions[:location_id]     = loc.subtree_ids if loc
+    where(conditions)
+  }
+
   # This `where` shouldn't be necessary, but randomly the sql query on the line above
   # sometimes loses the `where` clause, leading to all Hosts on the Discovery index
   # this seems like a rails bug, TODO: figure out whats really wrong here
-  scope :list, lambda { where(:type => "Host::Discovered").includes(:model, :location, :organization) }
+  #scope :list, lambda { where(:type => "Host::Discovered").includes(:model, :location, :organization) }
```
domcleal: commented code?

GregSutcliffe: Whoops, yeah, that scope got removed, but I was going to check the problem in the comment is no longer an issue first. I'll do that and repush in a bit.

ares:
@GregSutcliffe not sure where it comes from but this usually happens if you do something like 

``` ruby
::Host::Discovered.authorized(current_permission)
```

since Host model is not normal STI we have to specify which class are we interested in, I'd suggest trying following in DiscoveredHostsController#resource_base

``` ruby
Host::Base.authorized(current_permission, ::Host::Discovered)
```

foreman_discovery.gemspec
```diff
@@ -22,5 +22,4 @@ Gem::Specification.new do |s|
 
   s.add_dependency "open4"
   s.add_dependency "ftools"
-  s.add_dependency "deface"
```
lzap, line 25: The remaining two guys can go away as well. https://github.com/theforeman/foreman-packaging/pull/288

Mind doing this in a separate commit? Smells with conflicts :-)

GregSutcliffe, line 25: Removing deface is part of the purpose of this commit, and is entirely relevant.

GregSutcliffe:
@ares thanks for the help. For the record, the taxonomy problem turned out to be bad data in my db.

@domcleal comments addressed.

app/views/dashboard/_discovery_widget.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% else %>
+  <table class="table table-striped ellipsis">
+    <% hosts.each do |host| %>
+      <tr>
+        <td><%= link_to h(host.name), discovered_host_path(host) %></td>
+        <td><%= host.ip  %></td>
+      </tr>
+    <% end %>
+  </table>
+  <p class="ca"><%= link_to n_("Total pool size"), discovered_hosts_path %> - <%= all_hosts.size %> </p>
```
domcleal: usage of n_() looks wrong.. double check ;)

GregSutcliffe: Fixed, how does that look @domcleal?

domcleal: :+1:

GregSutcliffe:
Merged as 26d76b8, thanks for the help @ares & @domcleal!

PR #84, opened 2014-08-20
i18n production support

domcleal:
.mo files added and locale domain name fixed so i18n works in prod, also extracted the latest strings + pulled from tx.


.gitignore
```diff
@@ -15,7 +15,6 @@ tmp
 Gemfile.lock
 locale/*.mo
 locale/*/*.pox
-locale/*/LC_MESSAGES
```
lzap, line 18: Why is that? I think these subdirs are being created during the build...

locale/en_GB/foreman_discovery.po
```diff
@@ -7,11 +7,11 @@
...
 msgstr ""
-"Project-Id-Version: Foreman\n"
+"Project-Id-Version: foreman_discovery 1.3.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2014-04-08 15:59+0200\n"
-"PO-Revision-Date: 2014-08-04 08:40+0000\n"
-"Last-Translator: Dominic Cleal <dcleal@redhat.com>\n"
+"POT-Creation-Date: 2014-08-07 10:19+0100\n"
+"PO-Revision-Date: 2014-08-08 09:41+0000\n"
+"Last-Translator: Lukáš Zapletal\n"
```
lzap, line 14: The born English expert. _cough_ _cough_ _cough_

lib/foreman_discovery/engine.rb
```diff
@@ -18,7 +18,7 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_discovery.register_gettext', :after => :load_config_initializers do |app|
       locale_dir = File.join(File.expand_path('../../..', __FILE__), 'locale')
-      locale_domain = 'discovery'
+      locale_domain = 'foreman_discovery'
```
lzap, line 21: Hmmm so this was not working at all...

lzap:
:+1:

.gitignore
domcleal, line 18: There is no build?  They need to be part of the gem, same as we did in Hammer.

It's only in core that we don't ship the mo files and do it in packaging.  For plugin installs, gems really need to contain them as they're installed directly.

locale/en_GB/foreman_discovery.po
domcleal, line 14: Think Transifex attributes any .pot update to you...

lib/foreman_discovery/engine.rb
domcleal, line 21: Indeed!

.gitignore
lzap, line 18: Isn't there a hook avaiable during the `gem build` so we can build it there? I'd rather prefer that and make it a standard.

lzap, line 18: Does not look like...

http://guides.rubygems.org/specification-reference/#post_install_message

Ok :+1:

domcleal, line 18: indeed... :(  only way you can hack around it IIRC is to pretend you have a native extension, it's not good

lzap:
Merged, thanks:
- c267331 i18n - pulling from tx <Dominic Cleal>
- 0bc30a2 i18n - add Makefile to generate .mo files, fix domain name <Dominic Cleal>

PR #85, opened 2014-08-29
Fix mac address validation

ares:
Foreman has mac address valdiator since 1.6. Regexp does not exist
anymore.


lzap:
Sorry it took so long. Thank you @ares ! Works as expected. Merged as 04c963b.

PR #87, opened 2014-09-17
Improved instructions with foreman_url

lzap:
Not sure why we missed this, David Caplan noted. Fixing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ template and add:
     MENU LABEL Foreman Discovery
     MENU DEFAULT
     KERNEL boot/tftpboot/vmlinuz0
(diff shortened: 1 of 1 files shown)


domcleal:
For the record, I think this is probably blocked on http://projects.theforeman.org/issues/7828 ?

lzap:
That's right, closing for now, I need to find a better way than this. I like Settings[:foreman_url] to be available in templates...

domcleal:
Trouble with exposing settings generally is that there are secrets stored inside it, I think we'd need to whitelist/blacklist them, or stick to exposing specific helpers.

PR #88, opened 2014-09-29
Fix puppet parser on Foreman > 1.6 again

ares:
Also fixes case where one fact is upcase, second downcase

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -73,9 +73,16 @@ module ForemanDiscovery
     end
     # Include extensions to models in this config.to_prepare block
     config.to_prepare do
+
+      begin
+        require_dependency 'puppet_fact_parser'
+        ::PuppetFactParser.send :include, PuppetFactParserExtensions
+      rescue LoadError
+        Rails.logger.warn 'PuppetFactParser not found, not loading Parser extensions'
+      end
+
       # Include host extensions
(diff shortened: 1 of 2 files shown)


lib/foreman_discovery/engine.rb
```diff
@@ -73,9 +73,16 @@ class Engine < ::Rails::Engine
     end
     # Include extensions to models in this config.to_prepare block
     config.to_prepare do
+
+      begin
+        require_dependency 'puppet_fact_parser'
+        ::PuppetFactParser.send :include, PuppetFactParserExtensions
+      rescue LoadError
+        Rails.logger.warn 'PuppetFactParser not found, ignoring the fix for that'
```
GregSutcliffe: nitpick: Its not really a fix, it's an extension. Mind if I reword that as "not found, not loading Parser extensions" ?

ares: please do :-)

ares:
I did the reword and also fix one more issue (when fact is not present)

lzap:
How do I reproduce this one? Develop works fine against discovery develop for me.

GregSutcliffe:
@lzap check the PuppetFactParser extensions are getting loaded (maybe at some debug to check the chained methods are being called at import time)

lzap:
I can confirm that, looks like we can merge and release then!

lzap:
Merged as 213eedd for Discovery 1.4, thanks @ares !

PR #89, opened 2014-10-02
PXELinux global default is protected

dLobatog:
Cloning and editing is the way it should be done now. Alternatively we can provide a "PXELinux discovery default" template, thoughts?


domcleal:
The default in Foreman isn't locked, it's meant to be edited - that's a Katello misfeature.

That said, Discovery (and other plugins) could do with a more automated setup for this step.

EDIT: providing a different template, or telling a user to clone it isn't much help, as the name is hardcoded in Foreman for deployment.  Only the config template with that name will be deployed.

dLobatog:
Oh sorry, I didn't stop to think about it twice, it's a Katello feature yes. I shall close this until a better solution comes up.

PR #94, opened 2014-10-28
Fixes #6134 - Discovered hosts prefix can be configured

orrabin:


app/models/host/discovered.rb
```diff
@@ -32,7 +32,8 @@ class Host::Discovered < ::Host::Base
   def self.import_host_and_facts facts
     raise(::Foreman::Exception.new(N_("Invalid facts, must be a Hash"))) unless facts.is_a?(Hash)
     fact_name = Setting[:discovery_fact] || 'macaddress'
-    hostname = facts[fact_name].try(:downcase).try(:gsub,/:/,'').try(:sub,/^/,'mac')
+    hostname_prefix = Setting[:discovery_prefix] || 'mac'
```
orrabin: I didn't check if the prefix starts with a letter since hosts that start with numbers can be created.
If the discovered hosts have to start with letters I'll change this

app/models/host/discovered.rb
```diff
@@ -42,7 +43,7 @@ def self.import_host_and_facts facts
     h = ::Host::Discovered.find_by_name hostname
     h ||= Host.new :name => hostname, :type => "Host::Discovered"
     h.type = "Host::Discovered"
-    h.mac = facts[fact_name].try(:downcase)
+    h.mac = facts['macaddress'].try(:downcase)
```
lzap: Leftover? I don't understand why this is here.

app/models/setting/discovered.rb
```diff
@@ -12,6 +12,12 @@ def self.load_defaults
       ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
     end
 
+    Setting.transaction do
+      [
+          self.set('discovery_prefix', _("The default prefix to use for the host name"), "mac"),
```
lzap: Please add ", must not start with number" or even "must start with a letter".

app/models/host/discovered.rb
lzap: That's right. We need to check this, it's the purpose of the prefix actually. Please check:
- it does start with A-Za-z
- it is not empty

If any fails, please fallback to `mac`. If it's possible, cover with unit tests for the edge cases. The code looks good, once we have the check, I am going to test. Ty.

app/models/host/discovered.rb
orrabin: With `h.mac = facts[fact_name].try(:downcase)` I changed the fact_name to a random fact, tried provisioning and saw it's value in the hosts mac so I changed it to always take the mac address

lzap: That's not correct. We must end up with both hostname set to `prefix+fact_name` and mac set to `fact_name`. I have to admit this was confusing:

```
app/models/host/discovered.rb
34:    fact_name = Setting[:discovery_fact] || 'macaddress'
```

Since the `discovery_fact` now defaults to `discovery_bootif` you should do this (it was a minor bug):

```
fact_name = Setting[:discovery_fact] || 'discovery_bootif'
....
h.mac = facts[fact_name].try(:downcase)
```

orrabin:
@lzap in PuppetFactParserExtensions should 

``` ruby
def discovery_mac_fact_name
    Setting[:discovery_fact] || 'macaddress'
  end
```

also change to `discovery_bootif`?

orrabin:
the tests are failing because the regex `/^[a-zA-z].*/` somehow accepts "^abc" and "_abc". Other special characters are filtered ("-abc","$abc" are not valid prefixes) 

update: tests should now pass. This was a result of a typo: [a-zA-z] should be [a-zA-Z]

lzap:
Yes for the `discovery_mac_fact_name` please.

lzap:
Ori, if you need some assistance with disabling the test, let me know. We have a new method of doing that. I don't see any other feasible other way. Thanks.

lib/foreman_discovery/engine.rb
```diff
@@ -61,7 +61,9 @@ class Engine < ::Rails::Engine
 
         # add dashboard widget
         widget 'discovery_widget', :name=>N_('Discovery widget'), :sizex => 4, :sizey =>1
-
+        if respond_to? :tests_to_skip
+          tests_to_skip "HostTest::import host and facts" => ["should find a host by certname not fqdn when provided"]
+        end
```
GregSutcliffe: This isn't necessary. The core test is failing due to a bug in our chained method in PuppetFactParserExtensions. Essentially, one of the interfaces has no mac in the facts, meaning `facts["macaddress_#{interface}"].try(:downcase)` is `nil`. In core, `facts[discovery_mac_fact_name].try(:downcase)` is also nil because the discovery_bootif fact is not defined. So nil == nil and we return the wrong interface, rather than falling back on the unchained method.

See https://gist.github.com/d86c495d49907b7d9fca for a possible fix. Alternatively we could skip the entire block if `facts[discovery_mac_fact_name].try(:downcase)` is `nil` since it is then not possible to match an interface.

app/lib/puppet_fact_parser_extensions.rb
```diff
@@ -18,6 +18,7 @@ def primary_interface_with_discovery_fact
   # search for IP of interface with primary interface macaddress (ipaddress fact does not have to be interface used for boot)
   def ip_with_discovery_fact
     facts[:interfaces].split(',').each do |interface|
+      next if facts["macaddress_#{interface}"].nil?
```
GregSutcliffe: @ares do you think this would be better as `next if facts[discovery_mac_fact_name].nil?` ?

GregSutcliffe:
@orrabin can you rebase? I'll re-review after that.

orrabin:
@GregSutcliffe after rebase the tests fail locally because `cannot load such file -- open4`.
@lzap you removed the dependency but there is still use of open4 in `/lib/discovery.rake`
can I remove it completely? or should I add the dependency?

GregSutcliffe:
we can drop open4, it was for building the tcl image which is long-dead

ohadlevy:
the next error is that a require on ftools is missing.

GregSutcliffe:
same for ftools. the whole build_image task can go, actually

orrabin:
@GregSutcliffe rebased

GregSutcliffe:
Merged as b221e9c, thanks @orrabin!

korekhov:
Which Foreman release is this a part of? I have 1.7.1 currently and I don't see a way for me to update default value of "discovery_prefix" anywhere in UI, which results in undesired hostname generated for discovered hosts:

screen shot 2015-01-08 at 2 51 49 pm
screen shot 2015-01-08 at 2 52 22 pm

domcleal:
@korekhov it hasn't been released yet.  Look for a new release of Foreman Discovery soon from our plugin repos, I think they're working on a 2.0 (current is 1.4.1).

PR #95, opened 2014-10-30
Fixes #8141 - correct instructions for testing in rake file

orrabin:


Rakefile
```diff
@@ -102,8 +18,15 @@ namespace :test do
...
-    Dir.glob('test/**/*_test.rb') { |f| require f.sub('test/','') unless f.include? '.foreman_app/' }
+  task :all => [ 'test:set_loadpath'] do
+    puts <<MESSAGE
+
+Foreman source code is not present.
+Run test from foreman source code dir
+
+MESSAGE
+
+    Dir.glob('test/**/*_test.rb') { |f| require f.sub('test/','') }
```
orrabin: should this still be here if the tests are run from somewhere else?
now using `rake test` shows the message but with an error under it

Rakefile
```diff
@@ -8,92 +8,8 @@ task :default => :test
...
-BUNDLE_WITHOUT: console:development:fog:jsonp:libvirt:mysql:mysql2:ovirt:postgresql:vmware
-FILE
-      end
-    end
-
-    local_gemfile = "#{FOREMAN_DIR}/bundler.d/Gemfile.local.rb"
-    unless File.exist?(local_gemfile)
-      File.open(local_gemfile, 'w') { |f| f << <<GEMFILE }
-gem "puppet"
-gem "facter"
```
orrabin, line 63: should these gems be added to the plugin gemfile?

domcleal, line 63: They were deps of Foreman core which we didn't used to have in Foreman's Gemfile.  Puppet's no longer a dep, and Facter is now in the Gemfile, so it's fine.

orrabin, line 63: thanks

Gemfile
```diff
@@ -1,17 +1,3 @@
...
-
-  rake setup:foreman_prepare
-
-and try again.
-MESSAGE
-
-else
-  self.instance_eval(Bundler.read_file(FOREMAN_GEMFILE))
-end
+gemspec
```
lzap: Nitpick, missing EOL.

I am getting `rake is not part of the bundle. Add it to Gemfile. (Gem::LoadError)` when trying to do `bundle exec rake -T` in the discovery directory. I guess its no longer any issue, not sure yet.

Rakefile
lzap: We can remove this. Does `rake test` work for you?

I only use `bundle exec` and I get:

```
[lzap@lzapx foreman_discovery]$ be rake test
/home/lzap/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/bundler-1.5.3/lib/bundler/rubygems_integration.rb:240:in `block in replace_gem': rake is not part of the bundle. Add it to Gemfile. (Gem::LoadError)
        from /home/lzap/.rbenv/versions/2.0.0-p247/bin/rake:22:in `<main>'
```

Adding it to Gemfile won't hurt, I suppose?

lzap:
I think we can get rid of the Rakefile. For i18n we don't need it...

http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Pulling-translations-from-Transifex

Opinion @domcleal ?

Gemfile
```diff
@@ -2,16 +2,4 @@ source "http://rubygems.org"
...
-
-  rake setup:foreman_prepare
-
-and try again.
-MESSAGE
-
-else
-  self.instance_eval(Bundler.read_file(FOREMAN_GEMFILE))
-end
+gem 'rake'
```
orrabin: @lzap after adding this when I run 'rake test' I see: 
`Foreman source code is not present.
Run test from foreman source code dir`
but when I run `bundle exec rake -T` in discovery directory I get nothing

orrabin:
@lzap I fixed your comments. 
I will remove Rakefile completely if @domcleal answers it's ok.

domcleal:
I don't _think_ it's needed for anything.

lzap:
It works, thank you.

lzap:
I've merged as d64a8ab, thank you @orrabin

PR #97, opened 2014-11-07
Fixes #3860 - automated discovery provisioning

lzap:
This is a WIP.

This implements simple rule engine which allows to assign host groups according
to scoped_search queries. Currently all our facts are strings, so searching is
little bit limited, but it works just fine.

Good testing query is something like `facts.architecture = x86_64` which is
likely to be true. Here are some rules to start with which can be created via
rails console:

```
=> [#<DiscoveryRule id: 1, name: "Database server rule", query: "facts.architecture = i386", hostgroup_id: 4, hostname: "test<%= facts['macaddress'] %>", max_count: 0, priority: 10, enabled: true, created_at: "2014-11-07 11:37:18", updated_at: "2014-11-07 11:37:18">,
 #<DiscoveryRule id: 2, name: "Web server rule", query: "facts.architecture = x86_64", hostgroup_id: 3, hostname: "test<%= facts['macaddress'] %>", max_count: 5, priority: 20, enabled: true, created_at: "2014-11-07 11:37:52", updated_at: "2014-11-07 11:37:52">,
 #<DiscoveryRule id: 3, name: "Web server drained rule", query: "facts.architecture = x86_64", hostgroup_id: 3, hostname: "test<%= facts['macaddress'] %>", max_count: 2, priority: 15, enabled: true, created_at: "2014-11-07 13:13:02", updated_at: "2014-11-07 13:13:02">,
 #<DiscoveryRule id: 4, name: "Disabled dummy rule", query: "facts.architecture = x86_64", hostgroup_id: 3, hostname: "test<%= facts['macaddress'] %>", max_count: 0, priority: 50, enabled: false, created_at: "2014-11-07 13:20:59", updated_at: "2014-11-07 13:20:59">]
```

There is no UI support yet, I am providing (untested) V2 API for rules CRUD. To
get this working in real-world scenario with Discovery 2.0, you need this patch
because systemd is too quick with shutdown:
https://github.com/theforeman/smart-proxy/pull/235

To test this, checkout the branch and do migration. Then create some rules via
V2 API. Once a discovery node is found, it should automaticaly start
provisioning (reboot can fail rolling back the transaction because of the shell
API provider issue). You can re-initiate auto-provisioning using an API call
(PUT discovery_rules/execute action).

Make sure the discovery_auto option is set to true when you want new nodes to
be automatically picked up for provisioning.

Also we need to make sure the host group has _everything_ set for provisioning,
otherwise the provisioning will fail (and possibly can end up in a loop). You
need to make sure root password is set as well. I think we need some kind of
check somewhere (in the Foreman core perhaps)?

TODO items:
- hammer CLI
- UI
- hostname setting via safe_mode
- validation of hostgroups
- testing

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -95,6 +95,11 @@ module Api
 
       def facts
         @discovered_host, state = Host::Discovered.import_host_and_facts(params[:facts])
+        if state && Setting['discovery_auto']
+          state = ForemanDiscovery.execute_auto_provisioning(@discovered_host)
+        else
+          Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning"
+        end
         process_response state
       rescue ::Foreman::Exception => e
         render :json => {'message'=>e.to_s}, :status => :unprocessable_entity
--- a/app/controllers/api/v2/discovery_rules_controller.rb
+++ b/app/controllers/api/v2/discovery_rules_controller.rb
@@ -1,0 +1,88 @@
+module Api
+  module V2
+    class DiscoveryRulesController < ::Api::V2::BaseController
+
+      before_filter :find_resource, :except => %w{index create facts}
+
+      resource_description do
+        resource_id 'discovery_rules'
+        api_version 'v2'
+        api_base_url "/api/v2"
+      end
+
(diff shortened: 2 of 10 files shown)


lzap:
This got replaced with https://github.com/theforeman/foreman_discovery/pull/97

PR #98, opened 2014-11-10
Fixes #3860 - automated discovery provisioning

lzap:
Replaces https://github.com/theforeman/foreman_discovery/pull/97

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -1,8 +1,9 @@
 module Api
   module V2
     class DiscoveredHostsController < ::Api::V2::BaseController
+      include Foreman::Controller::DiscoveredExtensions
 
-      before_filter :find_resource, :except => %w{index create facts}
+      before_filter :find_resource, :except => %w{index create facts auto_provision_all}
       skip_before_filter :authorize, :only => :facts
 
       resource_description do
@@ -95,9 +96,59 @@ module Api
 
       def facts
         @discovered_host, state = Host::Discovered.import_host_and_facts(params[:facts])
+        if state && rule = find_discovery_rule(@discovered_host)
+          state = perform_auto_provision(@discovered_host.becomes(::Host::Managed), rule) if Setting['discovery_auto']
+        else
+          Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning"
+        end
         process_response state
       rescue ::Foreman::Exception => e
         render :json => {'message'=>e.to_s}, :status => :unprocessable_entity
+      end
+
+      api :POST, "/discovered_hosts/:id/auto_provision", N_("Execute rules against a discovered host")
+
+      def auto_provision
+        @discovered_host.transaction do
+          if rule = find_discovery_rule(@discovered_host)
+            msg = _("Host %s was provisioned with rule %s") % [@discovered_host.name, rule.name]
+            process_response perform_auto_provision(@discovered_host.becomes(::Host::Managed), rule), msg
+          else
+            process_success _("No rule found for host %s") % @discovered_host.name
+          end
+        end
+      rescue ::Foreman::Exception => e
+        render :json => {'message' => e.to_s}, :status => :unprocessable_entity
+      end
+
+      api :POST, "/discovered_hosts/auto_provision_all", N_("Execute rules against all currently discovered hosts")
+
+      def auto_provision_all
+        result = true
+        Host.transaction do
+          overall_errors = ""
+          Host::Discovered.all.each do |discovered_host|
+            if rule = find_discovery_rule(discovered_host)
+              result &= perform_auto_provision(discovered_host.becomes(::Host::Managed), rule)
+              unless discovered_host.errors.empty?
+                errors = discovered_host.errors.full_messages.join(' ')
+                logger.warn "Failed to auto provision host %s: %s" % [discovered_host.name, errors] 
+                overall_errors << "#{discovered_host.name}: #{errors} "
+              end
+            else
+              logger.warn "No rule found for host %s" % discovered_host.name
+            end
+          end
+          if result
+            process_success _("Discovered hosts are provisioning now")
+          else
+            render_error :custom_error,
+              :status => :unprocessable_entity,
+              :locals => {
+                :message => _("Errors during auto provisioning: %s") % overall_errors
+              }
+          end
+        end
       end
 
       private
@@ -108,6 +159,17 @@ module Api
 
       def forward_request_url
         @discovered_host.request_url = request.host_with_port if @discovered_host.respond_to?(:request_url)
+      end
+
+      def action_permission
+        case params[:action]
+          when 'auto_provision'
+            :auto_provision
+          when 'auto_provision_all'
+            :auto_provision_all
+          else
+            super
+        end
       end
 
     end
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -1,0 +1,42 @@
+module Foreman::Controller::DiscoveredExtensions
+  extend ActiveSupport::Concern
+
+  # return auto provision rule or false when not present
+  def find_discovery_rule host
+    Rails.logger.debug "Finding auto discovery rule for host #{host.name} (#{host.id})"
+    # for each discovery rule ordered by priority
+    DiscoveryRule.where(:enabled => true).order(:priority).each do |rule|
+      max = rule.max_count
+      usage = rule.hosts.size
+      Rails.logger.debug "Found rule #{rule.name} (#{rule.id}) [#{usage}/#{max}]"
+      # if the rule has free slots
+      if max == 0 || usage < max
+        # try to match the search
+        begin
+          if Host::Discovered.where(:id => host.id).search_for(rule.search).size > 0
+            Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
+            return rule
+          end
+        rescue ScopedSearch::QueryNotSupported => e
+          Rails.logger.warn "Invalid query for rule #{rule.name} (#{rule.id}): #{e.message}"
+        end
+      else
+        Rails.logger.info "Skipping drained rule #{rule.name} (#{rule.id}) with max set to #{rule.max_count}"
+      end
+    end
+    return false
+  end
+
+  # trigger the provisioning
+  def perform_auto_provision host, rule
+    host.type = 'Host::Managed'
+    host.managed = true
+    host.build = true
+    host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
+    host.hostgroup_id = rule.hostgroup_id
+    host.comment = "Auto-discovered and provisioned via rule '#{rule.name}'"
+    host.discovery_rule = rule
+    host.save # save! does not work here
+  end
+
+end
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -1,12 +1,13 @@
 class DiscoveredHostsController < ::ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   include Foreman::Controller::TaxonomyMultiple
+  include Foreman::Controller::DiscoveredExtensions
   unloadable
 
   # Avoid auth for discovered host creation
   skip_before_filter :require_login, :require_ssl, :authorize, :verify_authenticity_token, :set_taxonomy, :session_expiry, :update_activity_time, :only => :create
 
-  before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert]
+  before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert auto_provision]
   before_filter :find_multiple, :only => [:multiple_destroy, :submit_multiple_destroy]
   before_filter :taxonomy_scope, :only => [:edit]
 
@@ -121,6 +122,44 @@ class DiscoveredHostsController < ::ApplicationController
     render :json => @items
   end
 
+  def auto_provision
+    @host.transaction do
+      if rule = find_discovery_rule(@host)
+        if perform_auto_provision(@host.becomes(::Host::Managed), rule)
+          process_success :success_msg => _("Host %s was provisioned with rule %s") % [@host.name, rule.name], :success_redirect => :back
+        else
+          errors = @host.errors.full_messages.join(' ')
+          logger.warn "Failed to auto provision host %s: %s" % [@host.name, errors] 
+          process_error :error_msg => _("Failed to auto provision host %s: %s") % [@host.name, errors], :redirect => :back
+        end
+      else
+        process_success :success_msg => _("No rule found for host %s") % @host.name, :success_redirect => :back
+      end
+    end
+  end
+
+  def auto_provision_all
+    result = true
+    Host.transaction do
+      overall_errors = ""
+      Host::Discovered.all.each do |discovered_host|
+        if rule = find_discovery_rule(discovered_host)
+          result &= perform_auto_provision(discovered_host.becomes(::Host::Managed), rule)
+          unless discovered_host.errors.empty?
+            errors = discovered_host.errors.full_messages.join(' ')
+            logger.warn "Failed to auto provision host %s: %s" % [discovered_host.name, errors] 
+            overall_errors << "#{discovered_host.name}: #{errors} "
+          end
+        end
+      end
+      if result
+        process_success :success_msg => _("Discovered hosts are provisioning now"), :success_redirect => :back
+      else
+        process_error :error_msg => _("Errors during auto provisioning: %s") % overall_errors, :redirect => :back
+      end
+    end
+  end
+
   private
 
   def resource_base
@@ -162,6 +201,10 @@ class DiscoveredHostsController < ::ApplicationController
         :edit
       when 'submit_multiple_destroy', 'multiple_destroy'
         :destroy
+      when 'auto_provision'
+        :auto_provision
+      when 'auto_provision_all'
+        :auto_provision_all
       else
         super
     end
--- a/app/controllers/discovery_rules_controller.rb
+++ b/app/controllers/discovery_rules_controller.rb
@@ -1,0 +1,73 @@
+class DiscoveryRulesController < ApplicationController
+  # Two-pane searching must be disabled for rules otherwise query completion will not work
+  #include Foreman::Controller::AutoCompleteSearch
+
+  before_filter :find_resource, :only => [:edit, :update, :destroy, :enable, :disable, :auto_provision]
+
+  def index
+    base = resource_base.search_for(params[:search], :order => (params[:order] || 'priority ASC'))
+    @discovery_rules = base.paginate(:page => params[:page]).includes(:hostgroup)
+  end
+
+  def new
+    @discovery_rule = DiscoveryRule.new
+  end
+
+  def create
+    @discovery_rule = DiscoveryRule.new(params[:discovery_rule])
+    if @discovery_rule.save
+      process_success
+    else
+      process_error
+    end
+  end
+
+  def edit
+  end
+
+  def update
+    if @discovery_rule.update_attributes(params[:discovery_rule])
+      process_success
+    else
+      process_error
+    end
+  end
+
+  def destroy
+    if @discovery_rule.destroy
+      process_success
+    else
+      process_error
+    end
+  end
+
+  def enable
+    set_enabled true
+  end
+
+  def disable
+    set_enabled false
+  end
+
+  private
+
+  def action_permission
+    case params[:action]
+    when 'enable'
+      :enable
+    when 'disable'
+      :disable
+    else
+      super
+    end
+  end
+
+  def set_enabled(enabled)
+    @discovery_rule.enabled = enabled
+    if @discovery_rule.save
+      process_success :success_msg => (enabled ? _('Rule enabled') : _('Rule disabled')), :success_redirect => :back
+    else
+      process_error
+    end
+  end
+end
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -4,6 +4,8 @@ module Host::ManagedExtensions
   included do
     # execute standard callbacks
     after_validation :queue_reboot
+
+    belongs_to :discovery_rule
   end
 
   def queue_reboot
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -9,6 +9,7 @@ class Setting::Discovered < ::Setting
     Setting.transaction do
       [
         self.set('discovery_fact', _("The default fact name to use for the MAC of the system"), "discovery_bootif"),
+        self.set('discovery_auto', _("Automatically provision newly discovered hosts, according to the provisioning rules"), false),
       ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
     end
 
--- a/db/migrate/20141107091416_create_discovery_rules.rb
+++ b/db/migrate/20141107091416_create_discovery_rules.rb
@@ -1,0 +1,15 @@
+class CreateDiscoveryRules < ActiveRecord::Migration
+  def change
+    create_table :discovery_rules do |t|
+      t.string :name, :length => 254
+      t.string :search, :length => 254
+      t.integer :hostgroup_id, :null => false
+      t.string :hostname, :length => 254, :default => ''
+      t.integer :max_count, :default => 0
+      t.integer :priority, :default => 0
+      t.boolean :enabled, :default => true, :null => false
+
+      t.timestamps
+    end
+  end
+end
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -6,6 +6,11 @@ module ForemanDiscovery
   #Thus, inherits from ::Rails::Engine and not from Rails::Engine
   class Engine < ::Rails::Engine
 
+    # support for concerns for pre-4 Rails versions
+    config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
+    config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
+    config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
+
     # Load this before the Foreman config initializers, so that the Setting.descendants
     # list includes the plugin STI setting class
     initializer 'foreman_discovery.load_default_settings', :before => :load_config_initializers do |app|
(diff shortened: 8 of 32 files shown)


app/controllers/discovery_rules_controller.rb
```diff
@@ -0,0 +1,92 @@
+class DiscoveryRulesController < ApplicationController
+  # Searching must be disabled for rules otherwise query completion will not work
+  #include Foreman::Controller::AutoCompleteSearch
```
GregSutcliffe, line 3: wait what? why do you have to _disable_ search to _enable_ autocompletion?

app/models/setting/discovered.rb
```diff
@@ -9,6 +9,7 @@ def self.load_defaults
     Setting.transaction do
       [
         self.set('discovery_fact', _("The default fact name to use for the MAC of the system"), "discovery_bootif"),
+        self.set('discovery_auto', _("Initiate auto provision rule processing for incoming hosts"), false),
```
GregSutcliffe: The description would read better as "Automatically provision newly discovered hosts, according to the provisioning rules"

db/migrate/20141107091416_create_discovery_rules.rb
```diff
@@ -0,0 +1,15 @@
+class CreateDiscoveryRules < ActiveRecord::Migration
```
GregSutcliffe, line 1: Oh well, I suppose discovery was going to gain some migrations sooner or later. I'd done so well avoiding that so far :D

lib/foreman_discovery.rb
```diff
@@ -1,4 +1,39 @@
 require 'foreman_discovery/engine'
 
 module ForemanDiscovery
+  def self.execute_auto_provisioning dhost=nil
```
GregSutcliffe: Does it really make sense to have this in lib? We've been moving code out of lib in foreman-core. Perhaps it should be in the DiscoveryRule model?

app/controllers/discovery_rules_controller.rb
lzap, line 3: > > -  # Searching must be disabled for rules otherwise query completion will not work
> 
> wait what? why do you have to _disable_ search to _enable_ autocompletion?

When there are two or more scoped_search components on the page, it
simply breaks and both does not work. If someone knows how to get around
it, please let me know.

## 

Later,
 Lukas #lzap Zapletal

lib/foreman_discovery.rb
lzap: > Does it really make sense to have this in lib? We've been moving code out of lib in foreman-core. Perhaps it should be in the DiscoveryRule model?

Yeah, actually I have a small refactoring locally that adds a Concern
for this. This way I can also use safe_mode which is only available in a
model classes.

Let me push on Friday.

## 

Later,
 Lukas #lzap Zapletal

app/controllers/discovery_rules_controller.rb
GregSutcliffe, line 3: Why are there two scoped_search components? Where is the other one coming from?

lib/foreman_discovery.rb
GregSutcliffe: Excellent, I'll test as-is and look forward to the update ;)

lib/foreman_discovery/engine.rb
```diff
@@ -50,20 +50,39 @@ class Engine < ::Rails::Engine
...
+          permission :view_discovery_rules, {:discovery_rules => [:index, :show, :auto_complete_search] }
+          permission :new_discovery_rules, {:discovery_rules => [:new, :create] }
+          permission :edit_discovery_rules, {:discovery_rules => [:edit, :update, :enable, :disable] }
+          permission :execute_discovery_rules, {:discovery_rules => [:execute, :execute_all] }
+          permission :delete_discovery_rules, {:discovery_rules => [:destroy] }
+        end
+
+        # roles (only added if they don't exist)
+        role "Discovery", [:view_discovered_hosts, :provision_discovered_hosts, :edit_discovered_hosts, :destroy_discovered_hosts, :view_discovery_rules, :new_discovery_rules, :edit_discovery_rules, :execute_discovery_rules, :delete_discovery_rules]
+        role "Discovery Reader", [:view_discovered_hosts, :view_discovery_rules]
```
GregSutcliffe, line 84: "View Discovery" would be more consistent with "View Hosts", "View Domains" etc

lib/foreman_discovery/engine.rb
```diff
@@ -50,20 +50,39 @@ class Engine < ::Rails::Engine
...
+        security_block :security_block_discovery_rules do
+          permission :view_discovery_rules, {:discovery_rules => [:index, :show, :auto_complete_search] }
+          permission :new_discovery_rules, {:discovery_rules => [:new, :create] }
+          permission :edit_discovery_rules, {:discovery_rules => [:edit, :update, :enable, :disable] }
+          permission :execute_discovery_rules, {:discovery_rules => [:execute, :execute_all] }
+          permission :delete_discovery_rules, {:discovery_rules => [:destroy] }
+        end
+
+        # roles (only added if they don't exist)
+        role "Discovery", [:view_discovered_hosts, :provision_discovered_hosts, :edit_discovered_hosts, :destroy_discovered_hosts, :view_discovery_rules, :new_discovery_rules, :edit_discovery_rules, :execute_discovery_rules, :delete_discovery_rules]
```
GregSutcliffe: I guess we should rename this to "Discovery Manager" to match the other full-edit roles in core?

lib/foreman_discovery.rb
```diff
@@ -1,4 +1,39 @@
...
+            break
+          end
+        else
+          Rails.logger.info "Skipping drained rule #{rule.name} (#{rule.id}) with max set to #{rule.max_count}"
+        end
+      end
+      return true
+    else
+      # For each discovered host
+      Host::Discovered.order(:last_report).each do |discovered_host|
```
GregSutcliffe: I'm curious: why order by last_report?

GregSutcliffe:
Doesn't work for this workflow:

1) Ensure there are no rules (fresh db in my case)
2) Discover a host, ensure it shows up in the discovered_hosts index
3) Create a rule which matches the discovered host, save it (I used 'subnet = 192.168.131.0')
4) Click Execute on the rule

Expected result:
Host reboots and provisions

Actual Result:
Logs show "Executing discovery rules for host Debian (1)" and then:
Foreman::Exception (ERF42-4516 [Foreman::Exception]: Unknown action name for success message: execute):

Host does not reboot.

EDIT: disregard this, this issue is with the lib code expecting a Host but receiving a Rule from the controller.

app/controllers/discovery_rules_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+  def destroy
+    if @discovery_rule.destroy
+      process_success
+    else
+      process_error
+    end
+  end
+
+  def execute
+    if ForemanDiscovery.execute_auto_provisioning(@discovery_rule)
```
GregSutcliffe: You pass a Rule into execute_auto_provisioning here, but the code in lib/foreman_discovery.rb seems to be expecting a Host (see below).

lib/foreman_discovery.rb
```diff
@@ -1,4 +1,39 @@
 require 'foreman_discovery/engine'
 
 module ForemanDiscovery
+  def self.execute_auto_provisioning dhost=nil
+    if dhost
+      Rails.logger.debug "Executing discovery rules for host #{dhost.name} (#{dhost.id})"
```
GregSutcliffe: (from above) You seem to pass a Rule in from the controller, but all the code in this method seems to be aimed at a Host (the var `dhost` implies this, as does the `.becomes(Host::Managed)` below). This leads to oddities like `Executing discovery rules for host Debian (1)` when I would expect to see `Executing discovery rules for host mac112233445566 (1)`. I don't see how this can work.

GregSutcliffe:
Auto-provisiong at discovery time also fails:

```
Merging facts for 'home525400a95489': added 57 facts
Import facts for 'home525400a95489' completed. Added: 57, Updated: 0, Deleted 0 facts
We have following interfaces 'ens3' based on facts
Skipping ens3 since it is primary interface of host home525400a95489
Discovered facts import unsuccessful, skipping auto provisioning
Completed 201 Created in 852.9ms (Views: 5.6ms | ActiveRecord: 0.0ms)
```

I'm not sure why it thinks the fact import was unsuccessful

GregSutcliffe:
Ah, it fails because the Setting is false. Needs a better error mesasge :)

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -95,6 +95,11 @@ def destroy
 
       def facts
         @discovered_host, state = Host::Discovered.import_host_and_facts(params[:facts])
+        if state && Setting['discovery_auto']
+          state = ForemanDiscovery.execute_auto_provisioning(@discovered_host)
+        else
+          Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning"
```
GregSutcliffe, line 102: This error meesage is misleading when the state is set but the Setting is false.

app/controllers/discovery_rules_controller.rb
GregSutcliffe: Should this not also check the value of Setting[:discovery_auto] ?

GregSutcliffe:
Ok, auto_provisioning at discovery time does fire correctly, but fails to reboot the host for some reason:

```
Match found for host home525400a95489 (5) rule Debian (1)
ForemanDiscovery: Rebooting home525400a95489.provisioning.sapphire.local as its being discovered and assigned
Failed to reboot: undefined method `[]' for true:TrueClass
Rolling back due to a problem: [Rebooting home525400a95489.provisioning.sapphire.local   10000   failed  [#<Host::Managed id: 5, name: "home525400a95489.provisioning.sapphire.local", ... discovery_rule_id: nil>, :setReboot]]
```

I'm not clear why it fails yet - normal manual provisioning from the UI seems fine...

EDIT: to clarify, reboot happens, but because the save fails, the host remains as a Host::Discovered instead of moving to Host::Managed, and then the installer fails to retrieve the /unattended/provision URL

app/controllers/discovery_rules_controller.rb
lzap, line 3: > Why are there two scoped_search components? Where is the other one coming from?

I have mine that is used for the rules themselves.

But then there is the standard search. And since the edit form is
in-line (pop up) one, the original is still visible and kills mine.

To me it looked like it is not necessary to have searching capabilities
for rules, so its a non-issue.

## 

Later,
 Lukas #lzap Zapletal

lib/foreman_discovery/engine.rb
lzap: > I guess we should rename this to "Discovery Manager" to match the other full-edit roles in core?

Sure, but is it worth the effort? I mean, I can't rename it here just
like that. We need a DB SQL migration I guess.

## 

Later,
 Lukas #lzap Zapletal

lib/foreman_discovery/engine.rb
lzap, line 84: > "View Discovery" would be more consistent with "View Hosts", "View Domains" etc

Not getting this one.

## 

Later,
 Lukas #lzap Zapletal

lib/foreman_discovery.rb
lzap: > I'm curious: why order by last_report?

Yeah, my idea was to achieve order of discovery. Changing this to
created_at.

## 

Later,
 Lukas #lzap Zapletal

lzap:
> Foreman::Exception (ERF42-4516 [Foreman::Exception]: Unknown action name for success message: execute):

Yeah, you were too fast. Sorry about that. The "manual" execution from
UI/API was not yet tested. Will finish that today hopefully.

## 

Later,
 Lukas #lzap Zapletal

app/controllers/discovery_rules_controller.rb
lzap: > You pass a Rule into execute_auto_provisioning here, but the code in lib/foreman_discovery.rb seems to be expecting a Host (see below).

Yeah, this is still prototype kind of thing. Will finish soon.

## 

Later,
 Lukas #lzap Zapletal

app/controllers/api/v2/discovered_hosts_controller.rb
lzap, line 102: > This error meesage is misleading when the state is set but the Setting is false.

For all the related comments - yeah, I fixed the if clause to give
proper error message. When it's disabled, it's disabled and we don't
actually do anything.

## 

Later,
 Lukas #lzap Zapletal

app/controllers/discovery_rules_controller.rb
lzap: > Should this not also check the value of Setting[:discovery_auto] ?

Yeah, I am moving this into this helper (concern now). Good idea.

## 

Later,
 Lukas #lzap Zapletal

lzap:
> Failed to reboot: undefined method `[]' for true:TrueClass

This is known and fixed bug. You must have an old build of FDI without
reboot race condition bug fix in the BMC shell provider. What is the FDI
image version?

This was fixed in 2.0:

https://github.com/theforeman/smart-proxy/pull/235

## 

Later,
 Lukas #lzap Zapletal

lzap:
Greg, thanks for your testing. Due to the course this week I was not able to stop you fast. Some portions of the code were only concepts (manual exec mostly). I am finishing with little bit of refactoring, hostname templating via safemode and manual execution. I will push this and then start with some unit tests.

lzap:
@GregSutcliffe so it is in much better shape now. Please test. You can now use ERB for hostname (e.g. "webserver-<%= rand(99999) %>" but for this you need the patch above in the core). The main code logic is in a model concern now. Provisioning works, BUT

there is one issue. For some reason, the host gets saved and orchestration is triggered including setting up proxy and rebooting. But the host type still stays as `Host::Discovered`. I don't see why it does not work while the code in `lib/` was working fine.

Please do formal review now, unit tests to come next week.

app/controllers/discovery_rules_controller.rb
GregSutcliffe, line 3: Ah, it's the two-pane thing, right. Can you update the comment?

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -95,11 +96,32 @@ def destroy
...
+        end
         process_response state
       rescue ::Foreman::Exception => e
         render :json => {'message'=>e.to_s}, :status => :unprocessable_entity
       end
 
+      api :POST, "/discovered_hosts/:id/auto_provision", N_("Execute rules against a discovered host")
+
+      def auto_provision
+        # XXX TODO HEY
```
GregSutcliffe: I think you fogot these TODOs :)

lzap: > I think you fogot these TODOs :)

Yeah, I know. I need to write tests for the API. Will be provided in the
next push. Updated the comment (above).

## 

Later,
 Lukas #lzap Zapletal

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -0,0 +1,37 @@
...
+  extend ActiveSupport::Concern
+
+  # return auto provision rule or false when not present
+  def find_discovery_rule host
+    Rails.logger.debug "Finding auto discovery rule for host #{host.name} (#{host.id})"
+    # for each discovery rule ordered by priority
+    DiscoveryRule.where(:enabled => true).order(:priority).each do |rule|
+      max = rule.max_count
+      usage = rule.hosts.size
+      Rails.logger.debug "Applying rule #{rule.name} (#{rule.id}) [#{usage}/#{max}]"
```
GregSutcliffe: This log line is inaccurate, as the rule isn't applied here now (sometimes we just call the method to find a rule and display it)

app/controllers/discovered_hosts_controller.rb
```diff
@@ -121,6 +122,36 @@ def auto_complete_search
     render :json => @items
   end
 
+  def auto_provision
+    @host.transaction do
```
GregSutcliffe, line 126: Why use a transaction here? What are we trying to protect?

GregSutcliffe, line 126: What I mean is, I'm not against a transaction, I just want to understand if we need similar protection in other places where we use .becomes()

app/models/host/managed_extensions.rb
```diff
@@ -4,6 +4,8 @@ module Host::ManagedExtensions
   included do
     # execute standard callbacks
     after_validation :queue_reboot
+
```
GregSutcliffe: trailing whitespace

app/models/host/managed_extensions.rb
```diff
@@ -20,6 +22,7 @@ def setReboot
     else
       logger.info "ForemanDiscovery: reboot result: failed"
     end
+    true
```
GregSutcliffe: Should we return the failure case to the UI so the users knows to go manually reboot the host?

GregSutcliffe:
Looking great! I tested all 3 methods (auto-on-discover, per-host-auto, all-auto) and all worked as expected. Need to finish up the tests and maybe check the API works, but then we're done :)

lib/foreman_discovery/engine.rb
```diff
@@ -50,19 +55,42 @@ class Engine < ::Rails::Engine
...
+        menu :top_menu, :discovery_rules, :url_hash => {:controller => :discovery_rules, :action => :index},
+          :caption => N_('Discovery rules'),
+          :parent => :configure_menu,
+          :after=> :hostgroups
+
         # add dashboard widget
         widget 'discovery_widget', :name=>N_('Discovery widget'), :sizex => 4, :sizey =>1
+
+        # add template helpers
+        allowed_template_helpers :rand, :facts_hash
```
GregSutcliffe, line 101: This works if I use `debian<%= facts_hash['uniqueid'] %>` but not if I use `debian<%= @host.facts_hash['uniqueid'] %>` - would be nice if users could use the same ERB conventions they use in other places, but I don't know if that's possible? Not a big deal...

lzap:
Ok tests are ready. I made one additional change - I do not allow nils for the hostname, it is either something (I validate if its starts with a letter or ERB) or blank. Here is the diff from the latest version you tested:

``` diff
diff --git a/app/controllers/concerns/foreman/controller/discovered_extensions.rb b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
index 1b96690..3f38441 100644
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -12,9 +12,13 @@ module Foreman::Controller::DiscoveredExtensions
       # if the rule has free slots
       if max == 0 || usage < max
         # try to match the search
-        if Host::Discovered.where(:id => host.id).search_for(rule.search).size > 0
-          Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
-          return rule
+        begin
+          if Host::Discovered.where(:id => host.id).search_for(rule.search).size > 0
+            Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
+            return rule
+          end
+        rescue ScopedSearch::QueryNotSupported => e
+          Rails.logger.warn "Invalid query for rule #{rule.name} (#{rule.id}): #{e.message}"
         end
       else
         Rails.logger.info "Skipping drained rule #{rule.name} (#{rule.id}) with max set to #{rule.max_count}"
@@ -27,7 +31,7 @@ module Foreman::Controller::DiscoveredExtensions
     host.type = 'Host::Managed'
     host.managed = true
     host.build = true
-    host.name = host.render_template(rule.hostname || host.name)
+    host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
     host.hostgroup_id = rule.hostgroup_id
     host.comment = "Auto-discovered and provisioned via rule '#{rule.name}'"
     host.discovery_rule = rule
diff --git a/app/models/discovery_rule.rb b/app/models/discovery_rule.rb
index 2771b6e..f02e2b8 100644
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -9,6 +9,8 @@ class DiscoveryRule < ActiveRecord::Base
   validates :name, :presence => true, :uniqueness => true,
     :format => { :with => /\A(\S+)\Z/, :message => N_("can't contain white spaces.") }
   validates :search, :presence => true
+  validates :hostname, :format => { :with => /\A[a-zA-Z<]/, :message => N_("must start with a letter or ERB.") },
+    :allow_blank => true
   validates :hostgroup_id, :presence => true
   validates :max_count, :numericality => { only_integer: true }
   validates :priority, :presence => true, numericality: true
diff --git a/db/migrate/20141107091416_create_discovery_rules.rb b/db/migrate/20141107091416_create_discovery_rules.rb
index 1fbd95a..a66cb40 100644
--- a/db/migrate/20141107091416_create_discovery_rules.rb
+++ b/db/migrate/20141107091416_create_discovery_rules.rb
@@ -4,7 +4,7 @@ class CreateDiscoveryRules < ActiveRecord::Migration
       t.string :name, :length => 254
       t.string :search, :length => 254
       t.integer :hostgroup_id, :null => false
-      t.string :hostname, :length => 254
+      t.string :hostname, :length => 254, :default => ''
       t.integer :max_count, :default => 0
       t.integer :priority, :default => 0
       t.boolean :enabled, :default => true, :null => false
diff --git a/test/unit/discovered_extensions_test.rb b/test/unit/discovered_extensions_test.rb
index 037ae31..0d9d975 100644
--- a/test/unit/discovered_extensions_test.rb
+++ b/test/unit/discovered_extensions_test.rb
@@ -11,35 +11,96 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     }
   end

-  test "no rule is found for zero discovered hosts and zero rules" do
+  test "no rule is found for empty rule set" do
     host = FactoryGirl.create(:host)
     refute find_discovery_rule(host)
   end

-  test "no rule is found for zero discovered hosts" do
-    host = FactoryGirl.create(:host)
-    FactoryGirl.create(:discovery_rule)
-    refute find_discovery_rule(host)
-  end
-
   test "no rule is found out of one for a discovered host with no facts" do
     host = FactoryGirl.create(:host)
-    FactoryGirl.create(:discovery_rule, :search => "facts.foo == 'bar'")
+    FactoryGirl.create(:discovery_rule, :search => "facts.foo == bar")
     refute find_discovery_rule(host)
   end

   test "no rule is found out of one for a discovered host with some facts" do
     host = FactoryGirl.create(:host, :with_facts)
-    FactoryGirl.create(:discovery_rule, :search => "facts.foo == 'does not exist'")
+    FactoryGirl.create(:discovery_rule, :search => "facts.foo == does not exist")
+    refute find_discovery_rule(host)
+  end
+
+  test "no rule is found out of two for a discovered host" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = xxx")
+    FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = zzz")
     refute find_discovery_rule(host)
   end

   test "rule out of one is found for a discovered host" do
-    r1 = FactoryGirl.create(:discovery_rule, :name => "r1", :priority => 1, :search => "facts.somefact == 'abc'")
-    # XXX Host::Discovered.where(:id => host.id).search_for(rule.search) does not work
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc")
     assert_equal find_discovery_rule(host), r1
   end

+  test "first rule out of two is found for a discovered host" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc")
+    FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc")
+    assert_equal find_discovery_rule(host), r1
+  end
+
+  test "second rule out of two is found for a discovered host" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = x")
+    r2 = FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc")
+    assert_equal find_discovery_rule(host), r2
+  end
+
+  test "drained rule does not match for a discovered host" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc", :max_count => 1)
+    r2 = FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc")
+    FactoryGirl.create(:host, :discovery_rule => r1)
+    assert_equal find_discovery_rule(host), r2
+  end
+
+  test "discovery rule is associated after auto provisioning" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc")
+    perform_auto_provision host.becomes(Host::Managed), r1
+    assert_equal host.managed, true
+    assert_equal host.build, true
+    assert_equal host.hostgroup_id, r1.hostgroup_id
+    assert_equal host.discovery_rule_id, r1.id
+  end
+
+  test "rules with incorrect syntax are skipped" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "=!^$#@?x")
+    r2 = FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc")
+    assert_equal find_discovery_rule(host), r2
+  end
+
+  test "hostname is copied after auto provisioning" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc")
+    perform_auto_provision host.becomes(Host::Managed), r1
+    assert_equal host.name, "macaabbccddeeff"
+  end
+
+  test "hostname is rendered after auto provisioning" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc", :hostname => 'x<%= 1+1 %>')
+    perform_auto_provision host.becomes(Host::Managed), r1
+    assert_equal host.name, "x2"
+  end
+
 end
```

app/controllers/discovered_hosts_controller.rb
lzap, line 126: > Why use a transaction here? What are we trying to protect?

Find and apply must be done in one transaction, otherwise we can
provision something twice.

## 

Later,
 Lukas #lzap Zapletal

app/models/host/managed_extensions.rb
lzap: > > -    true
> 
> Should we return the failure case to the UI so the users knows to go manually reboot the host?

This was a leftover from my previous testing. Removed.

This is a different discussion... ;-)

## 

Later,
 Lukas #lzap Zapletal

GregSutcliffe:
that failure looks weird [test]

test/unit/discovered_extensions_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    assert_equal host.managed, true
+    assert_equal host.build, true
+    assert_equal host.hostgroup_id, r1.hostgroup_id
+    assert_equal host.discovery_rule_id, r1.id
+  end
+
+  test "rules with incorrect syntax are skipped" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "=!^$#@?x")
```
GregSutcliffe: `"=!^$#@?x"` should be single-quoted as `'=!^$#@?x'` - Ruby is interpreting the `#@` and crashing.

lzap:
Pushed! Merry Christmas! :-)

GregSutcliffe:
Merged as d3e7c8b (minus extraneous whitespace :P) - Merry Christmas yourself :)

PR #101, opened 2014-11-19
Fixes #8317 - Implement Reboot action

orrabin:


app/models/host/managed_extensions.rb
```diff
@@ -4,6 +4,7 @@ module Host::ManagedExtensions
   included do
     # execute standard callbacks
     after_validation :queue_reboot
+    require 'concerns/reboot_host'
```
orrabin: this looks wrong to me but when I used `include` it didn't work unless I put this file into the `concerns` folder but when I did that I got an uninitialized error from the `engine.rb`

lzap: Hmmm what if we don't do that as a concern? Simple helper class in the lib/ folder would work, rails should autoload it I believe. This really looks ugly and I can't tell how to do this nicer. Maybe @GregSutcliffe knows better...

To give you the context, Ori is adding new Reboot button which is useful for testing images. The API is there, why not to have the button :-) We want to generalize the method so controllers can use it.

lzap:
And if you can add V2 API reboot command, could be pretty useful when mass upgrading images due to security issue or something (via CLI in a shell loop in the worst case). A unit test would not harm at all too :-) Overall looks great, goint to test once we settle down on the `require` thing.

orrabin:
@lzap if I create a unit test and stub `ProxyAPI::BMC` will that still be a useful test?

lzap:
> @lzap if I create a unit test and stub `ProxyAPI::BMC` will that still be a useful test?

Right, I was thinking more like functional tests. When you click on
Reboot button, something really happens.

## 

Later,
 Lukas #lzap Zapletal

app/models/host/managed_extensions.rb
orrabin: @GregSutcliffe ping?

orrabin:
@lzap I added a test
still waiting for a better solution for the `require`

app/models/host/managed_extensions.rb
GregSutcliffe: Slightly hacky, but you could just make it a class method on Host::Discovered, and then call old.becomes(Host::Discovered).reboot in the Managed orchestration?

orrabin: Sounds better then a concern.
I'll change it

app/models/host/managed_extensions.rb
```diff
@@ -14,14 +14,7 @@ def queue_reboot
...
   def setReboot
-    logger.info "ForemanDiscovery: Rebooting #{name} as its being discovered and assigned"
-    if ::ProxyAPI::BMC.new(:url => "http://#{old.try(:ip) || ip}:8443").power :action => "cycle"
-      logger.info "ForemanDiscovery: reboot result: successful"
-    else
-      logger.info "ForemanDiscovery: reboot result: failed"
-    end
-  rescue => e
-    failure _("Failed to reboot: %s") % e.message
+    old.becomes(Host::Discovered).reboot_host
```
ohadlevy: isnt old always discovered ?

GregSutcliffe: Maybe, it depends how the clone is done. We clone `old` during `HostsController#edit`, but we call `becomes` before loading the managed host edit page, in `DiscoveryController#update`. Worth testing anyway :)

app/models/host/discovered.rb
```diff
@@ -103,6 +102,17 @@ def refresh_facts
     return self.class.import_host_and_facts facts
   end
 
+  def reboot_host
+    logger.info "ForemanDiscovery: Rebooting #{name}"
+    if ::ProxyAPI::BMC.new(:url => "http://#{self.try(:ip) || ip}:8443").power :action => "cycle"
```
GregSutcliffe: Why `try(:ip) || ip` ? Surely they're the same thing?

orrabin: leftovers from the code in `managed_extensions`, I'll leave the `try`

app/models/host/managed_extensions.rb
ohadlevy: @GregSutcliffe then I guess its safest to keep as is?

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -100,6 +100,15 @@ def facts
         render :json => {'message'=>e.to_s}, :status => :unprocessable_entity
       end
 
+      api :POST, "/discovered_hosts/:id/reboot", N_("Rebooting a discovered host")
```
lzap: Nitpick - an infinitive would be better?

config/routes.rb
```diff
@@ -10,6 +10,7 @@
     resources :discovered_hosts do
       member do
         get 'refresh_facts'
+        get 'reboot_host'
```
lzap: For the record, this should be POST as this changes state, but let's just align with refresh_facts.

lzap:
Looks good, I am not able to test it right now.

GregSutcliffe:
@orrabin can you rebase please? In particular, the new permissions block in engine.rb is likely to be a conflict :)

orrabin:
@GregSutcliffe rebased

lzap:
I've tested it and it works great. But I am missing "Reboot" button on the Discovered Host detail page. Since we now have four buttons already, I suggest to merge the following three into a dropdown button (similar to the list page):
- Provision
- Auto-provision
- Refresh facts
- Reboot

Leaving maybe delete as a standalone button.

orrabin:
@lzap I added the reboot button and created a dropdown.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -84,13 +84,21 @@ def update
...
+    if @host.is_a?(::Host::Discovered) && @host.refresh_facts
       process_success :success_msg => _("Facts refreshed for %s") % @host.name, :success_redirect => :back
     else
       process_error :error_msg => _("Failed to refresh facts for %s") % @host.name, :redirect => :back
     end
   end
 
+  def reboot_host
+    if @host.is_a?(::Host::Discovered) && @host.reboot_host
+      process_success :success_msg => _("Rebooting host %s") % @host.name, :success_redirect => discovered_hosts_path
```
orrabin: I think this should always go back to the list because it is also used from the show page.
Should `refresh_facts` also always return to the list?

lzap: Actually, I think it should be `:back` in both cases, because we can now trigger this from both show and list pages? Will this work?

config/routes.rb
```diff
@@ -38,6 +39,7 @@
       resources :discovered_hosts, :except => [:new, :edit] do
         member do
           post 'auto_provision'
+          put 'reboot'
```
lzap, line 42: Why the difference between UI/API `reboot_host` vs `reboot`. IMHO the correct term is just `reboot` for both (as the URL already contains the "host" term once).

app/controllers/discovered_hosts_controller.rb
orrabin: It will stay on the page we triggered it from, in case of the show page it could be outdated if we reboot.
Keeping it as `:back` is consistent with `refresh_facts` so I don't mind changing it if you think it's correct.

config/routes.rb
orrabin, line 42: I had some routing issues at first, I'll change everything to `reboot`

app/models/host/discovered.rb
```diff
@@ -106,6 +105,17 @@ def refresh_facts
...
+  def reboot_host
+    logger.info "ForemanDiscovery: Rebooting #{name}"
+    if ::ProxyAPI::BMC.new(:url => "http://#{self.try(:ip)}:8443").power :action => "cycle"
+      logger.info "ForemanDiscovery: reboot result: successful"
+    else
+      logger.info "ForemanDiscovery: reboot result: failed"
+    end
+  rescue => e
+    raise _("Failed to reboot: %s") % e.message
+  end
```
lzap, line 133: I think this method should always explicitly return `true`. Because when this is triggered from orchestration (`setReboot`) then return value matters. When `false` orchestration fails and everything falls down.

Now the question is if we want to return `false` on failure and I say we should not. Host that does not reboot is not fatal error IMHO, administrator can easily fix this by rebooting it and provisioning will continue. Opinions?

If we agree, I tend to add a flash error message to the `managed_extensions.rb#setreboot` method. We don't need to show anything on succes, but on error administrator will not notice that the host was not rebooted. Actually you can see the progress bar changes the color from green to red but its just fraction of second and then everything seems to be nice but you can only see the error message in the log (and the host is not rebooting). I think a red flash would be good option here.

lzap, line 133: Correction - this method should return true/false depending on the action and the `setReboot` should always return true together with flash message on errors. Not sure what to do with API - maybe we can return 200 but some extra text in body. This is still an open question. @ohadlevy ?

app/controllers/discovered_hosts_controller.rb
lzap: > Keeping it as `:back` is consistent with `refresh_facts` so I don't mind changing it if you think it's correct.

If it works technically then go ahead please. I think it's better to be
consistent.

## 

Later,
 Lukas #lzap Zapletal

app/models/host/discovered.rb
orrabin, line 133: Not sure changing the progress bar will help as it might be too fast to notice if you're not staring at the screen the whole time but a notice message makes sense.
I'm not fixing this (or the change to `reboot`) until the api is determined to prevent your comments from disappearing.

lzap, line 133: > Not sure changing the progress bar will help as it might be too fast to notice if you're not staring at the screen the whole time but a notice message makes sense.

Oh sorry to be misleading. I mean - this is how it works today. I am not
sure if red color for the progress bar was related to the fact that host
was not booting,

Anyway the flash should make this more transparent. Or when @ohadlevy
thinks its better to rollback, then the problem is solved (return
false).

## 

Later,
 Lukas #lzap Zapletal

ohadlevy, line 133: @lzap orchestration will fail on both exception and false, so it does not matter actually.

app/models/host/discovered.rb
```diff
@@ -106,6 +105,18 @@ def refresh_facts
...
+    status = ::ProxyAPI::BMC.new(:url => "http://#{self.try(:ip)}:8443").power :action => "cycle"
+    msg    = status ? 'successful' : 'failed'
+    logger.info "ForemanDiscovery: reboot result: #{msg}"
+    status
+    rescue => e
+      logger.info "ForemanDiscovery: reboot result: #{failed}"
+      logger.warn e.backtrace.join('\n')
+      false
+  end
+
```
lzap, line 134: Is `#{failed}` defined?

Nitpick: formatting.

Looks good, I haven't tested tho.

orrabin, line 134: mistake, it should be `failed`

app/models/host/discovered.rb
```diff
@@ -106,6 +105,18 @@ def refresh_facts
...
   end
 
+  def reboot
+    logger.info "ForemanDiscovery: Rebooting #{name}"
+    status = ::ProxyAPI::BMC.new(:url => "http://#{self.try(:ip)}:8443").power :action => "cycle"
+    msg    = status ? 'successful' : 'failed'
+    logger.info "ForemanDiscovery: reboot result: #{msg}"
+    status
+    rescue => e
+      logger.info "ForemanDiscovery: reboot result: failed"
```
orrabin, line 130: @lzap I didn't understand what needed formatting, can you please explain?

config/routes.rb
```diff
@@ -10,6 +10,7 @@
     resources :discovered_hosts do
       member do
         get 'refresh_facts'
+        get 'reboot'
```
shlomizadok: `put 'reboot'` ?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -7,7 +7,7 @@ class DiscoveredHostsController < ::ApplicationController
   # Avoid auth for discovered host creation
   skip_before_filter :require_login, :require_ssl, :authorize, :verify_authenticity_token, :set_taxonomy, :session_expiry, :update_activity_time, :only => :create
 
-  before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert auto_provision]
+  before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert reboot auto_provision]
```
shlomizadok, line 10: maybe use `:except =>` for the actions we don't want to run this filter, the list here is a bit long (e.g., `:except => [index]`)

app/models/host/discovered.rb
```diff
@@ -106,6 +105,18 @@ def refresh_facts
     return self.class.import_host_and_facts facts
   end
 
+  def reboot
+    logger.info "ForemanDiscovery: Rebooting #{name}"
+    status = ::ProxyAPI::BMC.new(:url => "http://#{self.try(:ip)}:8443").power :action => "cycle"
```
shlomizadok: Is there a chance that `ip` will not exist? Maybe no need to use try? (especially when rescuing later)

shlomizadok:
@orrabin -- overall looking good. I had some minor comments on the code.
One issue, though: I don't seem to have api route to reboot a discovered host. (/api/v2/discovered_hosts/:id/reboot). Also a test for the api would be nice.

app/controllers/discovered_hosts_controller.rb
orrabin, line 10: it would be except: index, create, auto_complete_search, submit_multiple_destroy, multiple_destroy, auto_provision_all
and maybe other actions that will be added so I'm not sure it will be better

app/models/host/discovered.rb
orrabin: I'll fix this, there shouldn't be a case where ip doesn't exist

config/routes.rb
orrabin: already discussed this with @lzap, it should be put but we are being consistent with `refresh_facts`

shlomizadok: Why does it have to be consistent with `refresh_facts`?

app/controllers/discovered_hosts_controller.rb
shlomizadok, line 10: :+1: leave as is... :)

shlomizadok:
Tested and works as expected. @ohadlevy - seems that it's ready to merge.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -84,13 +84,21 @@ def update
...
     else
       process_error :error_msg => _("Failed to refresh facts for %s") % @host.name, :redirect => :back
     end
   end
 
+  def reboot
+    if @host.is_a?(::Host::Discovered) && @host.reboot
+      process_success :success_msg => _("Rebooting host %s") % @host.name, :success_redirect => :back
+    else
+      process_error :error_msg => _("Failed to reboot host %s") % @host.name, :redirect => :back
```
ohadlevy, line 102: do we provide any message on failure? if the response is false, we would not be returning any valuable err msg?

config/routes.rb
ohadlevy: +1 on the question @shlomizadok asked.

orrabin: in the api both use put as they trigger a kind of update, we could change both of them to put here

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -176,13 +185,13 @@ def action_permission
...
             :auto_provision_all
-          when 'refresh_facts'
+          when 'refresh_facts', 'reboot'
             :edit
           else
             super
         end
       end
-
+
```
ohadlevy: nitpick:whitespace

app/controllers/discovered_hosts_controller.rb
```diff
@@ -84,13 +84,27 @@ def update
...
+      process_error :error_msg => _("Host of type %s can not be rebooted") % @host.type, :redirect => :back
+    end
+
+    if @host.reboot
+      process_success :success_msg => _("Rebooting host %s") % @host.name, :success_redirect => :back
+    else
+      process_error :error_msg => _("Failed to reboot host %s") % @host.name, :redirect => :back
+    end
+    rescue  => e
+      process_error :error_msg => _("Failed to reboot host %s with error %s") % [@host.name, e.message], :redirect => :back
```
shlomizadok: Use hash of parameters instead of interpolation (e.g., `_("Failed to reboot host %{hostname} with error %{message}") % {:hostname => @host.name, :message => e.message}`

app/models/host/discovered.rb
```diff
@@ -112,6 +113,18 @@ def refresh_facts
     return self.class.import_host_and_facts facts
   end
 
+  def reboot
+    logger.info "ForemanDiscovery: Rebooting #{name}"
+    status = ::ProxyAPI::BMC.new(:url => proxy_url).power :action => "cycle"
```
shlomizadok: This will not work on smart-proxy discovery plugin. Smart-proxy discovery plugin expects to get a url `<proxy_url>/discovery/:ip/reboot`

orrabin: @shlomizadok so I can remove the `power :action => "cycle"` part?

shlomizadok: @orrabin Only if the reboot is done via smart-proxy.
If it's directly to the host you need to use the code above (`::ProxyAPI::BMC.new(:url => proxy_url).power :action => "cycle"`). If it's through the smart proxy `PUT <proxy_url>/discovery/:ip/reboot`

app/controllers/discovered_hosts_controller.rb
lzap: Right more info here - http://projects.theforeman.org/projects/foreman/wiki/Translating#Exceptions

lzap:
I will retest once the exception is corrected and the parameter thing is solved.

Did we agree on allowing directo communication optionally or did we agree on proxy as mandatory component?

orrabin:
@lzap we decided to start using proxy only but to keep support in discovery in case someone is using an older image

lib/foreman_discovery/proxy_operations.rb
```diff
@@ -0,0 +1,17 @@
+module ForemanDiscovery
+  class ProxyOperations < ::ProxyAPI::Resource
+    def initialize args
+      @url  = args[:url] + "/#{args[:operation]}"
+      super args
+    end
+
+    def parse_put_operation
+      # parse put did not work from here, until problem is fixed using rest client directly 
+      RestClient.put @url, {}, :content_type => :json, :accept => :json
```
orrabin: does anyone use parse put?
maybe I could try to figure it out from a usage example

lzap:
Tested, rebased. Thanks, works as expected. Merged. c081812..7ac42f1

PR #103, opened 2014-11-20
Fixes #8422 - primary detection for non-discovered

ares:
We injected discovery way to chose primary interface to puppet fact
parser that is used also for not discovered host which are missing
discovery specific facts. Mac for such value was evaluated to nil
which matched also macaddress_lo so lo was chosen as primary
interface.

### DIFF
--- a/app/lib/puppet_fact_parser_extensions.rb
+++ b/app/lib/puppet_fact_parser_extensions.rb
@@ -8,16 +8,18 @@ module PuppetFactParserExtensions
 
   # we prefer discovery_bootif fact to choose right primary interface (interface used to boot the image)
   def primary_interface_with_discovery_fact
-    mac = facts[discovery_mac_fact_name]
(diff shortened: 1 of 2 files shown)


domcleal:
A test would be good to prevent regressions, as this is modifying core.

ares:
I added tests and fixed the same error in primary interface detection (also covered by tests)

domcleal:
Test failures appear to be due to https://github.com/theforeman/foreman/commit/eaefbecc2052dfea1805b4b7119ae124f823149c, which I logged against core as http://projects.theforeman.org/issues/8456.

domcleal:
[test]

domcleal:
Tests are now passing.  Looks correct I think :+1:

mmoll:
tested, fixes the bug for me :+1:

GregSutcliffe:
Merged as 83ee2e2, thanks @ares !

PR #105, opened 2014-11-27
Fixes #8514 - Add support for requests to smart proxy discovery plugin

orrabin:


app/models/host/discovered.rb
```diff
@@ -94,8 +94,13 @@ def root_pass
   def refresh_facts
     # TODO: Can we rely on self.ip? The lease might expire/change....
     begin
-      logger.debug "retrieving facts from proxy on ip: #{self.ip}"
-      facts = ForemanDiscovery::Facts.new(:url => "http://#{self.ip}:8443").facts
+      proxy_url = "http://#{self.ip}:8443"
+      unless self.subnet.nil? || self.subnet.discovery_id.nil?
```
ohadlevy: nitpick: self is not needed and you can also write

``` ruby
if subnet.try(:discovery_id)
```

ohadlevy: i would simply write it as:

``` ruby
proxy_url = if subnet.try(:discovery_id)
  SmartProxy.find(self.subnet.discovery_id).url + "/discovery/#{self.ip}"
else
  "http://#{self.ip}:8443"
end
```

db/migrate/20141126165451_add_discovery_id_to_subnet.rb
```diff
@@ -0,0 +1,9 @@
+class AddDiscoveryIdToSubnet < ActiveRecord::Migration
+  def self.up
+    add_column :subnets, :discovery_id, :integer
```
ohadlevy, line 3: should we add relationship on the subnet model? this would allow subnet.discovery_proxy.url or similar?

orrabin, line 3: @ohadlevy thanks, I've added the relationship and it made the rest of the changes much more readable

orrabin:
@shlomizadok can you please check this still works well with your plugin?

app/models/host/discovered.rb
```diff
@@ -94,8 +94,14 @@ def root_pass
   def refresh_facts
     # TODO: Can we rely on self.ip? The lease might expire/change....
     begin
-      logger.debug "retrieving facts from proxy on ip: #{self.ip}"
-      facts = ForemanDiscovery::Facts.new(:url => "http://#{self.ip}:8443").facts
+      proxy_url = if subnet.try(:discovery_id)
+                      subnet.discovery.url + "/discovery/#{self.ip}"
```
ohadlevy, line 101: nitpick: not sure, but maybe its more readable if its discovery_proxy.url ?

orrabin, line 101: I don't mind changing it.
I just wanted the `belongs_to` to look like the rest of them in subnet:

``` ruby
 belongs_to :dhcp, :class_name => "SmartProxy"
 belongs_to :tftp, :class_name => "SmartProxy"
 belongs_to :dns,  :class_name => "SmartProxy"
```

shlomizadok:
@orrabin works as expected.

foreman_discovery.gemspec
```diff
@@ -19,4 +19,6 @@ Gem::Specification.new do |s|
   s.homepage = %q{http://github.com/theforeman/foreman_discovery}
   s.licenses = ["GPL-3"]
   s.summary = %q{MaaS Discovery Plugin for Foreman}
+
+  s.add_dependency 'deface', '< 1.0'
```
GregSutcliffe, line 23: Sigh, I only just managed to get rid of deface from discovery :)

app/models/host/discovered.rb
GregSutcliffe, line 101: I'm fine with `:discovery`

app/models/host/discovered.rb
```diff
@@ -94,8 +94,14 @@ def root_pass
   def refresh_facts
     # TODO: Can we rely on self.ip? The lease might expire/change....
     begin
-      logger.debug "retrieving facts from proxy on ip: #{self.ip}"
-      facts = ForemanDiscovery::Facts.new(:url => "http://#{self.ip}:8443").facts
+      proxy_url = if subnet.try(:discovery_id)
```
GregSutcliffe: Should this be `if subnet.discovery.present?` ?

app/models/host/discovered.rb
```diff
@@ -94,8 +94,14 @@ def root_pass
...
+                      subnet.discovery.url + "/discovery/#{self.ip}"
+                  else
+                      "http://#{self.ip}:8443"
+                  end
+
+      logger.debug "retrieving facts from proxy from url: #{proxy_url}"
+      facts = ForemanDiscovery::Facts.new(:url => proxy_url).facts
     rescue Exception => e
       raise _("Could not get facts from proxy: %s") % e
```
lzap: I'd definitely add the proxy url to the error message as well. We will see many of these.

app/overrides/subnet_form_with_discovery_proxy.rb
```diff
@@ -0,0 +1,9 @@
+selector_text = " <%= select_f f, :discovery_id, SmartProxy.with_features('Discovery'), :id, :name,
+                    {:include_blank => _('None')},
+                    :label => _('Discovery Proxy'),
+                    :help_inline => _('Discovery Proxy to use within this subnet for managing connection to discovered hosts') %>"
+
+Deface::Override.new(:virtual_path  => "subnets/_form",
+                     :name          => "add_discovery_proxy_to_subnet",
+                     :insert_bottom => 'div#proxies',
+                     :text          =>  selector_text)
```
lzap, line 9: Does it make sense to add this into the core to get rid of the deface @ohadlevy ?

Where do we add a new "Discovery" feature? I don't see this in our patch.

foreman_discovery.gemspec
lzap, line 23: @ohadlevy please explicitly confirm you want deface here, because for example DHCP is also a proxy plugin now, you can disable it the same way while the field on the subnet stays. With discovery we could even identify on that screen if plugin was loaded (and hide the option if not). This is also not perfect, but less deface the better I think :-)

lzap, line 23: I think I was not clear in the previous comment - my thinking was to bring this into the core and only show it when plugin is present.

app/models/concerns/discovery_subnet.rb
```diff
@@ -0,0 +1,9 @@
+module Concerns
+  module DiscoverySubnet
+    extend ActiveSupport::Concern
+    included do
+      belongs_to :discovery, :class_name => "SmartProxy"
+    end
+  end
+end
+
```
lzap, line 7: Nitpick - the existing extension does not introduce a "Concerns" namespace. Ideally, it should be in the own namespace. To be refactored later...

app/models/host/discovered.rb
lzap, line 101: Me too, maybe it will not be only proxy in future.

app/models/concerns/discovery_subnet.rb
orrabin, line 7: @lzap when I remove the "Concerns" namespace and change it in the engine.rb I get:
`uninitialized constant DiscoverySubnet (NameError)`

lzap, line 7: > @lzap when I remove the "Concerns" namespace and change it in the engine.rb I get:
> `uninitialized constant DiscoverySubnet (NameError)`

Oh right, I have added concerns subdir to autoload paths in my own patch
that was not yet merged. Let's leave this as is and we can refactor
later on.

## 

Later,
 Lukas #lzap Zapletal

app/overrides/subnet_form_with_discovery_proxy.rb
GregSutcliffe, line 9: I doubt we'd get that merged into core as-is. We had a similar discussion in the template-proxying PR, and decided it didn't make sense to have code in core that depended on plugins - it breaks the whole point of extensibility.

We _could_ get a patch into core that makes it easier to extend the subnet from a plugin without deface, but thats a whole bigger thing about extensible form builders :)

app/models/concerns/discovery_subnet.rb
GregSutcliffe, line 7: That was merged, so you can refactor now :)

app/overrides/subnet_form_with_discovery_proxy.rb
GregSutcliffe, line 9: For myself, I agree about not wanting to re-introduce deface to the plugin. However, we have little choice - the 'discovery' feature will be present on the proxy, and Foreman needs to know how to handle that. I see that foreman-salt had to do a similar thing in adding the salt_proxy_id to the host and hostgroup forms. As such, I don't think we can get around this :)

orrabin:
@GregSutcliffe refactored and rebased

foreman_discovery.gemspec
ohadlevy, line 23: @lzap no issue from my side, as we won't be removing deface any time soon.

I agree, that we should reconsider tftp/dhcp/dns etc outside the context of this pull request :)

ohadlevy:
merged as 55571c0 thanks @orrabin

PR #109, opened 2014-12-10
Fixes #8649 - Register to Apipie via plugin methods

mbacovsky:
The purpose of this is to be able to pre-build the Apipie docs during the package build and save some time during installation, when we have to build just the indexes.
There are some packaging changes to follow...


lzap:
Thanks, will rebase and merge next week.

lzap:
Pushed as f1f29ca, thanks Martin!

PR #112, opened 2014-12-21
Fixes #8723 - Adds CPU, mem and disk size to discovered hosts index page

orrabin:


unorthodoxgeek:
can you please add a few tests for the expected behaviour?

orrabin:
@unorthodoxgeek we usually don't add tests for which columns are present in the UI.
The expected behavior is the columns are there, can be sorted and searched by. 
I could add a test for the search part.

unorthodoxgeek:
@orrabin - because that score search is not the straight forward way of using scope search, I think the search itself should be tested. I agree that there's no need to do UI tests, but we should unit test it :smile:

orrabin:
@unorthodoxgeek agreed, I'll add tests and also add the columns to the the api to be consistent with the UI. After that this can be reviewed and tested.

app/models/host/discovered.rb
```diff
@@ -29,6 +33,22 @@ class Host::Discovered < ::Host::Base
     where(conditions)
   }
 
+  def self.search_by_fact_value(key, operator, value)
+    case key
+      when 'cpu'
```
ohadlevy: nitpick, put the fact_name before the case, so you dont have to RY all the time.

app/models/host/discovered.rb
```diff
@@ -29,6 +33,22 @@ class Host::Discovered < ::Host::Base
...
+        fact_name = 'physicalprocessorcount'
+      when 'memory'
+        fact_name = 'memorysize_mb'
+      when 'disk_size'
+        fact_name = 'blockdevice_sda_size'
+    end
+
+    fact_id = FactName.where(:name => fact_name).first.id
+    conditions = sanitize_sql_for_conditions(["fact_values.value #{operator} ?",  value_to_sql(operator, value)])
+    host_ids = FactValue.where("fact_values.fact_name_id = #{fact_id}").where(conditions).map(&:host_id) + [-1]
```
ohadlevy: -1 ?

you can probably simplify this query by avoiding map etc and using inner select instead?

app/models/host/discovered.rb
```diff
@@ -29,6 +33,22 @@ class Host::Discovered < ::Host::Base
...
+      when 'cpu'
+        fact_name = 'physicalprocessorcount'
+      when 'memory'
+        fact_name = 'memorysize_mb'
+      when 'disk_size'
+        fact_name = 'blockdevice_sda_size'
+    end
+
+    fact_id = FactName.where(:name => fact_name).first.id
+    conditions = sanitize_sql_for_conditions(["fact_values.value #{operator} ?",  value_to_sql(operator, value)])
```
ohadlevy: does this work only for equality ? i assume != and >< bigger smaller then wont work here at all? (which might be important)

app/models/host/discovered.rb
```diff
@@ -110,6 +130,18 @@ def self.model_name
     ActiveModel::Name.new(Host)
   end
 
+  def cpu
+    facts_hash['physicalprocessorcount']
```
ohadlevy: DRY: extract eh first method to another method and use it here too?

app/models/host/discovered.rb
orrabin: they all work but auto complete shows !=, =, ~ but not ><

app/models/host/discovered.rb
orrabin: it was added in case of an empty result because of the next line but I'll try using an inner select and then it won't be needed

orrabin:
@ohadlevy I fixed your comments
@unorthodoxgeek there are now tests for the search on the new columns

app/models/host/discovered.rb
```diff
@@ -29,6 +33,25 @@ class Host::Discovered < ::Host::Base
...
   }
 
+  def self.facts_for_search(key)
+    case key
+      when 'cpu'
+        'physicalprocessorcount'
+      when 'memory'
+        'memorysize_mb'
+      when 'disk_size'
+        'blockdevice_sda_size'
```
ohadlevy: what happens if there is more than one disk ?

ohadlevy: also, note that sda is very opinionated, imho you should loop over all blockdevices... and potentially filter out non relevant ones (e.g. I assume cdrom does not get shown if such exists etc)

app/models/host/discovered.rb
```diff
@@ -29,6 +33,25 @@ class Host::Discovered < ::Host::Base
...
+    end
+  end
+
+  def self.search_by_fact_value(key, operator, value)
+    fact_name = facts_for_search(key)
+    fact_id = FactName.where(:name => fact_name).first.id
+    conditions = sanitize_sql_for_conditions(["fact_values.value #{operator} ?",  value_to_sql(operator, value)])
+    host_ids = FactValue.where("fact_values.fact_name_id = #{fact_id}").where(conditions).map(&:host_id) + [-1]
```
ohadlevy: maybe use pluck instead of map?
also, why did you give up on the inner query?

app/views/api/v2/discovered_hosts/main.json.rabl
```diff
@@ -2,7 +2,7 @@ object @discovered_host
 
 extends "api/v2/discovered_hosts/base"
 
-attributes :ip, :mac, :last_report, :subnet_id, :subnet_name
+attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :cpu, :memory, :disk_size
```
ohadlevy: cpu is misleading, are you returning the amount of cpus, or the type of cpu?

app/models/host/discovered.rb
orrabin: so what should disk size show? the total size? the sizes of each blockdevice?

app/models/host/discovered.rb
orrabin: I had a problem with sort and it looked like it happened after I changed to the inner query buy now it's happening in both cases so I'll change this

orrabin:
@ohadlevy 2 commits for now to make it easier to go back to the first solution if needed.

app/models/discovery_attribute.rb
```diff
@@ -0,0 +1,4 @@
+class DiscoveryAttribute < ActiveRecord::Base
+  belongs_to :discovery
```
ohadlevy: I'm guessing this doesnt work, as discovery is not a class?

app/models/host/discovered.rb
```diff
@@ -1,11 +1,13 @@
 require 'foreman_discovery/facts'
 
 class Host::Discovered < ::Host::Base
+  include ScopedSearchExtensions
 
   belongs_to :location
   belongs_to :organization
   belongs_to :subnet
   belongs_to :hostgroup
+  has_one    :discovery_attribute, :foreign_key => :discovery_id
```
ohadlevy: i wonder if attribute_set is a better name, but I assume we can leave it as is.

app/models/host/discovered.rb
```diff
@@ -19,6 +21,9 @@ class Host::Discovered < ::Host::Base
   scoped_search :in => :location, :on => :name, :rename => :location, :complete_value => true         if SETTINGS[:locations_enabled]
   scoped_search :in => :organization, :on => :name, :rename => :organization, :complete_value => true if SETTINGS[:organizations_enabled]
   scoped_search :in => :subnet, :on => :network, :complete_value => true, :rename => :subnet
+  scoped_search :in => :discovery_attribute, :on => :cpu_count, :rename => :cpu_count, :complete_value => true, :only_explicit => true
+  scoped_search :in => :discovery_attribute, :on => :memory, :rename => :memory, :complete_value => true, :only_explicit => true
+  scoped_search :in => :discovery_attribute, :on => :disk_count, :rename => :disk_count, :complete_value => true, :only_explicit => true
```
ohadlevy: missing disks size?

app/models/host/discovered.rb
```diff
@@ -74,6 +80,17 @@ def import_facts facts
     super
   end
 
+  def add_discovery_attribute
+    cpu_count = facts_hash['physicalprocessorcount']
+    memory = facts_hash['memorysize_mb']
+    disks = facts_hash.select{ |key, value| key.to_s =~ /blockdevice.*_size/ }
+    disks_size = 0
+    disks.each { |name, size| disks_size += size.to_i}
```
ohadlevy: can also do

``` ruby
disks.values{|size| disks_size += size.to_i)
```

I would also make sure that disks is not empty?

app/models/host/discovered.rb
```diff
@@ -74,6 +80,17 @@ def import_facts facts
...
+
+    self.discovery_attribute = DiscoveryAttribute.new(:cpu_count => cpu_count.to_i, :memory => memory.to_i,
```
ohadlevy: hmm.. does new actually save it?

db/migrate/20141223101707_create_discovery_attributes.rb
```diff
@@ -0,0 +1,14 @@
+class CreateDiscoveryAttributes < ActiveRecord::Migration
+  def change
+    create_table :discovery_attributes do |t|
+      t.references :discovery
+      t.integer :memory
+      t.integer :cpu_count
+      t.integer :disk_count
```
ohadlevy: should we add default db values?
what about FK

app/models/host/discovered.rb
ohadlevy: dependent destroy?

test/factories/discovery_attributes.yml
```diff
@@ -0,0 +1,15 @@
+# Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/Fixtures.html
```
ohadlevy: change to factory?

db/migrate/20141223101707_create_discovery_attributes.rb
orrabin: we could default everything to 0 but maybe just seeing that the value is missing is more informative?

app/models/host/discovered.rb
orrabin: it did save, there was a problem with new hosts that `facts_hash` didn't exist so new hosts didn't get the correct attributes but I don't mind changing this to `create` to make sure it always save correctly

app/models/discovery_attribute.rb
ohadlevy: missing validation?

test/unit/discovery_attribute_set_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  test "can search discovered hosts by disks_size" do
+    raw = parse_json_fixture('/facts.json')
+    host = Host::Discovered.import_host_and_facts(raw['facts']).first
+    results = Host::Discovered.search_for("disks_size = #{host.facts_hash['blockdevice_sda_size'].to_f}")
+    assert_equal 1, results.count
+    results = Host::Discovered.search_for("disks_size > #{host.facts_hash['blockdevice_sda_size'].to_f}")
+    assert_equal 0, results.count
+  end
+
+  def parse_json_fixture(relative_path)
```
orrabin, line 40: this needs to move, maybe to `test_helper`? so it can be used in both tests

orrabin:
locally tests run fine with mysql so try to [test] again

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -81,7 +81,12 @@ def update
         @host.managed = true
         @host.build   = true
         forward_request_url
-        process_response @host.update_attributes(params[:discovered_host])
+        update_response = @host.update_attributes(params[:discovered_host])
+        if (update_response)
+          discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => @host.id)
```
ohadlevy: @host.discovery_attribute_set.delete_all ?

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -81,7 +81,12 @@ def update
         @host.managed = true
         @host.build   = true
         forward_request_url
-        process_response @host.update_attributes(params[:discovered_host])
+        update_response = @host.update_attributes(params[:discovered_host])
+        if (update_response)
```
ohadlevy: nitpick: parenthesis not required?

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -36,7 +36,10 @@ def perform_auto_provision host, rule
     host.hostgroup_id = rule.hostgroup_id
     host.comment = "Auto-discovered and provisioned via rule '#{rule.name}'"
     host.discovery_rule = rule
-    host.save # save! does not work here
+    if (host.save) # save! does not work here
```
ohadlevy: nitpick: parenthesis not required?

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -36,7 +36,10 @@ def perform_auto_provision host, rule
...
+      discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => host.id)
+      discovery_attribute_set.destroy_all unless discovery_attribute_set.nil?
```
ohadlevy: ditto: use rails methods?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -73,6 +73,8 @@ def update
     forward_url_options
     Taxonomy.no_taxonomy_scope do
       if @host.update_attributes(params[:host])
+        discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => @host.id)
```
ohadlevy: since this repeats 3 times already, cant this be a callback in the model instead?

app/models/discovery_attribute_set.rb
```diff
@@ -0,0 +1,22 @@
+class DiscoveryAttributeSet < ActiveRecord::Base
+  belongs_to :host, :class_name => "Host::Discovered", :foreign_key => :host_id
+  attr_accessible :cpu_count, :disk_count, :disks_size, :memory
+
+  validates :cpu_count, :presence => true
```
ohadlevy: nitpick, unify into one line

app/models/host/discovered.rb
```diff
@@ -86,6 +92,9 @@ def populate_fields_from_facts facts = self.facts_hash, type = 'puppet'
       importer = super(facts)
     end
     self.subnet = Subnet.subnet_for(importer.ip)
+    atribute_set = DiscoveryAttributeSet.new()
+    atribute_set.add_discovery_attribute_set(self)
+    self.discovery_attribute_set = atribute_set if atribute_set.valid?
```
ohadlevy: if its not valid, would it silently skip the attributes?

app/models/discovery_attribute_set.rb
unorthodoxgeek: I don't agree. declarative stuff like validates should not declare multiple columns, it makes reading harder later on IMHO.

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -22,6 +27,10 @@
       </td>
       <td><%= link_to("#{host}", discovered_host_path(host) ) %> </td>
       <td class="hidden-tablet hidden-xs"><%= model_name host %></td>
+      <td class="hidden-tablet hidden-xs"><%= host.discovery_attribute_set.cpu_count %></td>
```
ohadlevy: i think we could delegate all of the methods to discovery_attribute_set, e.g.
in the discovery host model

``` ruby
delegate :cpu_count, :memory .., :to => discovery_attribute_set
```

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -22,6 +27,10 @@
...
+      <td class="hidden-tablet hidden-xs"><%= number_to_human_size(host.discovery_attribute_set.memory * 1024 * 1024) %></td>
```
ohadlevy: since we dont save DAS if its not valid, it might not be accessible here?

app/models/discovery_attribute_set.rb
ohadlevy: @unorthodoxgeek in second thought, i agree, and probably should extend the validation to include numerical validations (e.g. cant be negative's etc).

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -36,7 +36,16 @@ def perform_auto_provision host, rule
     host.hostgroup_id = rule.hostgroup_id
     host.comment = "Auto-discovered and provisioned via rule '#{rule.name}'"
     host.discovery_rule = rule
-    host.save # save! does not work here
+    if host.save #save! does not work here
+      delete_discovery_attribute_set(host.id)
+    end
+  end
+
+  # discovery_attribute_set should be deleted only after a host has been successfully provisioned
```
ohadlevy, line 47: i guess you mean changed from host discovered to host managed class?

orrabin, line 47: yes, I'v added a sentence to explain

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -36,7 +37,17 @@ def perform_auto_provision host, rule
...
+  # (after discovered host becomes managed host)
+  # after this happens the host is a manged host so non of the callbacks in discovered hosts are relevant anymore
+  # for this reason the orphaned discovery_attribute_set needs to be searched for and deleted
+  def delete_discovery_attribute_set(host_id)
+    DiscoveryAttributeSet.delete_all(:host_id => host_id)
```
ohadlevy, line 52: nitpick: would it be better to pass the discovery attribute set id instead of the host id?

ohadlevy, line 52: probably not... feel free to ignore :)

app/models/host/discovered.rb
```diff
@@ -86,9 +95,27 @@ def populate_fields_from_facts facts = self.facts_hash, type = 'puppet'
       importer = super(facts)
     end
     self.subnet = Subnet.subnet_for(importer.ip)
+    self.discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => id).first_or_create()
+    self.discovery_attribute_set.update_attributes(import_from_facts())
```
ohadlevy: nitpick: no need for parenthesis for a method with no arguments.

app/views/dashboard/_discovery_widget.html.erb
```diff
@@ -2,15 +2,30 @@
 <% all_hosts = Host::Discovered.search_for '' %>
 <% hosts = all_hosts.limit(5) %>
 <% if hosts.empty? %>
-  <p class="ca"><%= _("No discovered hosts available") %></p>
+    <p class="ca"><%= _("No discovered hosts available") %></p>
```
ohadlevy: nitpick: whitespace

app/views/dashboard/_discovery_widget.html.erb
```diff
@@ -2,15 +2,30 @@
...
+          <th>ip</th>
+          <th>cpu count</th>
+          <th>memory</th>
+          <th>disk count</th>
+          <th>disks size</th>
+      </thead>
+      <tbody>
+      <% hosts.each do |host| %>
+          <tr>
+            <td><%= link_to h(host.name), discovered_host_path(host) %></td>
```
ohadlevy: extract to partial?

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -1,10 +1,15 @@
 <%= javascript "jquery.cookie", "host_checkbox" %>
+<% require 'action_view' %>
```
ohadlevy: ?

app/models/host/discovered.rb
unorthodoxgeek: Not nitpick. This violates ruby coding standards.

app/views/discovered_hosts/_discovered_hosts_list.html.erb
orrabin: for the `number_to_human_size` helper

app/models/host/discovered.rb
orrabin: leftovers from refactoring the method to not receive arguments, I'll remove them.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -29,6 +29,7 @@ def find_discovery_rule host
 
   # trigger the provisioning
   def perform_auto_provision host, rule
+    host = host.becomes(::Host::Managed)
```
lzap, line 32: Is this really needed? There were some hidden dragons and if we can work with the generic instance, I'd rather stick with it.

app/models/discovery_attribute_set.rb
```diff
@@ -0,0 +1,8 @@
+class DiscoveryAttributeSet < ActiveRecord::Base
+  belongs_to :host, :class_name => "Host::Discovered", :foreign_key => :host_id
+  attr_accessible :cpu_count, :disk_count, :disks_size, :memory
+
+  validates :cpu_count, :presence => true, :numericality => {:greater_than_or_equal_to => 0}
+  validates :memory, :presence => true,    :numericality => {:greater_than_or_equal_to => 0}
+  validates :host, :presence => true
+end
```
lzap, line 8: I was not following your discussion about introducing new columns and I take it it is not possible to join against facts table. With Facter 2.0 memory (and cpu_count) are reported as numbers instead of strings. Just thinking loud.

app/models/host/discovered.rb
```diff
@@ -19,6 +24,10 @@ class Host::Discovered < ::Host::Base
   scoped_search :in => :location, :on => :name, :rename => :location, :complete_value => true         if SETTINGS[:locations_enabled]
   scoped_search :in => :organization, :on => :name, :rename => :organization, :complete_value => true if SETTINGS[:organizations_enabled]
   scoped_search :in => :subnet, :on => :network, :complete_value => true, :rename => :subnet
+  scoped_search :in => :discovery_attribute_set, :on => :cpu_count, :rename => :cpu_count, :complete_value => true, :only_explicit => true
+  scoped_search :in => :discovery_attribute_set, :on => :memory, :rename => :memory, :complete_value => true, :only_explicit => true
+  scoped_search :in => :discovery_attribute_set, :on => :disk_count, :rename => :disk_count, :complete_value => true, :only_explicit => true
+  scoped_search :in => :discovery_attribute_set, :on => :disks_size, :rename => :disks_size, :complete_value => true, :only_explicit => true
```
lzap, line 30: Won't this be confusing: host.memory vs host.facts.memorysize ? I mean for end users. We might want to filter out irrelevant facts from results.

test/unit/discovered_extensions_test.rb
```diff
@@ -99,7 +99,7 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     facts = @facts.merge({"somefact" => "abc"})
     host = Host::Discovered.import_host_and_facts(facts).first
     r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc", :hostname => 'x<%= 1+1 %>')
-    perform_auto_provision host.becomes(Host::Managed), r1
+    perform_auto_provision host, r1
     assert_equal host.name, "x2"
   end
```
lzap, line 104: Oh I see why to change the superclass, I agree. +1

lzap:
Works fine here, nice. How to prevent our users from using facts.memorysize instead of memory etc?

app/models/discovery_attribute_set.rb
orrabin, line 8: core currently turns them into string anyway and the facts table only accepts  strings so for now it wont help

app/models/host/discovered.rb
orrabin, line 30: @lzap I can change it to memory_size, is there any others that need changing to make this less confusing?

app/controllers/concerns/foreman/controller/discovered_extensions.rb
orrabin, line 32: @lzap I might be mistaken but from what I saw every place that called `perform_auto_provision` did it's own `becomes` so I moved it here and removed it from the other method calls.
Is there somewhere that uses auto provision and passes the host as is?

orrabin, line 32: @lzap saw you're other comment now, disregard this

app/views/discovered_hosts/_discovered_hosts_list.html.erb
ohadlevy: afaik this is not needed, you are already in a view context?

app/models/host/discovered.rb
lzap, line 30: > @lzap I can change it to memory_size, is there any others that need changing to make this less confusing?

I think `memory` is totally fine. By the confusion I mean that user
simple enters `facts.memorysize > 8000` or similar which won't work. I
was thinking maybe to delete that fact, or mask it somehow?

Any ideas?

## 

Later,
 Lukas #lzap Zapletal

app/controllers/concerns/foreman/controller/discovered_extensions.rb
lzap, line 32: > @lzap I might be mistaken but from what I saw every place that called `perform_auto_provision` did it's own `becomes` so I moved it here and removed it from the other method calls.
> Is there somewhere that uses auto provision and passes the host as is?

Yeah, I realized. This stems from refactoring we did because of some
nasty bugs. But we have that covered now. It's okay.

## 

Later,
 Lukas #lzap Zapletal

orrabin:
@ohadlevy I added transactions around delete attributes

app/models/host/discovered.rb
```diff
@@ -86,9 +95,27 @@ def populate_fields_from_facts facts = self.facts_hash, type = 'puppet'
       importer = super(facts)
     end
     self.subnet = Subnet.subnet_for(importer.ip)
+    self.discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => id).first_or_create
+    self.discovery_attribute_set.update_attributes(import_from_facts)
     self.save
   end
 
+  def import_from_facts(facts = self.facts_hash)
+    cpu_count  = facts['physicalprocessorcount'].to_i
```
lzap, line 104: One last nitpick - can you add a rescue branch just in case someone sends wrong input?

```
cpu_count  = facts['physicalprocessorcount'].to_i rescue 0
```

The same for memory and perhaps in the each block for the disks size. Converting to numbers can fail hard.

lzap, line 104: Forget it, I will do that that is faster. Merging!

lzap:
Thank you, outstanding improvement. This makes auto-provisioning much more useful. 181d2b0..cd560e5

PR #113, opened 2015-01-06
Fixes #8831 - Turn discovered hosts show form to two-pane

orrabin:


app/views/discovered_hosts/show.html.erb
```diff
@@ -1,16 +1,17 @@
-<% title(_("Discovered host: %s") % @host.to_label) %>
-
-<%= discovered_hosts_title_actions(@host) %>
```
orrabin, line 3: should the action buttons still exist in show?
they are all present in the index

app/assets/javascripts/discovery-two-pane.js
```diff
@@ -0,0 +1,7 @@
+$(document).on('click', ".table-two-pane td[class$='display-two-pane'] a", function(e) {
```
orrabin: This can be removed after https://github.com/theforeman/foreman/pull/2048 is merged.
It's possible to merge this PR and create a refactoring issue to fix it instead of waiting for the change in core.

app/assets/javascripts/discovery-two-pane.js
```diff
@@ -0,0 +1,7 @@
...
+    debugger
```
ohadlevy: debugger

orrabin:
@GregSutcliffe, @lzap, @shlomizadok this is an initial version just so you can see how discovery looks with two-pane and give your opinion on - should there be action buttons? what else is needed besides the facts? maybe the facts table should look a little different?

app/assets/javascripts/discovery-two-pane.js
orrabin: oops :)

app/assets/javascripts/discovery-two-pane.js
orrabin: if this is going to be merged before the change in core then the js needs to be added to the precompiled assets like: https://github.com/theforeman/foreman_discovery/pull/118/files#diff-9c896d7540d8746964738086b33782b7R25

lzap:
I merged the core patch, please remove the CSS.

ohadlevy:
@lzap, note that discovery needs to work with 1.7-stable, hence the core
patch does not matter much..

lzap:
You are right. Ok testing now, @orrabin if you can create a refactoring ticket.

orrabin:
@lzap http://projects.theforeman.org/issues/8967

lzap:
It looks like there is overflow issue thank to the ssh keys which are long facts:

lzap:
From the UX perspective, I think we need the buttons on the page. We can put them to the very top of the pane:
- Provision (default button)
- Auto Provision
- Refresh Facts / Reboot (one dropdown button)
- Delete (red)
- Cancel

What you think?

orrabin:
@lzap sounds good, I'll add the buttons and fix the overflow, thanks for testing!

orrabin:
@lzap fixed the long value problem and added all of the buttons except Cancel since there is the `x` to close the form and it's show and not edit so I think it should be enough.

lzap:
Does work okay, but can you add some padding between the buttons?

Also it could be aligned to the table from the left.

orrabin:
@lzap fixed your comments, please review again

lzap:
It looks great. Nitpick: is it possible to return to the same page detail when doing Reboot/Refresh? I am not sure if this is technically possible (`:back`?). Otherwise :+1:

lzap:
Currently it goes back to the discovered hosts list which is acceptable.

orrabin:
@lzap `:back` is there now, I also tried to redirect to `discovered_host_path` with the host and it didn't work, if you have an idea I'll gladly change it

app/assets/javascripts/discovery-two-pane.js
```diff
@@ -0,0 +1,6 @@
+$(document).on('click', ".table-two-pane .display-two-pane a", function(e) {
+    if ($('.table-two-pane').length) {
```
ohadlevy: nitpick: two space ident

lzap:
Ok it works fine here, please confirm I can merge. I am a bit confused with the amended changes right now. :-)

lzap:
_For the record_: Some show/edit two-pane links will trigger twice in particular Foreman-Discovery version combinations. We have the refactoring ticket to remove the temporary javascript from this patch for Foreman 1.8 version.

lzap:
Merged as 7a6454704cf87f7 thank you @orrabin for your contribution.

PR #115, opened 2015-01-11
Fixes #8899 - Add requirement for id field in auto_provision for api docs

orrabin:


lzap:
Thank you @orrabin - merged as 71a5f79

PR #116, opened 2015-01-12
Fixes #8900 - disks_size is bigger to fix migration on pg

orrabin:


ohadlevy:
What about memory size?

Also, should we create a new migration instead of modifying an existing one?

orrabin:
@ohadlevy not sure memory_size of one host will be too big but I'll change it.
If I add a migration it will try to run after the existing failing migration which is why I changed the existing one.
I can add a new migration and switch between the order of the new and failing migration if that's better.

lzap:
Well `cpu_count` can theoreticaly ... </joke>

I actually prefer what @orrabin did because this requires us ( @GregSutcliffe @ohadlevy @orrabin @lzap @shlomizadok ) to re-migrate our development environments, but it prevents from getting this kind of errors years after in production upgrades. So I am merging this as is.

_ALL_: Please migrate back and forth this migration or alter the two columns to `BIGINT` to prevent from future errors and confusion. Thanks.

lzap:
Just for the record the `cpu_count` was a joke, `</joke>` fun-tag did not render properly... ;-)

lzap:
Merged as 067a64a please don't forget to update your dev dbs. Thanks for the quick fix guys!

PR #117, opened 2015-01-12
Fixes #8911 - introduced discovery_attribute accessor

lzap:
to prevent nil errors in case discovery_attribute_set failed to create.

Also making sure that import facts API action is enrolled in two atomic
transactions so we either get:
- host with discovery_attribute_set properly set
- the above + auto provisioning

Also fixed template name typo.


app/models/host/discovered.rb
```diff
@@ -168,4 +168,11 @@ def compute_resource
     false
   end
 
+  # this helper method is used to retrieve individual attributes rather than using the relation
+  # directly (which is not safe when it's not set)
+  def discovery_attribute(name, default = 0)
+    return default if discovery_attribute_set.nil?
```
ohadlevy: why is this in the model? I think this should be a ui helper ?

app/views/dashboard/_discovery_widget_host_list.html.erb
```diff
@@ -12,11 +12,11 @@
       <tr>
         <td><%= link_to h(host.name), discovered_host_path(host) %></td>
         <td><%= host.ip %></td>
-        <td><%= turn_zero_to_not_available(host.discovery_attribute_set.cpu_count) %></td>
-        <td><%= number_to_human_size(turn_zero_to_not_available(host.discovery_attribute_set.memory * 1024 * 1024)) %></td>
-        <td><%= turn_zero_to_not_available(host.discovery_attribute_set.disk_count) %></td>
-        <td><%= number_to_human_size(host.discovery_attribute_set.disks_size) %></td>
+        <td><%= turn_zero_to_not_available(host.discovery_attribute(:cpu_count)) %></td>
```
ohadlevy: If you want to display a default value, you can do it in a ui helper - for example

``` erb
<td><%= discovery_attribute(host, :memory) %></td>
```

and then in an helper

``` ruby
def discovery_attribute host, attribute, default_value = _('N/A')
  host.try(:attribute) || default_value
end
```

ohadlevy: a working gist on my system ->https://gist.github.com/anonymous/2a3f1158688d731aedac

ohadlevy: just realize this is also relevant for the API, are we missing tests?

orrabin: currently discovery only has api tests that check if the response is redirect/success and not tests on which attributes are received in the response

app/helpers/discovered_hosts_helper.rb
```diff
@@ -40,4 +40,8 @@ def turn_zero_to_not_available(value)
     value == 0 ? 'N/A' : value
   end
 
+  def discovery_attribute host, attr, default_value = _('N/A')
```
shlomizadok: parentheses for method arguments (https://github.com/styleguide/ruby)

lzap:
Ok I rebased @ohadlevy patch and added one test for V2 API. But I don't understand what you mean "this is also relevant for the API".

app/helpers/discovered_hosts_helper.rb
lzap: Fixed, thanks.

lzap:
I don't get what can be wrong with our API RABL:

```
attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :cpu, :memory, :disk_count, :disks_size
```

We just return nil in some cases. That's expected?

lzap:
@orrabin explained to me.

Isn't this little bit overengineered? Why don't we simply drop the three delegates and create three proper read accessors which returns a default value when the relation does not exist?

app/views/dashboard/_discovery_widget.html.erb
```diff
@@ -4,7 +4,7 @@
 <% if hosts.empty? %>
    <p class="ca"><%= _("No discovered hosts available") %></p>
 <% else %>
-    <%= render :partial => 'discoverey_widget_host_list', :locals => { :hosts => hosts } %>
+    <%= render :partial => 'discovery_widget_host_list', :locals => { :hosts => hosts } %>
```
ohadlevy, line 7: please double check the widget size, in my local testing it was bigger then its allowed size.

ohadlevy:
@lzap which part is overengineered?

app/views/dashboard/_discovery_widget.html.erb
lzap, line 7: Fixed.

lzap:
Pushed the allowed size fix, forget about overengineering. It was a bad idea to drop the delegates. I am merging this since this is not my code actually :-)

lzap:
Pushed as
- d4096fa Fixes #8911 - introduced discovery_attribute accessor <Lukas Zapletal>
- 7c1e55b Refs #8911 - fixed spelling of a template name <Lukas Zapletal>
- 7566d90 Refs #8911 - added SQL transaction for fact

PR #118, opened 2015-01-13
DO NOT MERGE! Fixes #8872 - Add review form for auto provision all

orrabin:


app/assets/stylesheets/provision.scss
```diff
@@ -0,0 +1,3 @@
+.row-button {
+  margin-bottom:15px;
```
orrabin, line 2: I know it's a little overkill to add a stylesheet just for this but we might have more to add in auto_provision_all form and I could change this to `application.scss` in case there might be changes in other areas of the plugin.
I don't mind removing this if that's preferred.

app/views/discovered_hosts/auto_provision_all.html.erb
```diff
@@ -0,0 +1,29 @@
+<% title(_("Discovered hosts to provision")) %>
+<% stylesheet("provision") %>
+
+<%= form_tag submit_auto_provision_all_discovered_hosts_path, :method => 'POST' do -%>
```
ohadlevy: afaik POST is the default.

app/views/discovered_hosts/auto_provision_all.html.erb
```diff
@@ -0,0 +1,29 @@
...
+
+    <div class="btn-toolbar pull-right row-button">
+      <%= link_to _("Cancel"), send("#{controller_name}_path"), :class => "btn btn-default" %>
+      <%= submit_tag _('Start'), :class => "btn btn-success" %>
+    </div>
+
+    <table class="table table-bordered table-striped table-condensed">
+      <tr>
```
ohadlevy: missing thead / tbody

app/views/discovered_hosts/auto_provision_all.html.erb
```diff
@@ -0,0 +1,29 @@
+<% title(_("Discovered hosts to provision")) %>
+<% stylesheet("provision") %>
+
+<%= form_tag submit_auto_provision_all_discovered_hosts_path, :method => 'POST' do -%>
+
+    <div class="btn-toolbar pull-right row-button">
+      <%= link_to _("Cancel"), send("#{controller_name}_path"), :class => "btn btn-default" %>
```
ohadlevy, line 7: can probably avoid the send here, and be explicit on the controller name?

app/assets/stylesheets/provision.scss
lzap, line 2: I think this is fine, but I'd go for `auto_provision.scss` or similar to the ERB which uses it.

Also you are missing end of the line (EOF).

app/controllers/discovered_hosts_controller.rb
```diff
@@ -173,9 +177,9 @@ def auto_provision_all
         end
       end
```
lzap, line 178: Here is one drawback of this approach. We are no longer able to find rules and provision in one transaction. But since we want to be interactive, we have to do this...

app/views/discovered_hosts/auto_provision_all.html.erb
```diff
@@ -0,0 +1,33 @@
+<% title(_("Discovered hosts to provision")) %>
+<% stylesheet("provision") %>
+
+<%= form_tag submit_auto_provision_all_discovered_hosts_path do -%>
+
```
lzap, line 5: Extra empty line?

app/helpers/discovered_hosts_helper.rb
```diff
@@ -40,4 +40,9 @@ def turn_zero_to_not_available(value)
     value == 0 ? 'N/A' : value
   end
 
+  def get_rule_attributes rule, attr, default_value = nil
+    return rule.last.try(attr) if rule.last # rule.last is a DiscoveryRule or false
+    default_value
+  end
```
lzap, line 46: Can you explain this a little bit?

PR #119, opened 2015-01-14
Fixes #8937 - Auto provision all hidden when there are no hosts

orrabin:


app/views/discovered_hosts/index.html.erb
```diff
@@ -1,3 +1,3 @@
 <% title_actions multiple_discovered_hosts_actions_select -%>
-<% title_actions display_link_if_authorized(_("Auto Provision All"), hash_for_auto_provision_all_discovered_hosts_path, :method => :post) %>
+<% title_actions display_link_if_authorized(_("Auto Provision All"), hash_for_auto_provision_all_discovered_hosts_path, {:method => :post, :disabled => @hosts.empty?}) %>
```
ohadlevy, line 2: looks good, I would consider adding another check in the controller (e.g. when you click on auto provision) that there are no 0 hosts. (I assume the same is relevant for the API once it exists)

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -145,13 +152,14 @@ def auto_provision_all
               logger.warn "No rule found for host %s" % discovered_host.name
             end
           end
+
           if result
             process_success _("Discovered hosts are provisioning now")
```
orrabin, line 157: @lzap when testing the changes I noticed that the api returns an empty hash when it is successful and not this message.

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -145,13 +152,14 @@ def auto_provision_all
...
           else
             render_error :custom_error,
               :status => :unprocessable_entity,
               :locals => {
-                :message => _("Errors during auto provisioning: %s") % overall_errors
+                :message => error_message % overall_errors
```
lzap, line 162: Hmm can we really do: `"No discovered hosts to provision" % overall_errors` ?

At least this does not look sane.

orrabin, line 162: it shows up fine but I can change it to make it look better

shlomizadok:
tested and lgtm. One little addition I might think of is to also hide the multi-action dropdown if there are no hosts. (seems to be a cookie issue?)

shlomizadok:
@orrabin explained me it seems to be a cookie issue. The multi-action disappears when there's no cookie. Anyhow, it looks good to merge.

lzap:
The `overall_errors` does not seem to cause any troubles to Ruby runtime. Merged as eb448f4.

PR #120, opened 2015-01-14
Fixes #8901 - Make new apipie method optional

GregSutcliffe:
This is pretty ugly, but it allows us to retain the previous behaviour from before https://github.com/theforeman/foreman_discovery/commit/f1f29ca8f9c4aa92a90a6d70d197e024502b21b5#diff-9c896d7540d8746964738086b33782b7L112 was added (which breaks 1.7 support, and I'd like to keep 1.7 for discovery 2.0). Suggestions on cleaner handling are welcome.

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -36,7 +36,7 @@ module ForemanDiscovery
 
     initializer 'foreman_discovery.register_plugin', :after=> :finisher_hook do |app|
       Foreman::Plugin.register :foreman_discovery do
-        requires_foreman '> 1.7'
+        requires_foreman '>= 1.6'
 
         # discovered hosts permissions
         security_block :discovery do
@@ -107,7 +107,10 @@ module ForemanDiscovery
         allowed_template_helpers :rand, :facts_hash
 
         # apipie API documentation
-        apipie_documented_controllers ["#{ForemanDiscovery::Engine.root}/app/controllers/api/v2/*.rb"]
+        # Only available in 1.8, otherwise it has to be in the initializer below
+        if (SETTINGS[:version].to_s.include?('develop') or Gem::Version.new(SETTINGS[:version]) >= Gem::Version.new('1.8'))
+          apipie_documented_controllers ["#{ForemanDiscovery::Engine.root}/app/controllers/api/v2/*.rb"]
+        end
       end
     end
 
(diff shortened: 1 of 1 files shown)


lib/foreman_discovery/engine.rb
```diff
@@ -117,8 +120,16 @@ class Engine < ::Rails::Engine
...
-      if Apipie.configuration.respond_to?(:checksum_path)
-        Apipie.configuration.checksum_path += ['/discovered_hosts/']
+      # Also need to handle the reverse of the 1.8 method above
+      unless Gem::Version.new(SETTINGS[:version]) >= Gem::Version.new('1.8')
+        if Apipie.configuration.api_controllers_matcher.is_a?(Array) && Apipie.configuration.respond_to?(:checksum_path)
+          Apipie.configuration.api_controllers_matcher << "#{ForemanDiscovery::Engine.root}/app/controllers/api/v2/*.rb"
+          Apipie.configuration.checksum_path += ['/discovered_hosts/']
+        end
+      else
+        if Apipie.configuration.respond_to?(:checksum_path)
```
domcleal: Can stay outside the version conditional?

shlomizadok:
@GregSutcliffe - tested with foreman-1.7-stable. Works well.

lib/foreman_discovery/engine.rb
lzap: Whatever, please update so I can test this and give it go. Thanks.

lzap:
Ha, does not work with develop branch :-(

```
/home/lzap/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/rubygems/version.rb:191:in `initialize': Malformed version number string 1.8.0-develop (ArgumentError)
        from /home/lzap/work/foreman_discovery/lib/foreman_discovery/engine.rb:111:in `new'
        from /home/lzap/work/foreman_discovery/lib/foreman_discovery/engine.rb:111:in `block (2 levels) in <class:Engine>'
        from /home/lzap/work/foreman/app/services/foreman/plugin.rb:64:in `instance_eval'
        from /home/lzap/work/foreman/app/services/foreman/plugin.rb:64:in `register'
```

lib/foreman_discovery/engine.rb
GregSutcliffe: Thanks @domcleal, good catch.

GregSutcliffe:
@lzap ouch, sorry. Updated to work with both 1.7 and nightly. Still ugly though :)

lib/foreman_discovery/engine.rb
```diff
@@ -107,7 +107,10 @@ class Engine < ::Rails::Engine
         allowed_template_helpers :rand, :facts_hash
 
         # apipie API documentation
-        apipie_documented_controllers ["#{ForemanDiscovery::Engine.root}/app/controllers/api/v2/*.rb"]
+        # Only available in 1.8, otherwise it has to be in the initializer below
+        if (SETTINGS[:version].to_s.include?('develop') or Gem::Version.new(SETTINGS[:version]) >= Gem::Version.new('1.8'))
```
domcleal, line 111: usually we use `Gem::Version.new(SETTINGS[:version].chomp('-develop'))`

(although ideally we never do version comparisons!)

GregSutcliffe, line 111: Agreed its ugly. @lzap is it better to revert the original change to the apipie usage until after we're done with 1.7?

lzap, line 111: I am fine with this. Filed issue ticket for 2.1 to remove this workaround: http://projects.theforeman.org/issues/9017

Going to test this. Thanks.

lzap:
Works both for 1.7 and develop. Merged as 451c486 thanks.

PR #121, opened 2015-01-19
Fixes #9023 - discovery_fact_column global setting added

lzap:
Looks like this:

Set to blank to remove the column (default).

Set to any fact name to show it e.g. `bios_vendor`.

When fact is not present for particular host, it's blank.

### DIFF
--- a/app/views/discovered_hosts/_discovered_host.html.erb
+++ b/app/views/discovered_hosts/_discovered_host.html.erb
@@ -6,4 +6,7 @@
 <% unless defined? hide_disk %>
 <td><%= discovery_attribute(host, :disk_count) %></td>
 <td><%= number_to_human_size(discovery_attribute(host, :disks_size, 0)) %></td>
+<% if Setting['discovery_fact_column'].present? %>
+  <td><%= host.facts_hash[Setting['discovery_fact_column']] || 'N/A' %></td>
+<% end %>
 <% end %>
--- a/app/views/discovered_hosts/_discovered_hosts_list.html.erb
+++ b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
@@ -12,6 +12,9 @@
     <th><%= sort :memory, :as => _('Memory') %></th>
     <th><%= sort :disk_count, :as => _('Disk count') %></th>
     <th><%= sort :disks_size, :as => _('Disks size') %></th>
+    <% if Setting['discovery_fact_column'].present? %>
+      <th><%= Setting['discovery_fact_column'].capitalize %></th>
+    <% end %>
     <% if SETTINGS[:locations_enabled] -%>
       <th><%= sort :location, :as => _('Location') %></th>
     <% end -%>
--- a/test/functional/discovered_hosts_controller_test.rb
+++ b/test/functional/discovered_hosts_controller_test.rb
@@ -2,7 +2,7 @@ require 'test_plugin_helper'
 
 class DiscoveredHostsControllerTest < ActionController::TestCase
   setup :initialize_host
-  
+
   setup do
     @request.env['HTTP_REFERER'] = '/discovery_rules'
     @facts = {
@@ -14,6 +14,18 @@ class DiscoveredHostsControllerTest < ActionController::TestCase
 
   def test_index
     get :index, {}, set_session_user
+    assert_response :success
+  end
+
+  def test_index_with_custom_column
+    FactoryGirl.create(:setting,
+                       :name => 'discovery_fact_column',
+                       :value => "bios_vendor",
+                       :category => 'Setting::Discovered')
+    facts = @facts.merge({"bios_vendor" => "QEMU"})
+    Host::Discovered.import_host_and_facts(facts).first
+    get :index, {}, set_session_user
+    assert_select "td", /QEMU/
     assert_response :success
   end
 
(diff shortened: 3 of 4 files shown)


app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -10,6 +10,9 @@
     <th class="hidden-tablet hidden-xs"><%= sort :memory, :as => _('Memory') %></th>
     <th class="hidden-tablet hidden-xs"><%= sort :disk_count, :as => _('Disk count') %></th>
     <th class="hidden-tablet hidden-xs"><%= sort :disks_size, :as => _('Disks size') %></th>
+    <% if !Setting['discovery_fact_column'].nil? && Setting['discovery_fact_column'].size > 0 %>
+      <th class="hidden-tablet hidden-xs">Fact <%= Setting['discovery_fact_column'] %></th>
```
orrabin: why add `Fact` to the column name?

app/views/discovered_hosts/_discovered_host.html.erb
```diff
@@ -6,4 +6,7 @@
 <% unless defined? hide_disk %>
 <td class="hidden-tablet hidden-xs"><%= discovery_attribute(host, :disk_count) %></td>
 <td class="hidden-tablet hidden-xs"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0)) %></td>
+<% if !Setting['discovery_fact_column'].nil? && Setting['discovery_fact_column'].size > 0 %>
```
orrabin: maybe you can access the settings only once? like:

``` ruby
<% if  Setting['discovery_fact_column'].try(:size).to_i > 0 %>
```

app/views/discovered_hosts/_discovered_hosts_list.html.erb
lzap: You mean to capitalize just the fact name? Ok.

app/views/discovered_hosts/_discovered_host.html.erb
lzap: But `try` won't work for `nil` and also it does not compare to numbers. Isn't there a nicer way anyway?

lzap: Present! Changing.

lzap:
Rebased!

app/views/discovered_hosts/_discovered_host.html.erb
```diff
@@ -1,9 +1,12 @@
 <td class="hidden-tablet hidden-xs display-two-pane"><%= link_to h(host.name), discovered_host_path(host) %></td>
-<td><%= model_name host %></td>
-<td><%= host.ip %></td>
-<td><%= discovery_attribute(host, :cpu_count) %></td>
-<td><%= number_to_human_size(discovery_attribute(host, :memory, 0) * 1024 * 1024) %></td>
+<td class="hidden-tablet hidden-xs"><%= model_name host %></td>
```
orrabin: The `hidden` classes don't disappear with two-pane so they either need to be on the headers too or remove them completely.
I prefer to remove them because it makes the two-pane look better but if you want to keep them for resize this needs to be re-added to the headers.

lzap: Yeah this should be `|| 'N/A'`

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -12,6 +12,9 @@
     <th><%= sort :memory, :as => _('Memory') %></th>
     <th><%= sort :disk_count, :as => _('Disk count') %></th>
     <th><%= sort :disks_size, :as => _('Disks size') %></th>
+    <% if Setting['discovery_fact_column'].present? %>
+      <th class="hidden-tablet hidden-xs"><%= Setting['discovery_fact_column'].capitalize %></th>
```
orrabin: missed one hidden

orrabin:
tested this with a fact that exists for all my hosts, some of my hosts, none of my hosts, an empty fact and no fact in the settings and it works great :)

should this be added to the api too?

orrabin:
@lzap as we discussed I'll merge after tests pass and we can decide about the api later, thanks for creating an issue to remind us: http://projects.theforeman.org/issues/9032

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -2,13 +2,14 @@
...
   setup :initialize_host
-  
+
   setup do
     @request.env['HTTP_REFERER'] = '/discovery_rules'
     @facts = {
       "ipaddress" => "192.168.100.42",
       "macaddress" => "AA:BB:CC:DD:EE:FF",
       "discovery_bootif" => "AA:BB:CC:DD:EE:FF",
+      "bios_vendor" => "QEMU",
```
orrabin: nitpick: maybe use `facts = @facts.merge({"bios_vendor" => "QEMU"})` in the test instead on adding it here? because we don't want future tests to keep adding facts that only they use

orrabin:
merged as 8c8956e67881acdfdc8c79d477fbb5c5d19b5453 thanks @lzap!

PR #126, opened 2015-01-20
Refs #8831 - Reverts two-pane forms

lzap:
We are pulling two-pane patch off the 2.0 for now. It's too complicated for 
now. Let's do proper implementation for Foreman 1.8 later on.

This reverts commit 7a6454704cf87f712d9f32fae584245c41f24fa1.

Conflicts:
app/views/discovered_hosts/_discovered_host.html.erb
app/views/discovered_hosts/_discovered_hosts_list.html.erb

### DIFF
--- a/app/assets/javascripts/discovery-two-pane.js
+++ b/app/assets/javascripts/discovery-two-pane.js
@@ -1,6 +1,0 @@
-$(document).on('click', ".table-two-pane .display-two-pane a", function (e) {
-  if ($('.table-two-pane').length) {
-    e.preventDefault();
-    two_pane_open(this);
-  }
-});
--- a/app/assets/stylesheets/discovery-two-pane.scss
+++ b/app/assets/stylesheets/discovery-two-pane.scss
@@ -1,3 +1,0 @@
-.btn-group {
-  margin-right: 10px;
-}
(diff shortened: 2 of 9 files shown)


app/views/dashboard/_discovery_widget_host_list.html.erb
```diff
@@ -1,10 +1,10 @@
 <table class="table table-striped ellipsis">
   <thead>
-  <th><%= _('Host') %></th>
-  <th><%= _('Model') %></th>
-  <th><%= _('IP Address') %></th>
-  <th><%= _('CPU') %></th>
-  <th><%= _('Memory') %></th>
```
orrabin, line 7: these have nothing to do with two-pane, keep them in

GregSutcliffe:
Merged, thanks @lzap

ohadlevy:
Why?

GregSutcliffe:
1.7-stable has a display bug which causes issues with two-pane and discovery. Since we need Discovery 2.0 for the next Sat release, we had the choice of some very ugly (and long-lived) JS hacks, or delaying two-pane to Discovery 2.1

lzap:
We basically had three options:
- ugly fix to get it working
- disable two-pane for discovery hosts list page (until it's fixed)
- revert and re-think everything

Since the two-pane already introduced hacks (JS/CSS) into the discovery we voted that the cleanest way is to just dropp the patch for 2.0. We can easily add it back after the release - in even better form. Once required patches are in the 1.8 release, it will be cleaner.

PR #129, opened 2015-01-22
Fixes #9027 - adding sorting on discovery rule and enabled

orrabin:


lzap:
Works as expected, the test failure is not releated. Merged as aceafa5 thank you.

PR #131, opened 2015-01-22
Fixes #9077 - added tests for safemode

lzap:
This patch basically adds tests for safemode (and also both non-safemode tests)
and it slightly improve hostname rendering (when it renders to empty string it
falls back to the original name `macMACADDRESS`.

Also I removed the `facts_hash` from allowed helpers because this is not a
helper but method. It will be allowed (together with alias `facts`) in a
different patch in core: https://github.com/theforeman/foreman/pull/2102

One test is set to be skipped - it won't be green until the above patch is
applied.

### DIFF
--- a/test/unit/discovered_extensions_test.rb
+++ b/test/unit/discovered_extensions_test.rb
@@ -95,12 +95,81 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     assert_equal host.name, "macaabbccddeeff"
   end
 
-  test "hostname is rendered after auto provisioning" do
-    facts = @facts.merge({"somefact" => "abc"})
-    host = Host::Discovered.import_host_and_facts(facts).first
-    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc", :hostname => 'x<%= 1+1 %>')
-    perform_auto_provision host, r1
-    assert_equal host.name, "x2"
+  def setup_normal_renderer
+    Setting[:safemode_render] = false
+    @facts.merge!({"somefact" => "abc"})
+  end
+
+  def setup_safemode_renderer
+    Setting[:safemode_render] = true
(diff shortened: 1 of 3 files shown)


orrabin:
[test]

lzap:

lzap:
Rebased, keeping the two tests skipped for 2.0 stable (let's remove this later).

test/unit/discovered_extensions_test.rb
```diff
@@ -95,12 +95,91 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
...
+                              :search => "facts.somefact = abc",
+                              :hostname => '<%= "" %>')
+      perform_auto_provision host, r1
+      assert_equal "macaabbccddeeff", host.name
+    end
+
+    test "hostname is rendered after auto provisioning using #{renderer_name}" do
+      facts = @facts.merge({"somefact" => "abc"})
+      host = Host::Discovered.import_host_and_facts(facts).first
+      r1 = FactoryGirl.create(:discovery_rule,
```
orrabin, line 129: @lzap why are you adding the priority and search if they already exist in the factory?
maybe change the fact in the factory or the fact here and then you can just ad `hostname`?
you could also move the `@facts.merge` to the setup

orrabin:
@lzap I tested and this looks good, you can DRY tests a bit but other then that :+1: to merge

test/unit/discovered_extensions_test.rb
lzap, line 129: Fixed, priority is not necessary as there is only one rule, yeah. Keeping search tho, I want to search for `abc` and not `fact1`. This is more obvious to newcomers.

lzap:
Fixed, thanks.

orrabin:
merged as 90bd37b86f3e26273051de28beb0f412998b347d, thanks @lzap

PR #132, opened 2015-01-22
Fixes #9029 - explanation popover for hostname in discovery_rules

orrabin:


orrabin:
the popover is pretty narrow
I'm not sure if giving it a fixed width is a good idea, any other ideas are welcome

app/views/discovery_rules/_form.html.erb
```diff
@@ -6,7 +6,18 @@
                      :control_group_id => 'search_group', :size => 'col-md-9' %>
   <%= select_f f, :hostgroup_id, accessible_hostgroups, :id, :to_label, { :include_blank => false },
     { :help_inline => _('Target host group for this rule with all properties set')} %>
-  <%= text_f f, :hostname, :help_inline => _('Target hostname template pattern (empty for MAC)')  %>
+  <%= text_f f, :hostname, :help_inline => popover(_("Target hostname template pattern"),_("<dl>" +
+                                                                  "<dd>Domain will be appended automatically.</dd>" +
```
lzap, line 10: We can't translate like this, otherwise translators will mess this up. http://projects.theforeman.org/projects/foreman/wiki/Translating

The best approach here is something like:

```
"<dl>" +
"<dd>" + _("Domain will be appended automatically.") + "</dd>" +
...
```

app/views/discovery_rules/_form.html.erb
```diff
@@ -6,7 +6,18 @@
...
+                                                                  "<dd>A hostname based on MAC address will be used when left blank.</dd>" +
+                                                                  "<dd>Use ERB templating to access host attributes and facts. Examples:</dd>" +
+                                                                  "<li>myhost-\<%= rand(99999) %%\></li>" +
```
lzap, line 13: Double per cent signs: `%%>`

lzap:
Ok I agree its too narrow, but I think we can make it. Let's just drop these `dd` and `li` tags and do this:

``` html
<div>First paragrah - three sentences.</div>
<small>example1</small><br/>
<small>exampleN</small><br/>
<div>Last paragrah - the remaining text</div>
```

Maybe try variation on `<small>` and `<code>`.

lzap:
Let me do this.

orrabin:
[test]

app/views/discovery_rules/_form.html.erb
orrabin, line 13: this is to escape `%`

PR #133, opened 2015-01-23
Fixes #9029 - popup for discovery rule hostname

lzap:
Replaces https://github.com/theforeman/foreman_discovery/pull/132

### DIFF
--- a/app/views/discovery_rules/_form.html.erb
+++ b/app/views/discovery_rules/_form.html.erb
@@ -6,7 +6,27 @@
                      :control_group_id => 'search_group', :size => 'col-md-9' %>
   <%= select_f f, :hostgroup_id, accessible_hostgroups, :id, :to_label, { :include_blank => false },
     { :help_inline => _('Target host group for this rule with all properties set')} %>
-  <%= text_f f, :hostname, :help_inline => _('Target hostname template pattern (empty for MAC)')  %>
+  <%= text_f f, :hostname, :help_inline => popover(
+    _("Template"),
+    "<div>" +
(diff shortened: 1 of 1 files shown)


app/views/discovery_rules/_form.html.erb
```diff
@@ -6,7 +6,27 @@
                      :control_group_id => 'search_group', :size => 'col-md-9' %>
   <%= select_f f, :hostgroup_id, accessible_hostgroups, :id, :to_label, { :include_blank => false },
     { :help_inline => _('Target host group for this rule with all properties set')} %>
-  <%= text_f f, :hostname, :help_inline => _('Target hostname template pattern (empty for MAC)')  %>
+  <%= text_f f, :hostname, :help_inline => popover(
+    _("Template"),
```
orrabin, line 10: nitpick: maybe change to `Hostname template` or `Hostname pattern`?

orrabin:
@lzap tested and this works fine

app/views/discovery_rules/_form.html.erb
lzap, line 10: Well actually if I do that then the booble will be narrow because the more text I add to the link the less space remains. If you have a solution how to change the arrow direction I can do that.

orrabin, line 10: I can't think of anything better other then changing this from inline help to just a popover under the hostname which can give it a little more room.
I'll merge as is.

orrabin:
merged as 6c8df37ac215cbd97a3fe79cab2de694b3810710, thanks @lzap!

PR #134, opened 2015-01-25
Fixes #9103 - show cpu_count in api

orrabin:


app/views/dashboard/_discovery_widget_host_list.html.erb
```diff
@@ -3,7 +3,7 @@
   <th><%= _('Host') %></th>
   <th><%= _('Model') %></th>
   <th><%= _('IP Address') %></th>
-  <th><%= _('CPU') %></th>
+  <th><%= _('CPU count') %></th>
```
lzap: This will occupy more space and there is not much space there. Mind changing this to `CPUs` both here and in the main discovery list (we have `CPU` there as well).

lzap: Or maybe CPU(s) is even better.

domcleal: `CPUs` I'd say.

app/views/api/v2/discovered_hosts/main.json.rabl
```diff
@@ -2,7 +2,8 @@ object @discovered_host
 
 extends "api/v2/discovered_hosts/base"
 
-attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :cpu, :memory, :disk_count, :disks_size
+attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :memory, :disk_count, :disks_size
+attribute :cpu_count => :CPUs
```
tstrachota: Do we use capital attributes anywhere else? It feels a bit irregular to me.

orrabin: we show it with capital letters in the UI and I couldn't think of a better way to differ the 's' but I can change it

domcleal: I don't know if this one API attribute should be changed if you also have disk_count.  Isn't cpu_count OK here?

app/views/api/v2/discovered_hosts/main.json.rabl
```diff
@@ -2,7 +2,8 @@ object @discovered_host
 
 extends "api/v2/discovered_hosts/base"
 
-attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :cpu, :memory, :disk_count, :disks_size
+attributes :ip, :mac, :last_report, :subnet_id, :subnet_name, :memory, :disk_count, :disks_size
+attribute :cpu_count => :cpus
```
orrabin, line 6: @domcleal  we have `Disk count` and `CPUs` in the UI  so I wanted to make the api look close

lzap, line 6: Although I think this was not necessary, I am fine with this.

lzap:

lzap:
It works just fine. Since the API is fresh, I am not against the change there.

Merged as 5cb2142, thank you.

PR #135, opened 2015-01-25
refs #9044 - Fixes broken mac tests

orrabin:
Tests for discovery were broken: http://ci.theforeman.org/job/test_plugin/4906/#showFailuresLink
since https://github.com/theforeman/foreman/commit/ee1f56de6b2bdca4dfd104b99ec25340bfb51aaa was committed.

### DIFF
--- a/test/unit/facts.json
+++ b/test/unit/facts.json
@@ -88,7 +88,7 @@
     "lsbmajdistrelease": "6",
     "swapfree": "16.72 GB",
     "netmask": "255.255.255.192",
-    "discovery_bootif": "AA:BB:CC:DD:EE:FF",
+    "discovery_bootif": "AA:BB:CC:DD:EE:F5",
     "memorysize_mb": "742.13",
     "blockdevice_sda_size": "214749184"
   }
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -54,7 +54,16 @@ class Host::Discovered < ::Host::Base
     h = ::Host::Discovered.find_by_name hostname
     h ||= Host.new :name => hostname, :type => "Host::Discovered"
     h.type = "Host::Discovered"
-    h.mac = facts[fact_name].try(:downcase)
+
+    macfact = facts[fact_name].try(:downcase)
+    begin
+      macfact = Net::Validations.normalize_mac(macfact)
+    rescue ArgumentError => e
+      macfact = facts['discovery_bootif'].try(:downcase)
+      h.name = facts['discovery_bootif'].try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
+    end
+    h.mac = macfact
+
 
     if SETTINGS[:locations_enabled]
       begin
--- a/test/unit/host_discovered_test.rb
+++ b/test/unit/host_discovered_test.rb
@@ -21,7 +21,7 @@ class HostDiscoveredTest < ActiveSupport::TestCase
   test "should import facts from yaml as Host::Discovered" do
     raw = parse_json_fixture('/facts.json')
     assert Host::Discovered.import_host_and_facts(raw['facts'])
-    assert Host::Discovered.find_by_name('macaabbccddeeff')
+    assert Host::Discovered.find_by_name('macaabbccddeef5')
   end
 
   test "should raise when fact_name setting isn't present" do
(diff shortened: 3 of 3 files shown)


test/unit/facts.json
```diff
@@ -88,7 +88,7 @@
     "lsbmajdistrelease": "6",
     "swapfree": "16.72 GB",
     "netmask": "255.255.255.192",
-    "discovery_bootif": "AA:BB:CC:DD:EE:FF",
+    "discovery_bootif": "AA:BB:CC:DD:EE:F5",
```
lzap, line 91: Is this change really required?

lzap:
For the record the error is:

```
HostDiscoveredTest#test_0005_should create discovered host with fact_name as a name:
ArgumentError: 'santiago' is not a valid MAC address
    /home/lzap/work/foreman/lib/net/validations.rb:83:in `normalize_mac'
    /home/lzap/work/foreman/app/models/host/base.rb:111:in `block in set_interfaces'
    /home/lzap/work/foreman/app/models/host/base.rb:96:in `each'
    /home/lzap/work/foreman/app/models/host/base.rb:96:in `set_interfaces'
    /home/lzap/work/foreman/app/models/host/base.rb:83:in `populate_fields_from_facts'
    /home/lzap/work/foreman_discovery/app/models/host/discovered.rb:93:in `populate_fields_from_facts'
    /home/lzap/work/foreman/app/models/host/base.rb:61:in `import_facts'
    /home/lzap/work/foreman_discovery/app/models/host/discovered.rb:83:in `import_facts'
    /home/lzap/work/foreman_discovery/app/models/host/discovered.rb:75:in `import_host_and_facts'
    /home/lzap/work/foreman_discovery/test/unit/host_discovered_test.rb:45:in `block in <class:HostDiscoveredTest>'
```

Now, I would like to keep our code logic as is, but check it first like:

```
   mac = facts[fact_name]
   begin
     mac = normalize_mac(mac)
   rescue
     mac = facts['discovery_bootif']
   end
```

test/unit/facts.json
orrabin, line 91: yes, many tests where failing because of mac turning into nil because of https://github.com/theforeman/foreman/pull/1923/files#r22298588
the "next mac" was trying to be created and since this one ends with `ff` it would be an invalid mac

orrabin:
@lzap I'll change the behavior to allow custom facts that hold mac addresses and default to `discovery_bootif` only when the fact doesn't create a valid mac, thanks for the explanation.

lzap:
Great, would you mind adding additional test to cover the exception code path?

orrabin:
@lzap I don't understand, we had 2 tests failing with this exception

lzap:
Then the patch is not correct. This should fail:

```
  test "should create discovered host with fact_name as a name" do
    raw = parse_json_fixture('/facts.json')
    Setting[:discovery_fact] = 'lsbdistcodename'
    host = Host::Discovered.import_host_and_facts(raw['facts']).first
    assert_equal 'macsantiago', host.name
  end
```

But it does not. Expected contract now is:

```
  test "should create discovered host with fact_name as a name" do
    raw = parse_json_fixture('/facts.json')
    Setting[:discovery_fact] = 'lsbdistcodename'
    host = Host::Discovered.import_host_and_facts(raw['facts']).first
    assert_equal 'macaabbccddeef5', host.name
  end
```

orrabin:
@lzap, the name is ok, the problem was setting the mac:

``` ruby
h.mac = facts[fact_name].try(:downcase)
```

was trying to set the mac to be 'santiago'.
Now the test should result in a host with the name: `macsantiago` and the mac: `aa:bb:cc:dd:ee:f5` since the fix here will result in taking `discovery_bootif` for the mac after 

``` ruby
mac = facts[fact_name]
begin
      mac = normalize_mac(mac)
```

will fail

orrabin:
@lzap I can add to the test:

``` ruby
 assert_equal 'aa:bb:cc:dd:ee:f5', host.mac
```

or

``` ruby
 assert_equal 'aa:bb:cc:dd:ee:f5', raw['facts']['discovery_bootif'].try(:downcase)
```

after checking the name
will that be a good enough test to avoid this error?

lzap:
Uh, @orrabin that was a pseudo code, it does not work. The method is defined elsewhere. Try something like (still not finished code - does not work):

``` ruby
    macfact = facts[fact_name]
    begin
      macfact = ::Net::Validations.normalize_mac(macfact)
    rescue ArgumentError
      macfact = ::Net::Validations.normalize_mac(facts['discovery_bootif'])
    end
    h.mac = macfact.try(:downcase)
```

orrabin:
@lzap we could just try to set h.mac and check if h is valid, `Net::Validations.normalize_mac` will run on the mac before validations, do we need it here too?

lzap:
Up to you. That would work too yeah.

lzap:
I apologize but I've changed my mind :-D This can trigger false errors when something else is not valid but MAC is okay.

The `valid?` test is too broad. Let's do it via the direct validation. Also can you make sure that the hostname gets corrected when the validation error triggers? Update the two tests then as they will fail too.

orrabin:
@lzap why would the hostname change? it's just the mac that will be taken from a different fact

orrabin:
@lzap I'v changed back to the validation and added an assert mac for each test.
I still don't understand why the name should change, tests are passing locally.

lzap:
Hmmm fails for me locally.

```
# Running tests:

...................................................F..........

Finished tests in 10.897937s, 5.6892 tests/s, 8.6255 assertions/s.

  1) Failure:
HostDiscoveredTest#test_0006_should create discovered host with default name if fact_name isn't a valid mac [/home/lzap/work/foreman_discovery/test/unit/host_discovered_test.rb:55]:
<"macaabbccddeef5"> expected but was
<"macsantiago">.

62 tests, 94 assertions, 1 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /home/lzap/work/foreman/coverage. 10043 / 17620 LOC (57.0%) covered.
rake aborted!
Command failed with status (1): [ruby -I"lib:test:/home/lzap/work/foreman_discovery/lib/../test" -I"/home/lzap/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/rake-10.4.2/lib" "/home/lzap/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/rake-10.4.2/lib/rake/rake_test_loader.rb" "/home/lzap/work/foreman_discovery/lib/../test/**/*_test.rb" ]

Tasks: TOP => test:discovery
```

Let's [test] again?

lzap:
All is green now, thanks. Merged as 46b7148.

PR #136, opened 2015-01-25
refs #8963 - remove trunc from discovery_rules index

orrabin:
DO NOT MERGE before discovery works with foreman 1.8

### DIFF
--- a/app/views/discovery_rules/index.html.erb
+++ b/app/views/discovery_rules/index.html.erb
@@ -13,7 +13,7 @@
   </tr>
   <% for rule in @discovery_rules %>
     <tr>
-      <td class='col-md-3'><%= link_to_if_authorized(trunc(rule.name), hash_for_edit_discovery_rule_path(:id => rule).merge(:auth_object => rule, :authorizer => authorizer)) %></td>
+      <td class='col-md-3'><%= link_to_if_authorized(rule.name, hash_for_edit_discovery_rule_path(:id => rule).merge(:auth_object => rule, :authorizer => authorizer)) %></td>
       <td><%= rule.priority %></td>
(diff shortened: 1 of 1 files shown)


lzap:
We need this patch in 2.0 discovery implemented in a way that it will work both with 1.7 and 1.8 foreman. Plans for 2.0 is to support both. @ohadlevy @GregSutcliffe ?

GregSutcliffe:
Support for both would be ideal, but if that's not possible, I'd recommend releasing 2.0.0 for 1.7 (as thats the current Foreman release), using 2.0.x for the bugfix stream, and schedule a quick release next week of 2.1.0 for develop/1.8

orrabin:
@GregSutcliffe if you have an idea how to change this PR I think it's the only thing that breaks for now.
I'm not sure what else might break until 1.8 is released so we might still get to a situation where we need 2.1.

GregSutcliffe:
My only suggestion would be a helper (`discovery_trunc`?) that calls the correct version in turn. or add an alias for one of them to the other?

lzap:
Yeah something like that. Maybe `trunc_with_tooltip_compat` is better, but quite long name.

lzap:
This could be maybe in the core since the change will break all the plugins? Quite used function that was.

lzap:

orrabin:
@lzap we don't know if we'll need to truncate names of rules, they might all be very short.
I'm removing the trunc so this supports both versions of core and I opened: http://projects.theforeman.org/issues/9130 to remind us.

lzap:
This works just fine (we got rid of the trunc completely for 2.0). Great. Merged as 9d34f1b, thanks.

PR #137, opened 2015-01-26
Fixes #9090 - In discovery_rule host limit and priority can not be negative numbers

orrabin:


lzap:
Pushed as ebba0eb thank you.

PR #140, opened 2015-01-28
Fixes #9082 - adding auto_complete to discovery_rules routes

orrabin:


lzap:
Looks like it does not change anything:

```

Started GET "/discovered_hosts/auto_complete_search?search=cpu_count+%3D" for 127.0.0.1 at 2015-01-28 12:36:54 +0100
Processing by DiscoveredHostsController#auto_complete_search as JSON
  Parameters: {"search"=>"cpu_count ="}
  User Load (0.2ms)  SELECT "users".* FROM "users" WHERE "users"."id" = ? LIMIT 1  [["id", 1]]
Setting current user thread-local variable to admin
   (0.1ms)  SELECT COUNT(*) FROM "taxonomies" WHERE "taxonomies"."type" IN ('Organization')
Setting current organization thread-local variable to none
   (0.1ms)  SELECT COUNT(*) FROM "taxonomies" WHERE "taxonomies"."type" IN ('Location')
Setting current location thread-local variable to none
  AuthSource Load (0.2ms)  SELECT "auth_sources".* FROM "auth_sources" WHERE "auth_sources"."id" = 1 LIMIT 1
   (0.1ms)  SELECT id FROM "taxonomies" WHERE "taxonomies"."type" IN ('Location') LIMIT 1
   (0.1ms)  SELECT id FROM "taxonomies" WHERE "taxonomies"."type" IN ('Organization') LIMIT 1
Completed 500 Internal Server Error in 14.2ms

ActiveRecord::UnknownAttributeError (Host::Discovered(id: integer, name: string, last_compile: datetime, last_freshcheck: datetime, last_report: datetime, updated_at: datetime, source_file_id: integer, created_at: datetime, root_pass: string, serial: string, puppet_status: integer, architecture_id: integer, operatingsystem_id: integer, environment_id: integer, ptable_id: integer, medium_id: integer, build: boolean, comment: text, disk: text, installed_at: datetime, model_id: integer, hostgroup_id: integer, owner_id: integer, owner_type: string, enabled: boolean, puppet_ca_proxy_id: integer, managed: boolean, use_image: boolean, image_file: string, uuid: string, compute_resource_id: integer, puppet_proxy_id: integer, certname: string, image_id: integer, organization_id: integer, location_id: integer, type: string, otp: string, realm_id: integer, compute_profile_id: integer, provision_method: string, grub_pass: string, discovery_rule_id: integer) doesn't have column :ip.):
  app/controllers/concerns/application_shared.rb:13:in `set_timezone'
  app/models/concerns/foreman/thread_session.rb:33:in `clear_thread'
  lib/middleware/catch_json_parse_errors.rb:9:in `call'
```

orrabin:
@lzap this is not the same error as the bug states, can you please check if this is happening in develop too?

lzap:
For the record: I have some older discovered hosts that has both `ip` and `mac` set to nil.

I will look into this later today.

```
[5] pry(main)> ::Host::Discovered.all.collect(&:ip)
  Host::Discovered Load (0.5ms)  SELECT "hosts".* FROM "hosts" WHERE "hosts"."type" IN ('Host::Discovered')
  Nic::Base Load (0.3ms)  SELECT "nics".* FROM "nics" WHERE "nics"."host_id" = 15 ORDER BY identifier
  Nic::Base Load (0.2ms)  SELECT "nics".* FROM "nics" WHERE "nics"."host_id" = 21 ORDER BY identifier
=> [nil, nil]
```

lzap:
Unable to reproduce on my system... heh. But I see the different bug (above).

orrabin:
@lzap after this change or in develop too?

lzap:
With develop, when I create new rule and search for something it works:

```
Started GET "/discovered_hosts/auto_complete_search?search=+cpu_count+%3D++1" for 127.0.0.1 at 2015-01-28 13:44:11 +0100
```

The path is different tho.

lzap:
Ok the above bug was caused by NIC patch that was recently merged to core.

But I can't still reproduce with Foreman develop. Will test with 1.7.

lzap:
It does work fine with 1.7 as well on my system.

orrabin:
closing the PR and the issue as not reproducible

lzap:
It looks like we have fixed it already. Maybe it had something to do with floating/fixed panes.

orrabin:
@lzap that's why I closed the PR as resolved, this can be reopened if this is reproduced.

PR #141, opened 2015-01-28
Fixes #9145 - save disks_size in discovery_attribute_set as MB

orrabin:


lzap:
Good idea, mind adding a test case that covers this?

orrabin:
@lzap fixed the test and tested locally

lzap:
Works, tests are green. Merged as 5ce79c7, thank you.

PR #142, opened 2015-01-28
[Do Not Merge] WIP on multinic support

GregSutcliffe:
OK, here's my minimal changes to fix discovery for multinic support. Also needs https://github.com/theforeman/foreman/pull/2116. Seems to work for both manual and auto-provisioning on latest core/develop.

I'm not sure how to make this compatible with 1.7 - most could be wrapped in version tests or `.try()` calls I guess?

I'll check back tomorrow to see how the tests are going :)

I also think we'll need to revisit this once the new multinic support has settled down and discuss how we want to consume it in Discovery.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -76,10 +76,11 @@ module Api
       end
 
       def update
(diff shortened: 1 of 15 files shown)


GregSutcliffe:
Ugh, wrong target branch. Re-opening....

PR #147, opened 2015-01-29
Fixes #9161 - Discovering a host searches for discovery rules only when ...

orrabin:
...automatic auto_provision is enabled


lzap:
Looks correct, good catch. Re [test]

lzap:
Reworded commit message, please keep commit messages shorter than 50 characters. Thanks Ori! Merged as e97de42

PR #157, opened 2015-02-23
Fixes #9366 - host limit for discovery rule defaults to zero

orrabin:


lzap:
LGTM [test]

lzap:
Can we do the very same with priority? To be consistent.

lzap:
Thank you merged as 4c104cf.

PR #161, opened 2015-02-25
Fixes #9546 - added converter STI helper API

lzap:
This allows other plugin to convert to managed host correctly. We may need to
come with better name and richer API, this is WIP. Tests are green.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -76,19 +76,12 @@ module Api
       end
 
       def update
-        @host                           = @discovered_host.becomes(::Host::Managed)
-        @host.type                      = 'Host::Managed'
-        @host.managed                   = true
-        @host.primary_interface.managed = true
-        @host.build                     = true
-        forward_request_url
-        update_response = @host.update_attributes(params[:discovered_host])
         Host.transaction do
-          if update_response
-            delete_discovery_attribute_set(@host.id)
-          end
+          @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host)
+          forward_request_url
+          update_response = @host.update_attributes(params[:discovered_host])
+          process_response update_response
         end
-        process_response update_response
       end
 
       api :DELETE, "/discovered_hosts/:id/", N_("Delete a discovered host")
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -59,25 +59,18 @@ class DiscoveredHostsController < ::ApplicationController
   end
 
   def edit
-    unless @host.nil?
-      @host                           = @host.becomes(::Host::Managed)
-      @host.type                      = 'Host::Managed'
-      @host.managed                   = true
-      @host.primary_interface.managed = true
-      @host.build                     = true
+    Host.transaction do
+      @host = ::ForemanDiscovery::HostConverter.to_managed(@host) unless @host.nil?
+      render :template => 'hosts/edit'
     end
-
-    render :template => 'hosts/edit'
   end
 
   def update
-    @host         = @host.becomes(::Host::Managed)
-    @host.type    = 'Host::Managed'
-    forward_url_options
-    Taxonomy.no_taxonomy_scope do
-      Host.transaction do
+    Host.transaction do
+      @host = ::ForemanDiscovery::HostConverter.to_managed(@host, false, false)
+      forward_url_options
+      Taxonomy.no_taxonomy_scope do
         if @host.update_attributes(params[:host])
-          delete_discovery_attribute_set(@host.id)
           process_success :success_redirect => host_path(@host), :redirect_xhr => request.xhr?
         else
           taxonomy_scope
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -28,32 +28,21 @@ module Foreman::Controller::DiscoveredExtensions
   end
 
   # trigger the provisioning
-  def perform_auto_provision host, rule
-    host = host.becomes(::Host::Managed)
-    host.type = 'Host::Managed'
-    host.managed = true
-    host.build = true
-    host.hostgroup_id = rule.hostgroup_id
-    host.comment = "Auto-discovered and provisioned via rule '#{rule.name}'"
-    host.discovery_rule = rule
-    # render hostname only when all other fields are set
-    original_name = host.name
-    host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
-    # fallback to the original if template did not expand
-    host.name = original_name if host.name.empty?
+  def perform_auto_provision original_host, rule
     Host.transaction do
-      if host.save #save! does not work here
-        delete_discovery_attribute_set(host.id)
-      end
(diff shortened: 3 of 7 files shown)


lzap:
This solution relies of fact that the caller prepared SQL transaction in a way that all the work for the coversion is done there. In that case we can just delete the records directly and SQL commit will do all or nothing.

Since we use foreign keys, messing around with ActiveRecord callback would be tough. We'd need to remove the foreign key if we want to use `after_` callbacks anyway, so I prefer this solution which is more consistent. On the other hand, if the caller does not prepare transaction, then this code could end up with deleted attribute records in case conversion is canceled by user or error.

If you like this, I can add "downgrade" method which will re-create attribute records in the similar way.

lib/foreman_discovery/host_converter.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+  # discovery_attribute_set should be deleted only after a host has been successfully provisioned
+  # (after discovered host becomes managed host)
+  # after this happens the host is a manged host so non of the callbacks in discovered hosts are relevant anymore
+  # for this reason the orphaned discovery_attribute_set needs to be searched for and deleted
+  def self.delete_discovery_attribute_set(host_id)
+    DiscoveryAttributeSet.delete_all(:host_id => host_id)
```
ares: delete_all skips callbacks and potential associations, maybe it's safer to use destroy_all (slower but I don't think it matters too much in this case)

lzap:
I found a way to check for explicit transaction, so API users are warned before.

ares:
Much better, not tested though. Although the name `#update` is a bit confusing. How about `#to_managed`. Also similar approach should be taken for applying autodiscovery rule, but that's different issue :-)

lib/foreman_discovery/host_converter.rb
lzap: Yeah fixed.

lzap:
All fixed, renamed to `to_managed` that's a good name. Added two tests on deletion both variants (discovered and managed host records).

orrabin:
@lzap `refresh_facts` should recreate the attributes if needed so we don't have to have a "downgrade" method but maybe at some point add a `refresh_all` button in case all the attributes are somehow deleted (auto_provision_all failing for all hosts?)

lzap:
Yeah the `downgrade` was a bad idea. At first, nobody will likely downgrade to discovered host, the proper workflow is to delete the host and restart.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -59,24 +59,16 @@ def destroy
...
 
   def edit
-    unless @host.nil?
-      @host         = @host.becomes(::Host::Managed)
-      @host.type    = 'Host::Managed'
-      @host.managed = true
-      @host.build   = true
-    end
-
+    @host = ::ForemanDiscovery::HostConverter.to_managed(@host) unless @host.nil?
```
orrabin: this doesn't need to be in a transaction?

orrabin: this does need a transaction, provision from the UI isn't working because of this 
on the plus side your transaction warning is fine :)

orrabin:
@lzap tested against 1.7 for now - everything is working except provision through the UI and something that looks unrelated unrelated but I  haven't had a chance to look at it yet - auto_provision from a discovered_host view form (it does provision the host but there seems to be a routing issue after success)

lzap:
Fixed the missing transaction. The bug you hit looks like another redirect issue.

lib/foreman_discovery/host_converter.rb
```diff
@@ -0,0 +1,26 @@
+class ForemanDiscovery::HostConverter
```
GregSutcliffe: I thought we discouraged putting this kind of thing in `lib/` and instead prefer it in `app/services` ? Not too worried either way, but if that's the convention in core, then plugins should probably follow it too.

orrabin:
@lzap provision is working now

lib/foreman_discovery/host_converter.rb
lzap: > I thought we discouraged putting this kind of thing in `lib/` and instead prefer it in `app/services` ? Not too worried either way, but if that's the convention in core, then plugins should probably follow it too.

@mhulan is app services ok for staypuft?

## 

Later,
 Lukas #lzap Zapletal

ares: Frankly I don't see a reason why not putting it into app/models, but app/services works for me as well. Also I don't think staypuft will adopt this change anytime soon.

lzap:
Re [test]

lzap:
I will move from `lib/` to `apps/` ...

lzap:
Pushed the move to `apps/services`. Ready for testing.

orrabin:
I hope this is a random timeout and [test] again will work

orrabin:
@lzap is this for 2.1 or 3.0? 
#9546 is on 3.0 but it should also fix http://projects.theforeman.org/issues/9488 which is marked for 2.1

lzap:
Added milestone 2.0.1, this needs to be part of our update release as we have reports from users without Staypuft. ~~But we need another patch I think - a migration that will remove orpan records perhaps?~~

lzap:
~~We should do the migration fix as part of http://projects.theforeman.org/issues/9488 I guess.~~

lzap:
Sorry disregard my comments about deleting stuff, mixed two bugs together :-)

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -76,19 +76,12 @@ def create
...
-          if update_response
-            delete_discovery_attribute_set(@host.id)
-          end
+          @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host)
+          forward_request_url
+          update_response = @host.update_attributes(params[:discovered_host])
+          process_response update_response
         end
-        process_response update_response
       end
```
lzap, line 85: I think this is possible bug for the token bug @orrabin @GregSutcliffe ^^

Here we were doing the update_attributes call outside the transaction, now it's in. Definitely not a good practice anyway. This patch fixes similar patterns in all our codebase.

orrabin, line 85: @lzap it fixes the token bug at least for my scenario.
I'm closing the other PR

orrabin:
@lzap this needs rebase, also retested against develop and provisioning discovered hosts doesn't create managed hosts. The host is saved but with `managed = false, build = false`

lzap:
Found the issue, fixed, rebased. It was caused by incorrect conflict manual merge.

``` diff
diff --git a/app/services/host_converter.rb b/app/services/host_converter.rb
index b6e80ff..b881cd7 100644
--- a/app/services/host_converter.rb
+++ b/app/services/host_converter.rb
@@ -7,6 +7,7 @@ class ForemanDiscovery::HostConverter
     end
     host = original_host.becomes(::Host::Managed)
     host.type = 'Host::Managed'
+    host.managed = set_managed if set_managed
     host.primary_interface.managed = set_managed if set_managed
     host.build = set_build if set_build
     # this gets rolled back when anything in the transaction fails
```

orrabin:
@lzap maybe add a comment or just remove the default for `set_managed`? it took me a moment to understand how this caused the issue

orrabin:
thanks @lzap, merged as 4376e3cae7439a28ff965de4c649f775f224a631

PR #163, opened 2015-03-01
Fixes #9597 - deleting host token when provision fails

orrabin:


orrabin:
DO NOT MERGE for now - needs tests and a migration to fix existing hosts

lzap:
Do you reproduce this even with https://github.com/theforeman/foreman_discovery/pull/161 ?

orrabin:
@lzap yes

app/controllers/discovered_hosts_controller.rb
```diff
@@ -80,6 +80,7 @@ def update
           delete_discovery_attribute_set(@host.id)
           process_success :success_redirect => host_path(@host), :redirect_xhr => request.xhr?
         else
+          @host.expire_token
```
GregSutcliffe, line 83: Trying to follow the logic here... is the problem that a token is created during the attempt to provision the host, but not deleted if it fails? If so, why does the transaction not catch that?

orrabin, line 83: That's exactly what happens, the token is created once you hit 'submit' on the edit form because the previously discovered host causes this line to be true: https://github.com/theforeman/foreman/blob/develop/app/observers/host_observer.rb#L14
then when you cancel the token is not deleted.
I'm not sure why the transaction doesn't roll back the change, I'm guessing it's has something to do with these changes being on the managed host and not the discovered one.

GregSutcliffe, line 83: That's interesting, and no, I don't have an answer either. It seems very odd. Given the scenario, I'm :+1: for the patch, but add a comment about _why_ so we can revisit it after the Rails 4 upgrade - maybe it'll be fixed ;)

orrabin:
@lzap, @GregSutcliffe do you have any idea how to write a test for this? I've tried but I haven't been able to recreate the error in a test

GregSutcliffe:
@orrabin i guess you've already tried provisioning a host in some way that fails? Be careful, some of the test types disable transactions....

orrabin:
@GregSutcliffe I tried to update directly and make it fail but no token was created

app/controllers/discovered_hosts_controller.rb
lzap, line 83: Maybe we missed a case (UI/CLI) when an explicit transaction is not created. We now explicitly check them, therefore it is not possible to change to discovered host without a transaction.

db/migrate/20150302112545_remove_duplicate_tokens_from_hosts.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveDuplicateTokensFromHosts < ActiveRecord::Migration
+  def up
+    Token.where('id not in (?)', Host::Managed.all.map(&:token).compact).delete_all
+  end
+
+  def down
+  end
+end
```
lzap, line 8: The migration looks good.

lzap:
Just push the test here, let's see what you have and what can we do to get it working.

orrabin:
@lzap This isn't even always reproducible for me - only on a db where I haven't touched the token table. I wrote a test that updated the host and failed but it didn't create a token. I don't have a use case that always causes a problem and can be translated into a test.

orrabin:
no need for a workaround since the transactions have been fixed in: https://github.com/theforeman/foreman_discovery/pull/161

PR #165, opened 2015-03-04
refs #9572 - api changes for cli

orrabin:


app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -31,7 +31,7 @@ def show
           param :name, String, :required => true
           param :search, String, :required => true
           param :hostgroup_id, Integer, :required => true
-          param :hostname, String, :required => true
+          param :hostname, String
```
lzap: Why we do this? It was required, but blank was allowed.

app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -31,7 +31,7 @@ def show
...
           param :max_count, Integer
```
tstrachota: Isn't it confusing to have input parameter `max_count` and return the same value renamed as `hosts_limit`?

app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -52,7 +52,7 @@ def create
           param :name, String, :required => true
```
tstrachota: The whole `discovery_rule` hash could be replaced by `param_group :discovery_rule, :as => :update`.

tstrachota:
I'm no expert in discovery, the api parameters might be obvious to others but it could be worth adding some descriptions e.g. to `search` and `priority`.

app/controllers/api/v2/discovery_rules_controller.rb
orrabin: @tstrachota I didn't want to change the model which already uses `:max_count` but the UI shows `Hosts Limit` so I wanted the api to make sense

app/controllers/api/v2/discovery_rules_controller.rb
orrabin: @lzap is it still required? I think we changed it once we allowed blank:
https://github.com/theforeman/foreman_discovery/blob/develop/app/models/discovery_rule.rb#L12

lzap: You are right, correct.

app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -29,9 +29,9 @@ def show
       def_param_group :discovery_rule do
         param :discovery_rule, Hash, :action_aware => true do
           param :name, String, :required => true
-          param :search, String, :required => true
-          param :hostgroup_id, Integer, :required => true
-          param :hostname, String, :required => true
+          param :search, String, :required => true, :desc => N_("query to match discovered hosts for the particular rule.")
```
orrabin: @lzap I tried to change a little the text you put in the documentation so it's shorter. Do you have a better explanation for search?

lzap: That is very fine.

app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -29,9 +29,9 @@ def show
...
+          param :hostgroup_id, Integer, :required => true, :desc => N_("puts the rules in order, low numbers go first. Must be greater then zero.")
+          param :hostname, String
           param :max_count, Integer
           param :priority, Integer
           param :enabled, :bool
```
lzap: Can you add descriptions for all of them? It's all documented. Hostname is also interesting, max_count as well...

orrabin:
@lzap I added descriptions and rebased. Please check the descriptions are clear.

lzap:
Nitpick - remove trailing dots where not necessary.

app/controllers/api/v2/discovery_rules_controller.rb
```diff
@@ -28,13 +28,13 @@ def show
...
-          param :search, String, :required => true
-          param :hostgroup_id, Integer, :required => true
-          param :hostname, String, :required => true
-          param :max_count, Integer
-          param :priority, Integer
-          param :enabled, :bool
+          param :name, String, :required => true, :desc => N_("represents rule name shown to the users.")
+          param :search, String, :required => true, :desc => N_("query to match discovered hosts for the particular rule.")
+          param :hostgroup_id, Integer, :required => true, :desc => N_("the hostgroup that is used to auto provision a host")
+          param :hostname, String, :desc => N_("defines a pattern to assign human-readable hostnames to the matching hosts (default is macMACADDRESS)")
```
lzap: I'd remove the default explanation.

lzap:
Ok otherwise.

orrabin:
@lzap I removed the dots and default explanation

lzap:
Thanks, merged as 4e6b799

PR #167, opened 2015-03-09
i18n - pulling from tx

lzap:
Do you see any problem @domcleal? Looks sane... This is for 3.0.0


domcleal:
@lzap have you also extracted them recently?  I can't see it...

lzap:
Yeah, no new strings for the 3.0 :-P

lzap:
May I ask for a quick merge @orrabin ? Looks like all is set for 3.0.0.

lzap:
@domcleal just to confirm, I did:

```
$ rake plugin:gettext[foreman_plugin]
```

No changes in the POT file...

domcleal:
s/foreman_plugin/foreman_discovery/ and I think you may see the timestamp inside the POT update at least.

But yeah, I think you're right, in which case it's OK.

lzap:
Yeah thanks for confirmation, saw those timestamps and things, but I was only interested in the POT updates. So I fetched from Transiflex and it looks we have three languages complete now. Thanks! Ori, good to merge.

locale/de/foreman_discovery.po
```diff
@@ -1,18 +1,17 @@
...
 # 
 # Translators:
 # Arnold Bechtoldt <mail@arnoldbechtoldt.com>, 2014
-# Ettore Atalan <atalanttore@googlemail.com>, 2014
+# Ettore Atalan <atalanttore@googlemail.com>, 2014-2015
 # simon11 <simon.stieger.98@live.de>, 2014
 msgid ""
 msgstr ""
-"Project-Id-Version: foreman_discovery 1.4.0.rc2\n"
+"Project-Id-Version: foreman_discovery 2.0.0.rc1\n"
```
orrabin, line 11: is it still 2.0.0 here?

domcleal, line 11: This plugin's version number still says 2.0.0.rc1, so it'll use that.

orrabin, line 11: thanks, merging

orrabin:
thanks @lzap! merged as de513a059bf24f4733746fe7bf3b76ded24d20ce

PR #168, opened 2015-03-10
Fixes #9488 - migration to delete discovery_attribute_set from managed hosts

orrabin:
The problems with deleting `discovery_attribute_set` when provisioning a host should be fixed by: https://github.com/theforeman/foreman_discovery/pull/161.
This PR is to fix the existing managed hosts that can't be deleted because they are still attached to a `discovery_attribute_set`


lzap:
Causing STATEMENT INVALID on SQLITE3. This must be written as:

```
DiscoveryAttributeSet.where(:id => DiscoveryAttributeSet.joins(:host).where(:'hosts.managed' => true).pluck(:id))
```

lzap:
Tested. Thanks. 1bc77ee

PR #169, opened 2015-03-18
2.0 cherry picks

orrabin:
due to the amount of conflicts in the first PRs creating one branch with all the cherry-picks so it can be reviewd

### DIFF
--- a/app/services/host_converter.rb
+++ b/app/services/host_converter.rb
@@ -1,0 +1,30 @@
+class ForemanDiscovery::HostConverter
+
+  # must be called from SQL transaction 
+  def self.to_managed(original_host, set_managed = true, set_build = true)
+    if ActiveRecord::Base.connection.open_transactions <= 0
+      raise "This method must be executed with explicit transaction"
+    end
+    host = original_host.becomes(::Host::Managed)
+    host.type = 'Host::Managed'
+    # the following flags can be skipped when parameters are set to false
+    if set_managed
+      host.managed = set_managed
+      host.managed = set_managed
+    end
+    host.build = set_build if set_build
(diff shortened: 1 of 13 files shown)


test/unit/discovered_extensions_test.rb
```diff
@@ -73,7 +73,7 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     host = Host::Discovered.import_host_and_facts(facts).first
     r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc")
     perform_auto_provision host, r1
-    assert_equal host.managed, true
+    assert_equal host.primary_interface.managed, true
```
lzap: This will likely fail, `primary_interface` is not present in 1.7 Foreman.

app/services/host_converter.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.to_managed(original_host, set_managed = true, set_build = true)
+    if ActiveRecord::Base.connection.open_transactions <= 0
+      raise "This method must be executed with explicit transaction"
+    end
+    host = original_host.becomes(::Host::Managed)
+    host.type = 'Host::Managed'
+    # the following flags can be skipped when parameters are set to false
+    if set_managed
+      host.managed = set_managed
+      host.primary_interface.managed = set_managed
```
lzap: I think this should be

```
host.managed = set_managed
```

lzap:
Other than that - looks ok. Please test against Foreman 1.7 and Sat 6.1 and also run unit tests. Wow I did not expect so many conflicts. :-(

PR #170, opened 2015-03-19
refs #9546 - moving host_converter to proper directory for rails

orrabin:


lzap:
Thanks all green. d01c698

PR #171, opened 2015-03-23
Fixes #9870 - Auto-provisioning a host, when no rule is defined, raises ...

orrabin:
...proper error


lzap:
Merged as 5188cf1 for Discovery 4.0 :-)

PR #172, opened 2015-03-24
Fixes #9881 - Adding organization/location to discovery_rules

orrabin:


app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -14,8 +14,13 @@ def find_discovery_rule host
...
           if Host::Discovered.where(:id => host.id).search_for(rule.search).size > 0
-            Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
-            return rule
+            if rule.organizations.include?(host.organization) && rule.locations.include?(host.location)
+              Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
+              return rule
+            else
+              Rails.logger.warn "Rule #{rule.name} (#{rule.id}) can not be applied due to a difference in organization/location from host #{host.name} (#{host.id})"
+            end
+
```
lzap: Extra line.

app/models/discovery_rule.rb
```diff
@@ -26,8 +25,16 @@ class DiscoveryRule < ActiveRecord::Base
...
+                  with_taxonomy_scope do
+                    order("discovery_rules.name")
+                  end
+                }
+
   def default_int_attributes
-   self.max_count ||= 0
-   self.priority  ||= 0
+    self.max_count ||= 0
+    self.priority  ||= 0
```
lzap: If we are going to cherry pick this will likely cause conflicts, but I am fine with nicer code :-)

app/views/discovery_rules/_form.html.erb
```diff
@@ -1,34 +1,58 @@
...
+
+          <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @discovery_rule %>
+        </div>
+
+    <% else %>
+        <%= text_f f, :name %>
+    <% end %>
+
+    <%= submit_or_cancel f %>
 <% end %>
```
lzap, line 58: Ok this will cause pains downstream when cherry picking. I'd vote against reformat, but up to you if you want to keep this.

test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -75,17 +77,28 @@ def test_auto_provision_success
...
 
-  def test_auto_provision_success_and_delete
+  def test_auto_provision_with_wrong_org_or_loc_fail
     disable_orchestration
     facts = @facts.merge({"somefact" => "abc"})
     host = Host::Discovered.import_host_and_facts(facts).first
     FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc", :hostgroup => hostgroups(:common))
     post :auto_provision, { :id => host.id }
+    assert_response 200
+  end
```
lzap, line 93: Shouln't this test fail rather than give 200?

app/views/discovery_rules/_form.html.erb
lzap, line 58: Have you changed any line here?

orrabin, line 58: I moved everything into primary tab. 
This isn't reformatting, it is needed since I added tabs for organization and location.

ohadlevy:
what happens when a rule matches a host which is not in the org? is it filter?
should rules be extended to actually define the org automatically instead? (if that's possible at all)

test/functional/api/v2/discovered_hosts_controller_test.rb
orrabin, line 93: it doesn't fail, it just adds an error message but there is a line missing here with the error that disappeared so I will add it.

orrabin, line 93: This can only be tested correctly after: https://github.com/theforeman/foreman_discovery/pull/171 is merged. For now, it just succeeds with no error

orrabin:
@ohadlevy we filter it: https://github.com/theforeman/foreman_discovery/pull/172/files#diff-2e9f4565997ec456d7eb9c4bbcfca096R17
I'm not sure we should try to define the org through the rule - what happens if the rule wasn't defined correctly and has no org?

orrabin:
@lzap I fixed your comments, please review again

lzap:
@ohadlevy we've talked about this already, yes this is possible, but out of scope of 2.0.1/6.1. We are fixing bugs here, we have a ticket for orgs/locs per rule for future versions.

lzap:
Tested, works fine. Thanks. Merged as c8b6595

domcleal:
Does discovery require orgs/locs, or was this tested with orgs/locs disabled?

I don't use discovery to test, but the commit looks strange in a couple of places to me:
1. https://github.com/theforeman/foreman_discovery/commit/c8b659528940c33df38d67bfc9f75bc84b6ea2bb#diff-2e9f4565997ec456d7eb9c4bbcfca096R17 - with orgs/locs disabled, would a host even have one set to be able to check if it's in the rules?
2. If orgs/locs are disabled, this form is disabled apart from a name text box: https://github.com/theforeman/foreman_discovery/commit/c8b659528940c33df38d67bfc9f75bc84b6ea2bb#diff-a3123e7a434f6d9a4691681e36a25470R4

lzap:
Oh good catch, I haven't tested with org/loc disabled. Ori can you please add also one quick test for this case so we catch regression? Autoprovisioning might be the best workflow, success case.

orrabin:
@lzap this will also cause all of the rules to stop working since org/loc was not selected for them. Maybe I'll add a migration to pick the org/loc from the hostgroup? or all of them?

lzap:
Sorry what is causing this? I don't think we support turning org/loc off once they have been enabled already. Do we? @domcleal ?

domcleal:
We don't say it's unsupported anywhere, you should be able to turn them on and off as you please I think - core works that way.

Usually off, everything's better then.

domcleal:
By the way, I filed http://projects.theforeman.org/issues/9910.  (I'm unsure how closely you monitor Redmine - please do, as users have been filing issues against Discovery since the major release.)

PR #173, opened 2015-03-25
Fixes #9895 - user in Discovery Reader role should see discovery rules

orrabin:
after adding this I needed to re-add the filters with the correct resource and then the role started working for me.

### DIFF
--- a/db/migrate/20150512150432_remove_old_discovery_reader_permissions.rb
+++ b/db/migrate/20150512150432_remove_old_discovery_reader_permissions.rb
@@ -1,0 +1,9 @@
+class RemoveOldDiscoveryReaderPermissions < ActiveRecord::Migration
+  def up
+    Permission.where("name like '%_discovery_rules' and resource_type is null").destroy_all
+  end
+
+  def down
+  end
+end
+
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -67,23 +67,23 @@ module ForemanDiscovery
           permission :view_discovery_rules, {
             :discovery_rules => [:index, :show, :auto_complete_search],
             :"api/v2/discovery_rules" => [:index, :show]
-          }
+          }, :resource_type => 'DiscoveryRule'
           permission :new_discovery_rules, {
             :discovery_rules => [:new, :create],
             :"api/v2/discovery_rules" => [:create]
-          }
+          }, :resource_type => 'DiscoveryRule'
           permission :edit_discovery_rules, {
             :discovery_rules => [:edit, :update, :enable, :disable],
             :"api/v2/discovery_rules" => [:create, :update]
-          }
+          }, :resource_type => 'DiscoveryRule'
           permission :execute_discovery_rules, {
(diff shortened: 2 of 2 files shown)


lzap:
Can you add couple of tests for this? We can inspire from Foreman Core where we have some permission tests already.

lzap:
I think we need to do a db migration here or whatever is needed to get this working since we already released discovery 2.0 with this bug. Also, as part of this migration can you please delete the old role called `Discovery`? Do this only in case there are no users associated with it, otherwise skip. I have renamed this role to `Discovery Manager` for 2.0 and forgot to do this.

lzap:
Ok disregard part of the old role called `Discovery` from previous comment as I've already included this fix in #176 but we still need to delete extra permissions which has not been deleted. This will cause issues in the future. Before this patch:

```
[15] pry(main)> Permission.where("name like '%_discovery_rules'").select("name, resource_type")
2015-03-31 15:05:07 [D]   Permission Load (0.2ms)  SELECT name, resource_type FROM "permissions" WHERE (name like '%_discovery_rules')
=> [#<Permission name: "view_discovery_rules", resource_type: nil>, #<Permission name: "new_discovery_rules", resource_type: nil>, #<Permission name: "edit_discovery_rules", resource_type: nil>, #<Permission name: "execute_discovery_rules", resource_type: nil>, #<Permission name: "delete_discovery_rules", resource_type: nil>]
```

After this patch:

```
[16] pry(main)> Permission.where("name like '%_discovery_rules'").select("name, resource_type")
2015-03-31 15:06:01 [D]   Permission Load (0.4ms)  SELECT name, resource_type FROM "permissions" WHERE (name like '%_discovery_rules')
=> [#<Permission name: "view_discovery_rules", resource_type: nil>, #<Permission name: "new_discovery_rules", resource_type: nil>, #<Permission name: "edit_discovery_rules", resource_type: nil>, #<Permission name: "execute_discovery_rules", resource_type: nil>, #<Permission name: "delete_discovery_rules", resource_type: nil>, #<Permission name: "view_discovery_rules", resource_type: "DiscoveryRule">, #<Permission name: "new_discovery_rules", resource_type: "DiscoveryRule">, #<Permission name: "edit_discovery_rules", resource_type: "DiscoveryRule">, #<Permission name: "execute_discovery_rules", resource_type: "DiscoveryRule">, #<Permission name: "delete_discovery_rules", resource_type: "DiscoveryRule">]
```

We need to delete those. To trigger the error add a new testing role with a permission you have changed in the `engine.rb`:

```
role "Discovery Executor Test", [:execute_discovery_rules]
```

Restart and voilla:

```
/home/lzap/work/foreman/app/models/role.rb:131:in `add_permissions': some permissions were not found (ArgumentError)
        from /home/lzap/work/foreman/app/models/role.rb:146:in `add_permissions!'
        from /home/lzap/work/foreman/app/services/foreman/plugin.rb:198:in `block in role'
```

ohadlevy:
ping?

orrabin:
@lzap I added a migration, can you please retest?

lzap:
Merged as 0d9cee21e5360614c517176ab0c47ed312018246, thank you!

PR #175, opened 2015-03-30
refs #9881 - discovery_rules works when organizations and locations are disabled

orrabin:
This PR fixed @domcleal's comments on https://github.com/theforeman/foreman_discovery/pull/172

@lzap - this isn't completely ready yet. I opened this so we can test using it and decide what to do when discovery_rules doesn't have a organization/location (when they are enabled and disabled)? when the host doesn't have them? when neither have them?
I will write tests according to the conclusions to these questions.

### DIFF
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -14,7 +14,7 @@ module Foreman::Controller::DiscoveredExtensions
         # try to match the search
         begin
           if Host::Discovered.where(:id => host.id).search_for(rule.search).size > 0
-            if rule.organizations.include?(host.organization) && rule.locations.include?(host.location)
+            if validate_rule_by_taxonomy(rule, host)
               Rails.logger.info "Match found for host #{host.name} (#{host.id}) rule #{rule.name} (#{rule.id})"
               return rule
             else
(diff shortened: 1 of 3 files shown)


orrabin:
[test]

domcleal:
The failures currently are due to #9940.

orrabin:
@domcleal thanks, I wont run the tests again for now

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -29,7 +29,17 @@ def find_discovery_rule host
...
+    end
+
+    def validate_rule_by_taxonomy rule, host
+      if SETTINGS[:organizations_enabled]
+        return false unless rule.organizations.include?(host.organization)
+      end
+      if SETTINGS[:locations_enabled]
+        return false unless rule.locations.include?(host.location)
+      end
+      return true
```
lzap: Should be only `true` otherwise the cop will scream out

lzap:
This is fine, if user enable org/loc she needs to reboot hosts. We can document that.

lzap:
Hmmm #9940 looks like resolved, [test]

lzap:
Ok it works just fine, the edge case is when user discover hosts with o/l disabled, then enable it. Auto provision won't trigger rules because discovered hosts has no o/l set. Manual provisioning will not work either because o/l dropdowns are disabled (limitation of editing existing hosts with o/l enabled - RFE in the Core). Merged as 47ecc19 thanks.

domcleal:
@lzap sounds like you're describing http://projects.theforeman.org/issues/4426

PR #178, opened 2015-04-20
Fixes #9607 - discovery search succeeds on ip, mac and subnet

orrabin:


app/models/host/discovered.rb
```diff
@@ -3,7 +3,7 @@
 class Host::Discovered < ::Host::Base
   include ScopedSearchExtensions
 
-  belongs_to :subnet
+  has_one :subnet, :through => :primary_interface
   belongs_to :hostgroup
```
ohadlevy, line 6: btw: do we care about other subnets? (assuming the host has multiple interfaces)

lzap, line 6: Hmmm but Host:Base does already have this:

```
app/models/host/base.rb:25:    has_one :subnet, :through => :primary_interface
```

Why do we add this once again?

orrabin, line 6: We don't. Looks like the belongs_to was the problem, I'll remove it completely.

app/models/host/discovered.rb
```diff
@@ -2,8 +2,7 @@
 
 class Host::Discovered < ::Host::Base
   include ScopedSearchExtensions
-
-  belongs_to :subnet
+
```
lzap, line 5: Trailing whitespace? Other than that :+1: will test later...

orrabin, line 5: it was there before on line 5

app/models/host/discovered.rb
```diff
@@ -13,8 +12,8 @@ class Host::Discovered < ::Host::Base
 
   scoped_search :on => :name, :complete_value => true, :default_order => true
   scoped_search :on => :last_report, :complete_value => true
-  scoped_search :on => :ip, :complete_value => true
-  scoped_search :on => :mac, :complete_value => true
+  scoped_search :in => :primary_interface, :on => :ip, :complete_value => true
+  scoped_search :in => :primary_interface, :on => :mac, :complete_value => true
```
ohadlevy, line 16: don't we need to add back the subnet for search?
also, should we search in => :interfaces instead of primary?

orrabin, line 16: we don't need the subnet because of:  `has_one :subnet, :through => :primary_interface` in host/base.rb
we only show the ip and mac from primary interface in the search results so wouldn't that be a little confusing to search on all interfaces?

lzap:
To test this one on develop, we need to merge this one first: https://github.com/theforeman/foreman_discovery/pull/183

app/models/host/discovered.rb
lzap, line 5: Why the line 6 was removed?

lzap, line 5: In the other patch, I am doing:

```
  has_one :subnet, :through => :primary_interface
```

lzap:
Ok I have tested this together with patch https://github.com/theforeman/foreman_discovery/pull/183 and it works, if you leave the `:subnet` association untouched. Just keep the two scoped_search lines there.

app/models/host/discovered.rb
lzap, line 5: Can can you remove this hunk (leave lines 4-6 untouched in this patch). I am fixing this in a different patch.

lzap:
I have incorporated this patch in this PR: https://github.com/theforeman/foreman_discovery/pull/183 Thanks!

PR #179, opened 2015-05-05
Fixes #10365 - Removing all tokens when there are no managed hosts

orrabin:


lzap:
Ok I agree but I want to make sure this is migrated correctly in one migration and not in two (there is still a chance if other migration fails or something and user creates token or host that this will not work). Therefore can you do:

```
class RemoveDuplicateTokensFromHosts < ActiveRecord::Migration
  def up
    existing_tokens = Host::Managed.all.map(&:token).compact
    if existing_tokens.empty?
      Token.delete_all
    else
      Token.where('id not in (?)', existing_tokens).delete_all
    end
  end

  def down
  end
end
```

orrabin:
@lzap I agree, changed it

lzap:
Merged as 52cd340ad7cf31fc54948bfca844b162430c7f40, thank you!

PR #181, opened 2015-05-06
Fixes #10375 - entering edit form deletes attributes

lzap:
We simply deleted attributes on edit form render. I added a comment explaining
the method and how it should be used.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -29,7 +29,7 @@ class DiscoveredHostsController < ::ApplicationController
   # as a parameter and use refresh_facts to get the rest
   def create
     Taxonomy.no_taxonomy_scope do
-      host, imported = Host::Discovered.new(:ip => get_ip_from_env).refresh_facts
+      _, imported = Host::Discovered.new(:ip => get_ip_from_env).refresh_facts
       respond_to do |format|
         format.yml {
           if imported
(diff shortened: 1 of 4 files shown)


app/controllers/discovered_hosts_controller.rb
```diff
@@ -60,7 +60,7 @@ def destroy
 
   def edit
     Host.transaction do
-      @host = ::ForemanDiscovery::HostConverter.to_managed(@host) unless @host.nil?
+      @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false, false) unless @host.nil?
```
orrabin: is this the only place we shouldn't delete the attributes?
we have `to_managed` in `perform_auto_provision` in `discovered_extensions.rb`, if the auto_provision fails what happens?

orrabin:
I haven't has a chance to test with failing auto provision, I tested with provision (successful and unsuccessful) and successful auto privision and it works as expected.
Also, nitpick - can you change  the commit message to - entering edit form doesn't delete attributes?

app/controllers/discovered_hosts_controller.rb
lzap: Yes, this is the only place.

lzap:
Re-phrased.

ohadlevy:
missing tests?

orrabin:
@lzap I was able to delete the attributes before a successful provision by:
1. click provision
2. try to save
3. cancel
just entering the edit form or clicking cancel right away didn't delete the attributes

lzap:
@orrabin as per our call from this morning, you are actually working on a test to reproduce this, right? I will wait then...

orrabin:
@lzap, not this one but we need to add a test for this too

lzap:
Test added.

app/controllers/discovered_hosts_controller.rb
lzap: @orrabin here I changed the 3rd param to `false`

lzap:
Pushed

app/models/host/managed_extensions.rb
```diff
@@ -24,4 +25,9 @@ def delReboot
     # last action in the queue.
   end
 
+  def delete_discovery_attribute_set
+    return if new_record?
+    DiscoveryAttributeSet.destroy_all(:host_id => host_id) if type_changed?
```
orrabin: maybe add if old was discovered host?

orrabin:
tested locally and it looks good, waiting for Jenkins

orrabin:
looks like are tests are not so reliable right now: http://projects.theforeman.org/issues/10548
I tested manually and we need to rewrite the tests and check this but for now merging as is

orrabin:
merged as 009bb13e093bef968461ca4121e97fbfa1d2d370, thanks @lzap!

PR #186, opened 2015-05-19
fixes #10554 - remove the UI route for creating a discovered host.

ohadlevy:


orrabin:
@ohadlevy the title says 'non UI route' but you are removing the UI route.

ohadlevy:
@orrabin effectively i'm removing the non ui (e.g json format) from the ui controllers leaving only the ability to use the api controllers or the actions that are visible in the ui.
how would you want me to rephrase ?

ohadlevy:
@orrabin scratch that, I mixed it up with #187.

this removes methods that are not really used by the discovery image anymore.
so, since its not visible in the UI, I'm removing it from the code?

lzap:
Thanks, will do during our cleanup round.

lzap:
Merged as 1f30d55ff87cb2a04b421a2b119275d40e268411, thank you!

PR #189, opened 2015-05-27
Refs #9870 - adding error response code when auto-provision fails with no discovery rule

orrabin:


lzap:
Fixes #171 - Can you file new upstream issue please and amend the issue number? LGTM

orrabin:
@lzap it's not a new issue, #171 showed an error in response to the API call but returned 200 at the end so the only way to know there was an error was reading the message. 
Hammer (and anything else using the API and looking at the server's response) would think it was a successful call and this PR fixes the return code.

lzap:
Merged as ff125ee3f06e9f80532eae29d0fb22be11f6e11f, thank you!

PR #195, opened 2015-06-21
Fixes #10892 - Multiple columns support for discovery_fact_column

imriz:
This is a minor change to allow users to add multiple facts to the discovery table.
I find it useful, maybe you will too :)

### DIFF
--- a/app/views/discovered_hosts/_discovered_host.html.erb
+++ b/app/views/discovered_hosts/_discovered_host.html.erb
@@ -6,7 +6,7 @@
 <% unless defined? hide_disk %>
 <td class="hidden-tablet hidden-xs"><%= discovery_attribute(host, :disk_count) %></td>
 <td class="hidden-tablet hidden-xs"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
-<% if Setting['discovery_fact_column'].present? %>
-  <td class="hidden-tablet hidden-xs"><%= host.facts_hash[Setting['discovery_fact_column']] || 'N/A' %></td>
+<% Setting::Discovered.discovery_fact_column_array.each do |fact_column| %>
+<td class="hidden-tablet hidden-xs"><%= host.facts_hash[fact_column.strip] || 'N/A' %></td>
 <% end %>
 <% end %>
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -45,4 +45,15 @@ class Setting::Discovered < ::Setting
 
   end
 
+  def self.discovery_fact_column_array
+    return [] if !Setting['discovery_fact_column'].present?
+    list = []
+    Setting['discovery_fact_column'].to_s.split(",").each do |value|
+      list << value.strip
+    end
+  rescue => error
+    logger.warn "Failed to parse comma delimited list [%s] into array. Error: %s" % [list,error]
+  ensure
+    list
+  end
 end
--- a/app/views/discovered_hosts/_discovered_hosts_list.html.erb
+++ b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
@@ -10,8 +10,8 @@
     <th class="hidden-tablet hidden-xs"><%= sort :memory, :as => _('Memory') %></th>
     <th class="hidden-tablet hidden-xs"><%= sort :disk_count, :as => _('Disk count') %></th>
     <th class="hidden-tablet hidden-xs"><%= sort :disks_size, :as => _('Disks size') %></th>
-    <% if Setting['discovery_fact_column'].present? %>
-      <th class="hidden-tablet hidden-xs"><%= Setting['discovery_fact_column'].capitalize %></th>
+    <% Setting::Discovered.discovery_fact_column_array.each do |fact_column| %>
+      <th class="hidden-tablet hidden-xs"><%= fact_column.strip.capitalize %></th>
     <% end %>
     <% if SETTINGS[:locations_enabled] -%>
       <th class="hidden-tablet hidden-xs"><%= sort :location, :as => _('Location') %></th>


lzap:
Nice, can you amend the commit message to:

```
Fixes #10892 - Multiple columns support for discovery_fact_column
```

Tested this and it does not work when you specify columns with space:

```
fact1, fact2
```

Can we strip down whitespace? Other than that - :+1:

app/views/discovered_hosts/_discovered_host.html.erb
```diff
@@ -7,6 +7,8 @@
 <td class="hidden-tablet hidden-xs"><%= discovery_attribute(host, :disk_count) %></td>
 <td class="hidden-tablet hidden-xs"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
 <% if Setting['discovery_fact_column'].present? %>
-  <td class="hidden-tablet hidden-xs"><%= host.facts_hash[Setting['discovery_fact_column']] || 'N/A' %></td>
+<% Setting['discovery_fact_column'].to_s.split(",").each do |fact_column| %>
```
ohadlevy: which data in settings can break this feature? just wondering if we need to be a bit more protective? maybe consider extracting it to a helper instead?

imriz: I don't think it is more vulnerable than it previously was. At worse, it will just display an empty column(s) since the fact(s) cannot be found.

imriz:
> Nice, can you amend the commit message to:
> 
> Fixes #10892 - Multiple columns support for discovery_fact_column

Done

> Tested this and it does not work when you specify columns with space:
> 
> fact1, fact2
> Can we strip down whitespace? Other than that -

Done

app/views/discovered_hosts/_discovered_host.html.erb
lzap: > I don't think it is more vulnerable than it previously was. At worse, it will just display an empty column(s) since the fact(s) cannot be found.

Ohad is right, ideally there is a method with rescue catching all
exceptions returning empty array.

## 

Later,
 Lukas #lzap Zapletal

imriz: > Ohad is right, ideally there is a method with rescue catching all
> exceptions returning empty array.

I will try to work on it tomorrow and rebase.

imriz:
@ohadlevy @lzap I've moved it to an helper method. Hope it's good now.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -44,4 +44,17 @@ def discovery_attribute(host, attr, default_value = _('N/A'))
...
+      begin
+        list.to_s.split(",").each do |value|
+          arr << value.to_s.strip
+        end
+      rescue Error => error
+        logger.warn "Failed to parse comma delimited list [%s] into array. Error: %s" % [list,error]
+      end
+    end
+    arr
+  end
```
ohadlevy: how about

``` ruby
def disc_turn_comma_delimited_to_array(list)
  return if list.empty?
  list.to_s.split(",").map do |value|
    value.strip
  end
rescue Error => error
  logger.warn "Failed to parse comma delimited list [%s] into array. Error: %s" % [list,error]
end
```

ohadlevy: not sure about Error, is that the right exception?

imriz: > not sure about Error, is that the right exception?

Fixed to catch StandardError.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -44,4 +44,12 @@ def discovery_attribute(host, attr, default_value = _('N/A'))
     host.try(:discovery_attribute_set).try(attr) || default_value
   end
 
+  def disc_turn_comma_delimited_to_array(list)
+    return if list.empty?
+    list.to_s.split(",").map do |value|
+      value.strip
+    end
+  rescue => error
+    logger.warn "Failed to parse comma delimited list [%s] into array. Error: %s" % [list,error]
```
ohadlevy: needs to be extracted for i18n

lzap: We don't extract error log messages, do we?

lzap: Anyway, we must return `[]` here if something bad happens.

lzap:
Ok I think instead of coding in templates and creating "low-level" helper methods, I think the correct approach is:
- create new helper method `Setting::Discovered::discovery_fact_column_array`
  - checks if `discovery_fact_column` is present
  - _always_ returns an array (empty when not present or on parsing errors)
- get rid of `if Setting['discovery_fact_column'].present?` checks in ERB files
- use this new method in the templates

imriz:
> Ok I think instead of coding in templates and creating "low-level" helper methods, I think the correct approach is:
> 
> create new helper method Setting::Discovered::discovery_fact_column_array
> checks if discovery_fact_column is present
> always returns an array (empty when not present or on parsing errors)
> get rid of if Setting['discovery_fact_column'].present? checks in ERB files
> use this new method in the templates

Done as requested.

Thanks :)

dLobatog:
[test]

lzap:
Looks good, the last item remaining. Can you reword the description? `_("Show fact as an extra column in the discovered hosts list")` to something other like `Extra facter columns to show in host lists (separate by comma)` or similar.

shlomizadok:
:100:  @imriz thanks for this contribution (one of many, I hope).
@orrabin will change the text on merge :)

orrabin:
merged as 3ced88b1ea73ee1bf39945f04f10210cb8fee7fd thanks @imriz!

shlomizadok:
@imriz תודה רבה :smile:

lzap:
Thanks all involved. Díky moc!

PR #197, opened 2015-07-14
Fixes #10898 - review auto provisioning and rules permissions

lzap:
This patch fixes auto provisioning permissions and also aligns
action_permission strings with what is expected in security blocks. I
basically reviewed all our permissions and introduced
`submit_discovered_hosts` and `auto_provision_discovered_hosts`. Also,
permission `new_discovery_rules` was renamed to `create_*` because this was
never working (our permission stack expects `create_` prefix by default).

Since existing roles are never changed via the `role` DSL, I am attaching a
migration that adds new permissions (they are all new so we can do this
without security considerations) and deletes the renamed one.

We must add this to release notes for the next release, users need to update
their own roles.

@orrabin this needs extensive testing. We should consider some testing
scenarios similar to core I think.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -125,7 +125,7 @@ class DiscoveredHostsController < ::ApplicationController
   end
 
   def auto_provision
-    @host.transaction do
+    Host.transaction do
       if rule = find_discovery_rule(@host)
         if perform_auto_provision(@host, rule)
           process_success :success_msg => _("Host %{host} was provisioned with rule %{rule}") % {:host => @host.name, :rule => rule.name}, :success_redirect => discovered_hosts_path
@@ -202,16 +202,12 @@ class DiscoveredHostsController < ::ApplicationController
 
   def action_permission
     case params[:action]
-      when 'refresh_facts', 'reboot', 'reboot_all'
-        :view
-      when 'update_multiple_location', 'select_multiple_organization', 'update_multiple_organization', 'select_multiple_location'
+      when 'refresh_facts', 'reboot', 'reboot_all', 'update_multiple_location', 'select_multiple_organization', 'update_multiple_organization', 'select_multiple_location'
         :edit
       when 'submit_multiple_destroy', 'multiple_destroy'
         :destroy
-      when 'auto_provision'
+      when 'auto_provision', 'auto_provision_all'
         :auto_provision
-      when 'auto_provision_all'
-        :auto_provision_all
       else
         super
     end
--- a/db/migrate/20150714144500_review_discovery_permissions.rb
+++ b/db/migrate/20150714144500_review_discovery_permissions.rb
@@ -1,0 +1,17 @@
+class ReviewDiscoveryPermissions < ActiveRecord::Migration
+  def up
+    if (mgr = Role.find_by_name("Discovery Manager"))
+      perms = []
+      perms << "submit_discovered_hosts" if Permission.find_by_name("edit_discovered_hosts")
+      perms << "auto_provision_discovered_hosts" if Permission.find_by_name("provision_discovered_hosts")
+      perms << "create_discovery_rules" if Permission.find_by_name("new_discovery_rules")
+      perms << "destroy_discovery_rules" if Permission.find_by_name("delete_discovery_rules")
+      mgr.add_permissions!(perms)
+    end
+    Permission.find_by_name("new_discovery_rules").try(:destroy)
+  end
+
+  def down
+    # rollback is not supported
+  end
+end
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -45,16 +45,23 @@ module ForemanDiscovery
             :discovered_hosts          => [:index, :show, :auto_complete_search],
             :"api/v2/discovered_hosts" => [:index, :show]
           }, :resource_type => 'Host'
-          permission :provision_discovered_hosts, {
+          permission :submit_discovered_hosts, {
+            :"api/v2/discovered_hosts" => [:facts, :create]
+          }, :resource_type => 'Host'
+          permission :auto_provision_discovered_hosts, {
             :discovered_hosts          => [:auto_provision, :auto_provision_all],
-            :"api/v2/discovered_hosts" => [:create, :auto_provision, :auto_provision_all]
+            :"api/v2/discovered_hosts" => [:auto_provision, :auto_provision_all]
+          }, :resource_type => 'Host'
+          permission :provision_discovered_hosts, {
+            :discovered_hosts          => [:edit, :update],
+            :"api/v2/discovered_hosts" => [:update]
           }, :resource_type => 'Host'
           permission :edit_discovered_hosts, {
-            :discovered_hosts          => [:edit, :update, :update_multiple_location,
+            :discovered_hosts          => [:update_multiple_location,
                                            :select_multiple_organization,
                                            :update_multiple_organization,
                                            :select_multiple_location, :refresh_facts, :reboot, :reboot_all],
-            :"api/v2/discovered_hosts" => [:update, :facts, :refresh_facts, :reboot, :reboot_all]
+            :"api/v2/discovered_hosts" => [:refresh_facts, :reboot, :reboot_all]
           }, :resource_type => 'Host'
           permission :destroy_discovered_hosts, {
             :discovered_hosts          => [:destroy, :submit_multiple_destroy, :multiple_destroy],
@@ -68,7 +75,7 @@ module ForemanDiscovery
             :discovery_rules => [:index, :show, :auto_complete_search],
             :"api/v2/discovery_rules" => [:index, :show]
           }, :resource_type => 'DiscoveryRule'
-          permission :new_discovery_rules, {
+          permission :create_discovery_rules, {
             :discovery_rules => [:new, :create],
             :"api/v2/discovery_rules" => [:create]
           }, :resource_type => 'DiscoveryRule'
@@ -80,15 +87,35 @@ module ForemanDiscovery
             :discovery_rules => [:auto_provision, :auto_provision_all],
             :"api/v2/discovery_rules" => [:auto_provision, :auto_provision_all]
           }, :resource_type => 'DiscoveryRule'
-          permission :delete_discovery_rules, {
+          permission :destroy_discovery_rules, {
             :discovery_rules => [:destroy],
             :"api/v2/discovery_rules" => [:destroy]
           }, :resource_type => 'DiscoveryRule'
         end
 
-        # roles (only added if they don't exist)
-        role "Discovery Manager", [:view_discovered_hosts, :provision_discovered_hosts, :edit_discovered_hosts, :destroy_discovered_hosts, :view_discovery_rules, :new_discovery_rules, :edit_discovery_rules, :execute_discovery_rules, :delete_discovery_rules]
-        role "Discovery Reader", [:view_discovered_hosts, :view_discovery_rules]
+        READER = [
+          :view_discovered_hosts,
+          :view_discovery_rules,
+          :view_organizations,
+          :view_locations,
+        ]
+        MANAGER = READER + [
+          :assign_organizations,
+          :assign_locations,
+          # discovered_hosts
+          :submit_discovered_hosts,
+          :provision_discovered_hosts,
+          :auto_provision_discovered_hosts,
+          :edit_discovered_hosts,
+          :destroy_discovered_hosts,
+          # discovered_rules
+          :create_discovery_rules,
+          :execute_discovery_rules,
+          :edit_discovery_rules,
+          :destroy_discovery_rules,
+        ]
+        role "Discovery Reader", READER
+        role "Discovery Manager", MANAGER
 
         # menu entries
         menu :top_menu, :discovered_hosts, :url_hash => {:controller => :discovered_hosts, :action => :index},
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -100,7 +100,7 @@ module Api
           @discovered_host, state = Host::Discovered.import_host_and_facts(params[:facts])
         end
         if Setting['discovery_auto']
-          @discovered_host.transaction do
+          Host.transaction do
             if state && rule = find_discovery_rule(@discovered_host)
               state = perform_auto_provision(@discovered_host, rule)
             else
@@ -119,7 +119,7 @@ module Api
       param :id, :identifier, :required => true
 
       def auto_provision
-        @discovered_host.transaction do
+        Host.transaction do
           if rule = find_discovery_rule(@discovered_host)
             msg = _("Host %{host} was provisioned with rule %{rule}") % {:host => @discovered_host.name, :rule => rule.name}
             process_response perform_auto_provision(@discovered_host, rule), msg
@@ -217,12 +217,12 @@ module Api
 
       def action_permission
         case params[:action]
-          when 'auto_provision'
+          when 'auto_provision', 'auto_provision_all'
             :auto_provision
-          when 'auto_provision_all'
-            :auto_provision_all
           when 'refresh_facts', 'reboot', 'reboot_all'
             :edit
+          when 'facts', 'create'
+            :submit
           else
             super
         end
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -3,6 +3,7 @@ module Foreman::Controller::DiscoveredExtensions
 
   # return auto provision rule or false when not present
   def find_discovery_rule host
+    raise(::Foreman::Exception.new(N_("Unable to find a discovery rule, no host provided (check permissions)"))) if host.nil?
     Rails.logger.debug "Finding auto discovery rule for host #{host.name} (#{host.id})"
     # for each discovery rule ordered by priority
     DiscoveryRule.where(:enabled => true).order(:priority).each do |rule|
@@ -29,17 +30,17 @@ module Foreman::Controller::DiscoveredExtensions
       end
     end
     return false
-    end
+  end
 
-    def validate_rule_by_taxonomy rule, host
-      if SETTINGS[:organizations_enabled]
-        return false unless rule.organizations.include?(host.organization)
-      end
-      if SETTINGS[:locations_enabled]
-        return false unless rule.locations.include?(host.location)
-      end
-      true
+  def validate_rule_by_taxonomy rule, host
+    if SETTINGS[:organizations_enabled]
+      return false unless rule.organizations.include?(host.organization)
     end
+    if SETTINGS[:locations_enabled]
+      return false unless rule.locations.include?(host.location)
+    end
+    true
+  end
 
   # trigger the provisioning
   def perform_auto_provision original_host, rule
--- a/app/controllers/discovery_rules_controller.rb
+++ b/app/controllers/discovery_rules_controller.rb
@@ -53,10 +53,8 @@ class DiscoveryRulesController < ApplicationController
 
   def action_permission
     case params[:action]
-    when 'enable'
-      :enable
-    when 'disable'
-      :disable
+    when 'enable', 'disable'
+      :edit
     else
       super
     end
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -24,16 +24,15 @@ module DiscoveredHostsHelper
   end
 
   def multiple_discovered_hosts_actions_select
-    actions = [[_('Delete hosts'), multiple_destroy_discovered_hosts_path]]
-    actions <<  [_('Assign Organization'), select_multiple_organization_discovered_hosts_path] if SETTINGS[:organizations_enabled]
-    actions <<  [_('Assign Location'), select_multiple_location_discovered_hosts_path] if SETTINGS[:locations_enabled]
+    actions = [[_('Delete hosts'), multiple_destroy_discovered_hosts_path, hash_for_multiple_destroy_discovered_hosts_path]]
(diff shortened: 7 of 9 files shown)


app/controllers/discovered_hosts_controller.rb
```diff
@@ -202,16 +202,12 @@ def controller_permission
 
   def action_permission
     case params[:action]
-      when 'refresh_facts', 'reboot', 'reboot_all'
-        :view
-      when 'update_multiple_location', 'select_multiple_organization', 'update_multiple_organization', 'select_multiple_location'
+      when 'refresh_facts', 'reboot', 'reboot_all', 'update_multiple_location', 'select_multiple_organization', 'update_multiple_organization', 'select_multiple_location'
```
orrabin, line 205: shouldn't `select_multiple_organization` be under view permissions?

db/migrate/20150714144500_review_discovery_permissions.rb
```diff
@@ -0,0 +1,12 @@
+class ReviewDiscoveryPermissions < ActiveRecord::Migration
+  def up
+    if mgr = Role.find_by_name("Discovery Manager")
+      mgr.add_permissions!(["submit_discovered_hosts", "auto_provision_discovered_hosts", "create_discovery_rules"])
```
orrabin: maybe only add these if the `provision_discovered_hosts` permission exists since these three replace it and we wouldn't want to readd permissions if someone deliberately removed them

lzap:
@ares will you have a min to take a look on my permissions review?

app/controllers/discovered_hosts_controller.rb
lzap, line 205: Since all the actions reachable via multiple selections are edit-like, I'd leave it as is. The same case is Foreman Core Hosts controller.

lib/foreman_discovery/engine.rb
```diff
@@ -45,16 +45,23 @@ class Engine < ::Rails::Engine
             :discovered_hosts          => [:index, :show, :auto_complete_search],
             :"api/v2/discovered_hosts" => [:index, :show]
           }, :resource_type => 'Host'
-          permission :provision_discovered_hosts, {
+          permission :submit_discovered_hosts, {
+            :"api/v2/discovered_hosts" => [:facts, :create]
+          }, :resource_type => 'Host'
+          permission :auto_provision_discovered_hosts, {
             :discovered_hosts          => [:auto_provision, :auto_provision_all],
-            :"api/v2/discovered_hosts" => [:create, :auto_provision, :auto_provision_all]
```
lzap, line 50: FYI `create` action was removed by https://github.com/theforeman/foreman_discovery/commit/1f30d55ff87cb2a04b421a2b119275d40e268411 but we forgot to remove the permissions here

lzap:
For easier review, this is the code state before this patch:

```
        # discovered hosts permissions
        security_block :discovery do
          permission :view_discovered_hosts, {
            :discovered_hosts          => [:index, :show, :auto_complete_search],
            :"api/v2/discovered_hosts" => [:index, :show]
          }, :resource_type => 'Host'
          permission :provision_discovered_hosts, {
            :discovered_hosts          => [:auto_provision, :auto_provision_all],
            :"api/v2/discovered_hosts" => [:create, :auto_provision, :auto_provision_all]
          }, :resource_type => 'Host'
          permission :edit_discovered_hosts, {
            :discovered_hosts          => [:edit, :update, :update_multiple_location,
                                           :select_multiple_organization,
                                           :update_multiple_organization,
                                           :select_multiple_location, :refresh_facts, :reboot, :reboot_all],
            :"api/v2/discovered_hosts" => [:update, :facts, :refresh_facts, :reboot, :reboot_all]
          }, :resource_type => 'Host'
          permission :destroy_discovered_hosts, {
            :discovered_hosts          => [:destroy, :submit_multiple_destroy, :multiple_destroy],
            :"api/v2/discovered_hosts" => [:destroy]
          }, :resource_type => 'Host'
        end

        # discovery rules permissions
        security_block :discovery_rules do
          permission :view_discovery_rules, {
            :discovery_rules => [:index, :show, :auto_complete_search],
            :"api/v2/discovery_rules" => [:index, :show]
          }, :resource_type => 'DiscoveryRule'
          permission :new_discovery_rules, {
            :discovery_rules => [:new, :create],
            :"api/v2/discovery_rules" => [:create]
          }, :resource_type => 'DiscoveryRule'
          permission :edit_discovery_rules, {
            :discovery_rules => [:edit, :update, :enable, :disable],
            :"api/v2/discovery_rules" => [:create, :update]
          }, :resource_type => 'DiscoveryRule'
          permission :execute_discovery_rules, {
            :discovery_rules => [:auto_provision, :auto_provision_all],
            :"api/v2/discovery_rules" => [:auto_provision, :auto_provision_all]
          }, :resource_type => 'DiscoveryRule'
          permission :delete_discovery_rules, {
            :discovery_rules => [:destroy],
            :"api/v2/discovery_rules" => [:destroy]
          }, :resource_type => 'DiscoveryRule'
        end

        # roles (only added if they don't exist)
        role "Discovery Manager", [:view_discovered_hosts, :provision_discovered_hosts, :edit_discovered_hosts, :destroy_discovered_hosts, :view_discovery_rules, :new_discovery_rules, :edit_discovery_rules, :execute_discovery_rules, :delete_discovery_rules]
        role "Discovery Reader", [:view_discovered_hosts, :view_discovery_rules]
```

db/migrate/20150714144500_review_discovery_permissions.rb
lzap: Good point, this is what I am going to do:
- if `edit_discovered_hosts` is present add `submit_discovered_hosts`
- if `provision_discovered_hosts` is present add `auto_provision_discovered_hosts`
- if `new_discovery_rules` is present add `create_discovery_rules`

Objections?

lzap:
> 1) I can see the discovery rules but I can't edit so how do I see the hostname template? we don't  +have a readonly view of a rule, just the index for all rules.

Since we show almost all items in the index, I think this is fine. I 
wonder how this is done in core tho.

> 2) I can see the hostgroup name but just (I think this is ok but just checking).

Yeah, the same behavior as in (1). Core does the same. Let's keep that.

> 3) I can see the options to provision/reboot but when I click on them I get a permissions error -  +should they be hidden?

Good point, fixed.

> 4) When I try bulk actions I don't get permission denied but I can't actually perform the action   +so I get an empty screen with a spinner.

Fixed.

> Discover manager through the UI:
> 1) I can't create a discovery rule (I see hostgroups but can not save a rule with them)

Yeah, getting "Failed to save: Organization ids Invalid organizations
selection, you must select at least one of yours, Location ids Invalid
locations selection, you must select at least one of yours". When I add
`view_org/loc` permissions, I can see the tabs added but I see no 
loc/org.

I don't know yet how to allow org/loc to be shown on the New rule page.
Maybe a bug in our code, because the user apparently can see orgs/locs 
in the Org/Log selector in the top-left corner.

> 2) I can click provision, fill in a hg and see it's os and everything although I can't save

With the fixes above, I was able to provision just fine, except I get 
the Permission error in the end because there is no view_hosts 
permission. We cant give it it is too broad.

lzap:
Amending the role DSL change plus:

``` diff
diff --git a/app/helpers/discovered_hosts_helper.rb b/app/helpers/discovered_hosts_helper.rb
index b7369f9..82d0cb3 100644
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -24,16 +24,15 @@ module DiscoveredHostsHelper
   end

   def multiple_discovered_hosts_actions_select
-    actions = [[_('Delete hosts'), multiple_destroy_discovered_hosts_path]]
-    actions <<  [_('Assign Organization'), select_multiple_organization_discovered_hosts_path] if SETTINGS[:organizations_enabled]
-    actions <<  [_('Assign Location'), select_multiple_location_discovered_hosts_path] if SETTINGS[:locations_enabled]
+    actions = [[_('Delete hosts'), multiple_destroy_discovered_hosts_path, hash_for_multiple_destroy_discovered_hosts_path]]
+    actions <<  [_('Assign Organization'), select_multiple_organization_discovered_hosts_path,  hash_for_select_multiple_organization_discovered_hosts_path] if SETTINGS[:organizations_enabled]
+    actions <<  [_('Assign Location'), select_multiple_location_discovered_hosts_path,  hash_for_select_multiple_location_discovered_hosts_path] if SETTINGS[:locations_enabled]

     select_action_button( _("Select Action"), {:id => 'submit_multiple'},
       actions.map do |action|
-        link_to_function(action[0], "build_modal(this, '#{action[1]}')", :'data-dialog-title' => _("%s - The following hosts are about to be changed") % action[0])
+        link_to_function(action[0], "build_modal(this, '#{action[1]}')", :'data-dialog-title' => _("%s - The following hosts are about to be changed") % action[0]) if authorized_for(action[2])
       end.flatten
     )
-
   end

   def turn_zero_to_not_available(value)
diff --git a/app/views/discovered_hosts/_discovered_hosts_list.html.erb b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
index 1d9b738..37e60e4 100644
--- a/app/views/discovered_hosts/_discovered_hosts_list.html.erb
+++ b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
@@ -39,10 +39,10 @@
       <td class="hidden-tablet hidden-xs"><%= disc_report_column(host) %></td>
       <td>
         <%= action_buttons(
-          link_to(_("Provision"), hash_for_edit_discovered_host_path(:id => host)),
-          link_to(_("Auto Provision"), hash_for_auto_provision_discovered_host_path(:id => host), :method => :post),
-          link_to(_("Refresh facts"), hash_for_refresh_facts_discovered_host_path(:id => host)),
-          link_to(_("Reboot"), hash_for_reboot_discovered_host_path(:id => host), :method => :put),
+          display_link_if_authorized(_("Provision"), hash_for_edit_discovered_host_path(:id => host)),
+          display_link_if_authorized(_("Auto Provision"), hash_for_auto_provision_discovered_host_path(:id => host), :method => :post),
+          display_link_if_authorized(_("Refresh facts"), hash_for_refresh_facts_discovered_host_path(:id => host)),
+          display_link_if_authorized(_("Reboot"), hash_for_reboot_discovered_host_path(:id => host), :method => :put),
           display_delete_if_authorized(hash_for_discovered_host_path(:id => host), :confirm => _("Delete %s?") % host.name, :action => :destroy))%>
         </td>
       </tr>
```

Make sure your roles has expected permissions after you update.

We still have one open item - cannot create new rule.

db/migrate/20150714144500_review_discovery_permissions.rb
```diff
@@ -0,0 +1,16 @@
+class ReviewDiscoveryPermissions < ActiveRecord::Migration
+  def up
+    if mgr = Role.find_by_name("Discovery Manager")
```
ares: just a nit, it's a nice convention to surround assignment by brackets if it's part of `if` so everyone knows that there's not missing second equals sign

`if (mgr = Role.find_by_name("Discovery Manager"))`

db/migrate/20150714144500_review_discovery_permissions.rb
```diff
@@ -0,0 +1,16 @@
...
+      perms << "submit_discovered_hosts" if Permission.find_by_name("edit_discovered_hosts")
+      perms << "auto_provision_discovered_hosts" if Permission.find_by_name("provision_discovered_hosts")
+      perms << "create_discovery_rules" if Permission.find_by_name("new_discovery_rules")
+      mgr.add_permissions!(perms)
+    end
+    Permission.find_by_name("new_discovery_rules").try(:destroy)
+  end
+
+  def down
+    # Foreman plugin initializer will re-create
```
ares: I don't think this is true, also what would destroy newly created permissions?

lib/foreman_discovery/engine.rb
```diff
@@ -45,16 +45,23 @@ class Engine < ::Rails::Engine
...
+            :"api/v2/discovered_hosts" => [:auto_provision, :auto_provision_all]
+          }, :resource_type => 'Host'
```
ares, line 54: hm, just double check that it generates queries as expected, `Host` is proxy for `Host::Managed` but I suppose you need `Host::Discovered`. Chances are it works correctly because of some if since the same resource_type is use elsewhere but looks suspicious.

ares:
> @ares will you have a min to take a look on my permissions review?

what you described makes sense, except few nits I think it should work as you expect

db/migrate/20150714144500_review_discovery_permissions.rb
lzap: Yeah bad comment, the fact is - we don't need to rollback anything in this case. Changed the comment.

lib/foreman_discovery/engine.rb
lzap, line 54: Yeah this works correctly, actually when I use `Host::Discovered` it does not work. Dunno why.

lzap:
Done with @ares suggestions, thanks.

@orrabin I have added `assign_org/loc` permission to the Manager role. Since assignement currently do not work due to bug http://projects.theforeman.org/issues/11187 I think we can merge if this is re-tested.

In regard to unit tests, we don't test permissions for all controllers and roles in core. I tried to write something like this as a test helper:

```
def set_session_user_with_perms perms
  perms = perms.collect{|p| FactoryGirl.create(:permission, :name => p) rescue Permission.find_by_name(p) }
  role = FactoryGirl.create :role
  perms.each do |perm|
    FactoryGirl.create(:filter, :role => role, :permissions => [perm])
  end
  user = FactoryGirl.create :user, :with_mail, :admin => false
  user.roles << role
  SETTINGS[:login] ? {:user => user, :expires_at => 5.minutes.from_now} : {}
end
```

This loads the seed script where I put this:

```
class DiscoveryPermissions
  READER = [
    :view_discovered_hosts,
    :view_discovery_rules,
    :view_locations,
    :view_organizations,
  ]

  MANAGER = READER + [
    :assign_organizations,
    :assign_locations,
    # discovered_hosts
    :submit_discovered_hosts,
    :auto_provision_discovered_hosts,
    :provision_discovered_hosts,
    :edit_discovered_hosts,
    :destroy_discovered_hosts,
    # discovered_rules
    :create_discovery_rules,
    :edit_discovery_rules,
    :execute_discovery_rules,
    :delete_discovery_rules,
  ]
end

unless Rails.env.test?
  Role.without_auditing do
    create_role('Discovery Reader', DiscoveryPermissions::READER, 0)
    create_role('Discovery Manager', DiscoveryPermissions::MANAGER, 0)
  end
end
```

Unfortunately filters are not correct. I would need to reconstruct whole security blocks and permissions with filters, but since we cannot reuse the code we have in engine.rb and seed script, I doubt this is helpful testing. If we change it and not the tests then we are not testing real state. Any idea?

lzap:
Ok I made it working! I was missing `.save!` statements for all FactoryGirls! Now we have tests for rules, and it actually found a bug. Renamed `delete_discovery_rules` to `destroy_discovery_rules`.

lzap:
And added tests for discovery hosts! RTT.

lzap:
All green! :+1:

db/seeds.d/20-discovery-roles.rb
```diff
@@ -0,0 +1,31 @@
...
+    :destroy_discovery_rules,
+  ]
+end
+
+unless Rails.env.test?
+  Role.without_auditing do
+    create_role('Discovery Reader', DiscoveryPermissions::READER, 0)
+    create_role('Discovery Manager', DiscoveryPermissions::MANAGER, 0)
+  end
+end
```
lzap: @domcleal does this look reasonable as the standard for other plugins?

This deprecates role DSL and solves two issues:

1) Chicken-n-egg problem for permissions from core you want to assign to plugin roles
2) Possibility to test permissions in functional tests (we load this file in a test helper)
3) Ability to load this the same way from Foreman Core to add Role validation feature later on

domcleal: I would guess that (1) and (2) could probably be fixed in Foreman, and I don't know what (3) is.

If you want to deprecate it, then first let's see if the issues can be fixed to avoid disruption, and if not, then you need to be filing a PR against Foreman itself to mark the feature as deprecated.

lzap: One possible solution for core comes to my mind: to ignore role statements when we are doing migrations and seeds. So plugin roles are created _after_ database is migrated and seeded with core permissions. Ignore (3) for now, let me try a core patch.

lzap:
Ok I am dropping the seed file for now and trying once again the role DSL.

```
class DiscoveryPermissions
  READER = [
    :view_discovered_hosts,
    :view_discovery_rules,
    :view_locations,
    :view_organizations,
  ]

  MANAGER = READER + [
    :assign_organizations,
    :assign_locations,
    # discovered_hosts
    :submit_discovered_hosts,
    :auto_provision_discovered_hosts,
    :provision_discovered_hosts,
    :edit_discovered_hosts,
    :destroy_discovered_hosts,
    # discovered_rules
    :create_discovery_rules,
    :edit_discovery_rules,
    :execute_discovery_rules,
    :destroy_discovery_rules,
  ]
end

unless Rails.env.test?
  Role.without_auditing do
    create_role('Discovery Reader', DiscoveryPermissions::READER, 0)
    create_role('Discovery Manager', DiscoveryPermissions::MANAGER, 0)
  end
end
```

lzap:
Ok pushed again with role DSL, I cannot reproduce the chicken and egg problem locally. Waiting for Jenkins.

lzap:
@orrabin FYI I am trying to refactor tests with this helper patch: https://github.com/theforeman/foreman/pull/2572 to be able to do the same tests but without the seeds hack. Unfortunately tests on Jenkins fails and I am unable to reproduce this locally. I don't know how to workaround this checken and egg problem.

lzap:
[test] with `db:seed`

lzap:
Ok this fails, I am going to push this without unit tests so we can merge for 4.0 and downstream. Unit test will follow in a separate PR. @orrabin please test and merge

orrabin:
@lzap I tested - with discovery reader everything looks good but with manager I had a few problems:
- I get no route to host when trying to reboot or refresh facts
- auto_provision returned the error: "Unable to find a discovery rule, no host provided" but my host was provisioned
  I checked with an admin user to make sure it wasn't my setup and reboot/refresh facts worked fine

lzap:
@orrabin can you provide a screenshot of Manager role with all its filters?

lzap:
Also can you make sure it does really not work? Because I see a different bug in current upstream devel/devel branches: Refresh Facts and Reboot commands does work but I see weird long delays and AJAX call tends to timeout therefore success/error bubble does not show up (but behind the wheels it does what you request).

orrabin:
image

orrabin:
@lzap I didn't see the vm rebooting but I'll try again

lzap:
@orrabin found a reason - a bug on the image side, because when I do:

```
[lzap@lzapx ~]$ curl -3 -H "Accept:application/json" -H "Content-Type: application/json" -d '{}' -k -X PUT http://192.168.122.68:8443/bmc/ignored/chassis/power/cycle
```

It takes 10 seconds until it succeeds:

```
{"action":"cycle","result":true}
```

lzap:
Oh I found a reason. In my case I had multiple interfaces connected/routed to the same network. MAC filtering issues. Fixed with a dummy network:

```
sudo virt-install --os-variant rhel7 --vcpus 1 --ram 1900 \
    --name $NAME --nodisk --boot network \
    -w network=default,model=virtio,mac=52:54:00:13:05:13 \
    -w network=dummy,model=rtl8139 \
    -w network=dummy,model=e1000 \
    -w network=dummy,model=e1000 \
    --graphics spice,listen=0.0.0.0 --noautoconsole &
```

lzap:
Btw I doublechecked your role, it's fine. Should work really.

orrabin:
@lzap I retested against latest develop and I don't get the errors anymore
Edit: the only problem I had now was selecting the hosts for bulk actions, I could only select all and not one by one

lzap:
@orrabin fyi we are not backporting this in 3.0 and not even 4.0 if the upstream patch is not merged into 1.9 which is unlikely to happen (in that case we will branch 4.0 off from parent's commit)

lzap:
Ok this should now pass [test]

lzap:
@orrabin RTM!

lzap:
Fixed the multiselect issue:

``` diff
diff --git a/app/helpers/discovered_hosts_helper.rb b/app/helpers/discovered_hosts_helper.rb
index 82d0cb3..50cf8ac 100644
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -43,4 +43,8 @@ module DiscoveredHostsHelper
   end

+  def authorized_for_edit_destroy?
+    authorized_for(:controller => :discovered_hosts, :action => :edit) or
+        authorized_for(:controller => :discovered_hosts, :action => :destroy)
+  end
 end
diff --git a/app/views/discovered_hosts/_discovered_hosts_list.html.erb b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
index 37e60e4..5c0e171 100644
--- a/app/views/discovered_hosts/_discovered_hosts_list.html.erb
+++ b/app/views/discovered_hosts/_discovered_hosts_list.html.erb
@@ -26,7 +26,7 @@
   <% @hosts.each do |host| -%>
     <tr>
       <td class="ca">
-        <%= check_box_tag "host_ids[]", nil, false, :id => "host_ids_#{host.id}", :disabled => !authorized?, :class => 'host_select_boxes', :onclick => 'hostChecked(this)' -%>
+        <%= check_box_tag "host_ids[]", nil, false, :id => "host_ids_#{host.id}", :disabled => !authorized_for_edit_destroy?, :class => 'host_select_boxes', :onclick => 'hostChecked(this)' -%>
       </td>
       <%= render :partial => "discovered_host", :locals => {:host => host} %>
       <% if SETTINGS[:locations_enabled] -%>
```

orrabin:
@lzap the multiselect now works for me, I'll merge when tests are green

orrabin:
merged as a6826bee71d7818959aca7dbbba035fb94a8c204, thanks @lzap!

PR #199, opened 2015-07-16
cherry picks for 3.0.1

orrabin:
@lzap can you please go over the list to see if something is missing or you don't want part of it in 3.0.1?


lzap:
Can we remove `#10555, #10554`?

lzap:
Ok tomorrow I will give this a try.

lzap:
Maybe we can squeeze https://github.com/theforeman/foreman_discovery/pull/197 and/or https://github.com/theforeman/foreman_discovery/pull/188 in once merged :-) Then I can perform testing of this PR. What you think?

lzap:
Oh let's keep #197 out of this (change in roles needed manually), so just #188.

lzap:
Ok let's keep Ohads commits for 3.x series.

app/models/host/discovered.rb
```diff
@@ -160,18 +146,14 @@ def reboot
...
-
-    msg = status ? 'successful' : 'failed'
-    logger.info "ForemanDiscovery: reboot result: #{msg}"
-    status
-    rescue => e
-      logger.info "ForemanDiscovery: reboot result: failed"
-      logger.warn e.backtrace.join('\n')
-      raise e
+  rescue => e
+    ::Foreman::Logging.exception("Unable to reboot #{name}", e)
```
lzap: This will unlikely work as there is no such thing in 1.8.

orrabin: so I'll revert this to the previous logger.warn?

lzap: > so I'll revert this to the previous logger.warn?

Yes I think this is the only place. Doublecheck please.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Ok since there are no other patches for 3.x I am going to test this and cherry pick.

lzap:
It works fine, I will be merging this tomorrow and doing the release.

lzap:
Pushed into 3.0-stable

lzap:
Ok I forced pushed 3.0.1 into dummy release (copy of 3.0.0). It turned out, unfortunately _after_ I pushed, that this is not compatible with facter extensions. Let's review this after 4.0 is out.

PR #200, opened 2015-07-21
Fixes 11008 - Changed the view for discovered hosts show page.

alongoldboim:
I have also hidden the rest of the facts and added a button to expand the facts table.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -2,9 +2,11 @@ class DiscoveredHostsController < ::ApplicationController
   include Foreman::Controller::AutoCompleteSearch
   include Foreman::Controller::TaxonomyMultiple
   include Foreman::Controller::DiscoveredExtensions
+  include ActionView::Helpers::NumberHelper
   unloadable
 
-  before_filter :find_by_name, :only => %w[show edit update destroy refresh_facts convert reboot auto_provision]
+  before_filter :find_by_name, :only => %w[edit update destroy refresh_facts convert reboot auto_provision]
+  before_filter :find_by_name_incl_subnet, :only => [:show]
   before_filter :find_multiple, :only => [:multiple_destroy, :submit_multiple_destroy]
   before_filter :taxonomy_scope, :only => [:edit]
 
@@ -26,9 +28,16 @@ class DiscoveredHostsController < ::ApplicationController
   def show
     # filter graph time range
     @range = nil
-
     # summary report text
     @report_summary = nil
+    init_regex_and_categories
+    @interfaces = []
+    get_interfaces
+    @host.facts_hash.each do |key, value|
+      value = number_to_human_size(value) if /size$/.match(key)
+      assign_fact_to_category(key, value) unless @interfaces.any? {|interface| key.include? interface[:identifier]}
+    end
+    add_custom_facts
   end
 
   def destroy
@@ -171,13 +180,35 @@ class DiscoveredHostsController < ::ApplicationController
 
   private
 
+  def init_regex_and_categories
+    hightlights = Setting[:discovery_facts_highlights].empty? ? /^(productname|memorysize|manufacturer|architecture|macaddress$|processorcount|physicalprocessorcount|discovery_subnet|discovery_boot|ipaddress$)/ : Regexp.new(eval(Setting[:discovery_facts_highlights]))
+    storage = Setting[:discovery_facts_storage].empty? ? /^blockdevice/ : Regexp.new(eval(Setting[:discovery_facts_storage]))
+    hardware = Setting[:discovery_facts_hardware].empty? ? /^(hardw|manufacturer|memo|process)/ : Regexp.new(eval(Setting[:discovery_facts__hardware]))
+    network = Setting[:discovery_facts_network].empty? ? /^(ipaddress|interfaces|dhcp|fqdn|hostname|link|mtu|net|macaddress|wol|port|speed)/ : Regexp.new(eval(Setting[:discovery_facts_network]))
+    software = Setting[:discovery_facts_software].empty? ? /^(bios|os|discovery)/ : Regexp.new(eval(Setting[:discovery_facts_software]))
+    ipmi = Setting[:discovery_facts_ipmi].empty? ? /^ipmi/ : Regexp.new(eval(Setting[:discovery_facts_ipmi]))
+    @regex_array = [hightlights, storage, hardware, network, software, ipmi, false]
+    @categories = Array.new(7) { Hash.new }
+    @categories_names = [:Hightlights, :Storage, :Hardware, :Network, :Software, :IPMI, :Misceleneous]
+  end
+
+  def assign_fact_to_category(key, value )
+    @regex_array.each_with_index do |regex, index|
+      if !regex
+        @categories[index][key] = value
+      elsif regex.match key
+        @categories[index][key] = value
+        break
+      end
+    end
+  end
+
   def resource_base
     @resource_base ||= ::Host::Discovered.authorized(current_permission, ::Host::Discovered)
   end
 
   def load_vars_for_ajax
     return unless @host
-
     @environment     = @host.environment
     @architecture    = @host.architecture
     @domain          = @host.domain
@@ -213,11 +244,15 @@ class DiscoveredHostsController < ::ApplicationController
     end
   end
 
-  def find_by_name
+  def find_by_name(*includes)
     params[:id].downcase! if params[:id].present?
-    @host = resource_base.find_by_id(params[:id])
-    @host ||= resource_base.find_by_name(params[:id])
+    @host = includes.empty? ? resource_base.find_by_id(params[:id]) : resource_base.includes(includes).find_by_id(params[:id])
+    @host ||= includes.empty? ? resource_base.find_by_name(params[:id]) : resource_base.includes(includes).find_by_name(params[:id])
     return false unless @host
+  end
+
+  def find_by_name_incl_subnet
+    find_by_name({:interfaces => :subnet})
   end
 
   def find_multiple
@@ -261,5 +296,18 @@ class DiscoveredHostsController < ::ApplicationController
     yield
   ensure
     Bullet.enable = true if defined? Bullet
+  end
+
+  def get_interfaces
+    @host.interfaces.each do |interface|
+      @interfaces << {:identifier => interface["identifier"], :type => interface["type"], :mac => interface["mac"], :ip => interface["ip"]? interface["ip"] : "N/A", :primary => interface["primary"], :provision => interface["provision"]}
+    end
+  end
+
+  def add_custom_facts
+    unless @host.primary_interface.subnet.nil?
+      discovery_subnet = "#{@host.primary_interface.subnet.name} (#{@host.primary_interface.subnet.network})"
+      assign_fact_to_category("discovery_subnet", discovery_subnet)
+    end
   end
 end
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -1,5 +1,12 @@
 module DiscoveredHostsHelper
 
+  def attach_flags(interface)
+    flags = ""
+    flags += "flag-primary " if interface[:primary]
+    flags += "flag-provision" if interface[:provision]
+    flags
+  end
+
   def disc_report_column(record)
     record.last_report? ? (_("%s ago") % time_ago_in_words(record.last_report.getlocal)) : ""
   end
@@ -10,12 +17,20 @@ module DiscoveredHostsHelper
     actions <<  [_('Refresh facts') ,hash_for_refresh_facts_discovered_host_path(:id => host)]
     actions <<  [_('Reboot') ,hash_for_reboot_discovered_host_path(:id => host), :method => :put]
     title_actions(
+        button_group(
+            link_to(_("Back"), :back)
+        ),
         select_action_button( _("Select Action"), {},
                               actions.map do |action|
                                 method = action[2] if action.size > 1
                                 link_to(action[0] , action[1], method)
                               end.flatten
         ),
+      button_group(
+        link_to(_("Expand All"),"#",:class => "btn ",:onclick => "$('.glyphicon-plus-sign').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign');
+                $('.facts-panel').addClass('collapse in').height('auto');"
+        )
+      ),
       button_group(
         link_to(_("Delete"), hash_for_discovered_host_path(:id => host),
                 :class => "btn btn-danger", :confirm => _('Are you sure?'), :method => :delete)
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -2,6 +2,12 @@ class Setting::Discovered < ::Setting
   BLANK_ATTRS << "discovery_location"
   BLANK_ATTRS << "discovery_organization"
   BLANK_ATTRS << "discovery_fact_column"
+  BLANK_ATTRS << 'discovery_facts_highlights'
+  BLANK_ATTRS << 'discovery_facts_storage'
+  BLANK_ATTRS << 'discovery_facts_software'
+  BLANK_ATTRS << 'discovery_facts_hardware'
+  BLANK_ATTRS << 'discovery_facts_network'
+  BLANK_ATTRS << 'discovery_facts_ipmi'
 
   def self.load_defaults
     # Check the table exists
@@ -42,6 +48,16 @@ class Setting::Discovered < ::Setting
       end
     end
 
+    Setting.transaction do
+      [
+          self.set('discovery_facts_highlights', N_("Regex to organize facts for highlights section"), ""),
+          self.set('discovery_facts_storage', N_("Regex to organize facts for storage section"), ""),
+          self.set('discovery_facts_software', N_("Regex to organize facts for software section"), ""),
+          self.set('discovery_facts_hardware', N_("Regex to organize facts for hardware section"), ""),
+          self.set('discovery_facts_network', N_("Regex to organize facts for network section"), ""),
+          self.set('discovery_facts_ipmi', N_("Regex to organize facts for ipmi section"), ""),
+      ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
+    end
     true
 
   end
--- a/app/views/discovered_hosts/show.html.erb
+++ b/app/views/discovered_hosts/show.html.erb
@@ -2,15 +2,94 @@
 
 <%= discovered_hosts_title_actions(@host) %>
 
-<h3><%= _('Facts discovered on this host') -%></h3>
-<table class="table table-bordered table-striped table-condensed">
-  <tr>
-    <th><%= _('Fact') -%></th>
-    <th><%= _('Value') -%></th>
-  <% @host.facts_hash.sort.each do |fact| -%>
-    <tr>
-      <td><%= fact[0] %></td>
-      <td><%= truncate(fact[1], :length => 100) %></td>
-    </tr>
-  <% end -%>
-</table>
+<div class="row">
+  <div class="col-md-6">
+    <div class="panel panel-default">
+      <div class="panel-heading" ><strong> <%= @categories_names[0] %></strong> </div>
+        <table class="table table-bordered table-condensed table-fixed">
+        <% @categories[0].sort.each do |keys, val| %>
+            <tr class="">
+            <td class="ellipsis" width="40%"> <strong> <%= keys %> </strong></td>  <td><%= val %></td>
+            </tr>
+        <% end -%>
+        </table>
+    </div>
+    <% unless @interfaces.empty? %>
+      <div class="panel panel-default">
+        <div class="panel-heading" role="tab">
+          <h4 class="panel-title">
+            <a role="button" class="expendable-link" data-toggle="collapse" data-parent="#accordion" href="#interfaces-panel" aria-expanded="true" onclick="$(this).find(':first-child').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign')">
+              <span class="glyphicon glyphicon-minus-sign" aria-hidden="true"></span>
+              Interfaces
+            </a>
+          </h4>
+        </div>
+        <div id="interfaces-panel" class="panel-collapse collapse facts-panel in" >
+          <table class="table table-bordered table-condensed table-fixed" id="interfaceList">
+            <tr>
+              <th class="hidden-xs " width="12%"></th>
+              <th class="hidden-xs "><%= _('Identifier') %></th>
+              <th class="hidden-xs "><%= _("MAC address") %></th>
+              <th class="hidden-xs "><%= _("IP address") %></th>
+            </tr>
+            <% @interfaces.each do |interface| %>
+              <tr>
+                <td class="<%= attach_flags(interface) %> ellipsis"></td>
+                <td class="ellipsis"><%= interface[:identifier] %></td>
+                <td class="ellipsis"><%= interface[:mac] %></td>
+                <td class="ellipsis"><%= interface[:ip] %></td>
+              </tr>
+            <% end %>
+          </table>
+        </div>
+      </div>
+    <% end %>
+  </div>
+
+<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+    <div class="col-md-6">
+      <% @categories.each_with_index do |val, index| %>
+        <% next if index == 0 || val.empty? %>
+        <div class="panel panel-default">
+          <div class="panel-heading" role="tab">
+            <h4 class="panel-title">
+              <a role="button" class="expendable-link" data-toggle="collapse" data-parent="#accordion" href="#<%= @categories_names[index].to_s + "panel" %>" aria-expanded="true" onclick="$(this).find(':first-child').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign')">
+                <span class="glyphicon glyphicon<%= (@categories_names[index].to_s.include?("Storage")? "-minus-sign":"-plus-sign")%>" aria-hidden="true"></span>
+                <%= @categories_names[index] %>
+              </a>
+            </h4>
+          </div>
+          <div id="<%= @categories_names[index].to_s + "panel" %>" class="panel-collapse collapse facts-panel <%= "in" if @categories_names[index].to_s.include?"Storage"%>" role="tabpanel">
+            <table class="table table-bordered  table-condensed table-fixed">
+            <% val.sort.each do |key, value|  %>
+                <tr class="">
+                  <td width="40%" class="ellipsis"><strong> <%= key %> </strong> </td> <td class="ellipsis"> <%= value %> </td>
+                </tr>
+            <% end %>
+            </table>
+          </div>
+        </div>
+      <% end -%>
+    </div>
+  </div>
+</div>
+
+<script type="text/javascript" charset="utf-8">
+  $(document).ready(function(){
+    $('table tr:nth-child(even)').css('background-color','#f5f5f5');
+    $("table").css('padding', '0');
+    $("table").css("cssText", 'margin-bottom: 0 !important;');
+    $(".panel-default").css('margin-bottom', "10px")
+
+    provision_class = 'active'
+    primary_class = 'active'
+    var flag_primary = '<i class="glyphicon glyphicon glyphicon-tag primary-flag '+ primary_class +'" title="" data-original-title="'+ __('Primary') +'"></i>';
+    var flags_provision = '<i class="glyphicon glyphicon glyphicon-hdd provision-flag '+ provision_class +'" title="" data-original-title="'+ __('Provisioning') +'"></i>';
+
+    $('.flag-primary').html(flag_primary);
+    $('.flag-provision').append(flags_provision);
+    $('.primary-flag').tooltip();
+    $('.provision-flag').tooltip();
+  });
+</script>
+


alongoldboim:
@ohadlevy I'm not sure i have added all the facts you've wanted. need to check if i understood correctly.

ohadlevy:
@alongoldboim screenshot always help :)

app/assets/javascripts/discovered_hosts.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function(){
+
+$('#moreFactsBtn').click(function () {
+    $(this).text(function(i, text){
```
ohadlevy: indentation.. should be 2 spaces

app/assets/javascripts/discovered_hosts.js
```diff
@@ -0,0 +1,9 @@
...
+        return text === "More facts" ? "Less facts" : "More facts";
```
ohadlevy: strings are not extract

app/helpers/discovered_hosts_helper.rb
```diff
@@ -44,4 +44,19 @@ def discovery_attribute(host, attr, default_value = _('N/A'))
     host.try(:discovery_attribute_set).try(attr) || default_value
   end
 
+  def facts_table(values,with_header)
+    content_tag :table, class: "table table-bordered table-striped table-conden" do
+      rows = []
+      rows << content_tag(:th, _('Fact')) if with_header
```
ohadlevy: a bit cryptic imho to write the whole table as ruby. if you are creating a table helper you could do something like

``` erb
<% table do %>
<thead>
<th>
..
</thead>
<tbody>
<tr>
...
</tr>
</tbody>
<% end %>
```

alongoldboim:
selection_001
selection_002

app/assets/javascripts/discovered_hosts.js
```diff
@@ -0,0 +1,9 @@
...
+    })
+});
+
+});
```
lzap: Missing newline

lib/foreman_discovery/engine.rb
```diff
@@ -19,6 +19,19 @@ class Engine < ::Rails::Engine
...
+      )
+    end
+
+    initializer 'foreman_discovery.configure_assets', :group => :assets do
+      SETTINGS[:foreman_discovery] =
+          {:assets => {:precompile => [
+              'discovered_hosts.js',
+          ]}}
+    end
+
```
lzap, line 28: Isn't this an overkill for such a tiny feature? I mean, this needs a change in our packaging. Can you make a change for both Red Hat and Debian packages? https://github.com/theforeman/foreman-packaging

lzap, line 28: I think this could be also inline JavaScript, it's just one statement.

lzap, line 28: Maybe instead of changing the button value, we can just hide it?

alongoldboim, line 28: @lzap i can do it inline so we won't have to do a repackaging.
About the feature as a whole i haven't finished yet, this was just the proto.

alongoldboim:
@lzap @ohadlevy does this look better? any advises?
selection_004
selection_005

lzap:
As a prototype, this looks nice indeed.

alongoldboim:
Uploaded this version, if you could play around with it and give me some pointers ill be more then happy.

ohadlevy:
@alongoldboim do you think the general info is the most interesting information to show by default?

what about facts that do not fit a category?

alongoldboim:
@ohadlevy its not the most interesting your right, i got caught up with the whole category splitting :), ill gather the good facts in the general, the rest of the facts are in "miscellaneous" i added it on the commit but its not in the ss.
As for the look of it would you recommend on doing it different?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,6 +25,9 @@ def index
 
   def show
     # filter graph time range
+    @general= { :Highlights => ["interface", "ipaddress", "virtual", "manufacturer", "architecture", "memorysize_mb", "os", "macaddress"]}
+    @highlights ={:Network=>["discovery", "dhcp", "fqdn", "hostname", "link", "mtu", "net"], :Storage => ["blockdevice"], :Hardware => ["hardw", "manufacturer", "memo", "process"], :Software => ["bios", "os" ], :Etc => [] }
+    @facts = @host.facts_hash
     @range = nil
```
lzap, line 30: Things I'd scratch from general pane:
- memorysize_mb: 1855.26
- virtual: physical
- is_virtual: false
- macaddress: 52:54:00:26:fd:82

Things I miss in the general pane:

processorcount: 1
physicalprocessorcount: 1
memorytotal: 1.81 GB (this is more readable)
discovery_bootif: 52:54:00:13:05:13
discovery_bootip: 192.168.122.68

and total storage size as well.

lzap:
It looks very nice, several points:
- sort the facts by name in all panes
- rename "Etc" to somethign like "Miscellaneous"
- move discovery_version/release to "Software" pane
- since you know all interfaces names ("interfaces" fact), move all "*_inteface_name" facts in Network pane as well
- implement "..." or "Expand" for longer facts than 100 characters (typically ssh keys, an inline javascript one-way only is fine)
- move "productname" to general pane as well
- the most logical order of highlights is: Hardware, Storage, Network, Software, Etc
- open with Hardware expanded by default to leverage full space on the right side

Great addition, I like it.

lzap:
@ohadlevy fyi the pictures above were only UI protypes, it looks different in the code now...

alongoldboim:
@lzap thanks for the comments, applying changes.

app/controllers/discovered_hosts_controller.rb
alongoldboim, line 30: i agree, but i still think ill leave the mac address though.

lzap, line 30: Oh that was a pastebin error, scratch the mac of course.

lzap:
Ok, I don't see new commit yet. FYI unit tests failed.

alongoldboim:
Yea i need to see whats failing.
Monday it will be ready.

2015-07-30 11:34 GMT+03:00 Lukáš Zapletal notifications@github.com:

alongoldboim:
@lzap  I have added what you requested, check it and let me know if anything else is needed, thanks.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,8 +25,15 @@ def index
 
   def show
     # filter graph time range
+    @general =    {:Highlights => ["size", "productname", "memorysize", "interface", "manufacturer", "architecture", "os", "macaddress", "processorcount", "physicalprocessorcount", "discovery_boot"]}
+    @highlights = {:Hardware => ["hardw", "manufacturer", "memo", "process"],
+                  :Storage => ["blockdevice"],
+                  :Network=>["interfaces", "dhcp", "fqdn", "hostname", "link", "mtu", "net"] + @host.facts_hash["interfaces"].split(","),
```
lzap: Missing line space: `=>`

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,46 @@
...
+            <h4 class="panel-title">
+              <a role="button" data-toggle="collapse" data-parent="#accordion" href="#<%=keys.to_s+"panel"%>" aria-expanded="true" onclick="$(this).find(':first-child').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign')">
+                <span class="glyphicon glyphicon<%= (keys.to_s.include?("Hardware")? "-minus-sign":"-plus-sign")%>" aria-hidden="true"></span>
+                <%=keys%>
+              </a>
+            </h4>
+          </div>
+          <div id="<%=keys.to_s+"panel"%>" class="panel-collapse collapse <%="in" if keys.to_s.include?"Hardware"%>" role="tabpanel">
+            <div class="panel-body">
+              <%@facts.sort_by { |k,v| k.downcase} %>
```
lzap: Missing space.

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,46 @@
...
+
+  <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+    <div class="col-md-6">
+      <% @highlights.each do |keys, val| %>
+           <div class="panel panel-default">
+          <div class="panel-heading" role="tab">
+            <h4 class="panel-title">
+              <a role="button" data-toggle="collapse" data-parent="#accordion" href="#<%=keys.to_s+"panel"%>" aria-expanded="true" onclick="$(this).find(':first-child').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign')">
+                <span class="glyphicon glyphicon<%= (keys.to_s.include?("Hardware")? "-minus-sign":"-plus-sign")%>" aria-hidden="true"></span>
+                <%=keys%>
```
lzap: Missing spaces.

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,46 @@
...
+              </a>
+            </h4>
+          </div>
+          <div id="<%=keys.to_s+"panel"%>" class="panel-collapse collapse <%="in" if keys.to_s.include?"Hardware"%>" role="tabpanel">
+            <div class="panel-body">
+              <%@facts.sort_by { |k,v| k.downcase} %>
+              <% @facts.sort.select{ |key, value|  keys.to_s.include?("Miscellaneous") || val.any?{|v| key.include? v  }}.each do |fact| %>
+                <li class="list-group-item"> <strong> <%=fact[0] %>: </strong>  <%= truncate(breaking_wrap_wrap(fact[1],50),:length => 80)%></li>
```
lzap: Ditto

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,46 @@
...
+           <div class="panel panel-default">
+          <div class="panel-heading" role="tab">
+            <h4 class="panel-title">
+              <a role="button" data-toggle="collapse" data-parent="#accordion" href="#<%=keys.to_s+"panel"%>" aria-expanded="true" onclick="$(this).find(':first-child').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign')">
+                <span class="glyphicon glyphicon<%= (keys.to_s.include?("Hardware")? "-minus-sign":"-plus-sign")%>" aria-hidden="true"></span>
+                <%=keys%>
+              </a>
+            </h4>
+          </div>
+          <div id="<%=keys.to_s+"panel"%>" class="panel-collapse collapse <%="in" if keys.to_s.include?"Hardware"%>" role="tabpanel">
```
lzap: Ditto.

lzap:
Before you start reading this, please discover a host using this helper script: https://github.com/lzap/bin-public/blob/master/discover-host with "real" parameter:

```
 discover-host http://localhost:3000 admin:admin real
```

This gives us some decent fact list from DELL PowerEdge R810. Now to my review:

I think "Storage" should go first and be the only expanded by default. It looks like the "Hardware" section can be easily overhelmed with lots of cpu cores.

Move `hostname` from highlights, its static (never changes). The same for "os" which is always CentOS7.

I still see `ipaddress` in Miscellanaeous section. And I see some `ipmi_` facts in Network section. I think it is worth introducing "IPMI" section with all `ipmi_` facts.

Can you convert `blockdevice_*_size` to human-readable form (XYZ GB)?

Some facts are reported as JSON hashes, why is that @ohadlevy ?

Also can you do the "expansion" javascript for long lines? Or instead of javascript, you can make a link into /fact_values for the particular one to see it complete.

alongoldboim:
@lzap I've used the script it was very useful.
-moved Storage to be on top.
-cleared hostname, os from highlights.
-moved ipaddress to highlights, added a ipmi section when needed.
-blockdevice_*_size changed to GB.
-added a tool tip for truncated values.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,8 +26,19 @@ def index
 
   def show
     # filter graph time range
+    @general =    {:Highlights => ["size", "productname", "memorysize", "interface", "manufacturer", "architecture", "macaddress", "processorcount", "physicalprocessorcount", "discovery_boot"]}
+    @highlights = {:Storage => ["blockdevice"],
+                   :Hardware => ["hardw", "manufacturer", "memo", "process"],
+                   :Network => ["ipaddress", "interfaces", "dhcp", "fqdn", "hostname", "link", "mtu", "net"] + @host.facts_hash["interfaces"].split(","),
+                   :Software => ["bios", "os", "discovery"]}
+    @facts = @host.facts_hash
+    @facts.delete("memorysize_mb")
```
lzap: I'd prefer not to delete facts, can you simply move it to Hardware?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,8 +26,19 @@ def index
...
+    @facts.each do |key, val|
+       @highlights.merge!({:IPMI => ["ipmi"]}) if key.to_s.downcase.include?("ipmi")
+       @facts[key] = number_to_human_size(val)  if key.to_s.downcase.include?("blockdevice") && key.to_s.downcase.include?("_size")
```
lzap: Can you use (case insensitive) regular expressions which are faster than this? Also you can leverage begin/end of the string.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -65,5 +65,11 @@ def discovery_status_icon(host)
 
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
+
+  end
+
+  def breaking_wrap_wrap(txt, col = 80)
+    txt.gsub(/(.{1,#{col}})( +|$\n?)|(.{1,#{col}})/,
+             "\\1\\3\n")
   end
```
lzap, line 83: Ok please remove unnecessary wrapping and just truncate long facts appending a "..." there. Make a sane default maximum length so it fits there.

lzap:
Ok few comments, it looks good. Can you please add a button between Select Action and Delete "Expand All" with a proper function? No need to do a Collapse All button I think. Then we are RTM!

app/controllers/discovered_hosts_controller.rb
alongoldboim: can i make @facts a shallow copy of the facts hash and then the source will not be influenced?

app/helpers/discovered_hosts_helper.rb
alongoldboim, line 83: Ok wanted to give the facts a bit more length without overflowing, but like you said ill just truncate it.

lzap:
LGTM @ohadlevy opinions about this UI?

ohadlevy:
can you attach recent UI?

alongoldboim:
selection_006

ohadlevy:
I think some of the facts name needs to have pretty names (like
physicalprocesscount) you could consider using table-zebra etc

lzap:
Hmm I don't like changing expected names much, zebra is a good idea. Thanks.

alongoldboim:
Added zebra to the list, imho we should add assets dir and place scripts and css in an organized way.
Your call :), Thanks for the comments.

lzap:
Ok last issues:
- the expand all button has wrong text, should be: "Expand All"
- can you please move all blockdevice_ facts from Highlights back to Storage (as it is pre-expanded and better place to have them)
- I still see ipmi_ facts both in IPMI and Network groups (should only be in IPMI group)

That's it!

alongoldboim:
@lzap should be good now :+1:

lzap:
Ok looks good, can you reword the commit message? I see some extra text there. Also, why you are not the author? Looks like you messed up with git a bit. This is what I get when I try to commit:

```
Author:     Stephen Benjamin <stbenjam@redhat.com>
AuthorDate: Wed Jul 22 14:36:13 2015 -0400
Commit:     Lukas Zapletal <lzap+git@redhat.com>
CommitDate: Wed Aug 26 13:44:11 2015 +0200

    Fixes 11008 - Changed the view for discovered hosts show page.

    chagne to impi:

    chagne to impi:
```

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,8 +26,19 @@ def index
...
   def show
     # filter graph time range
+    @general =    {:Highlights => [ "productname", "memorysize", "interface", "manufacturer", "architecture", "macaddress", "processorcount", "physicalprocessorcount", "discovery_boot"]}
+    @categories = {:Storage => ["blockdevice"],
+                   :Hardware => ["hardw", "manufacturer", "memo", "process"],
+                   :Software => ["bios", "os", "discovery"]}
+    @facts = @host.facts_hash.clone
+    @facts.delete("memorysize_mb")
+    @facts.each do |key, val|
+      @categories.merge!({:IPMI => ["ipmi"]}) if key.to_s.downcase.include?("ipmi") && !@categories[:IPMI]
```
lzap: Please do `if /^ipmi_/.match(key)` here

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,55 @@
...
+        </div>
+      <% end -%>
+    </div>
+  </div>
+</div>
+<script type="text/javascript" charset="utf-8">
+  $(document).ready(function(){
+    $('.facts-list > li:nth-child(even)').css('background-color','#f5f5f5')
+  });
+</script>
```
lzap: Missing newline

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,55 @@
...
+  <% @general.each do |keys, val| %>
+    <div class="col-md-6">
+      <div class="panel panel-default">
+        <div class="panel-heading" ><strong> <%= keys %></strong> </div>
+          <div class="panel-body">
+            <ul class="list-group facts-list">
+              <% @facts.sort.select{|key, value| val.any?{|v| (key.include?(v) &&  !key.include?("bios") && !key.include?("ipmi")) || (key.eql? "ipaddress")}}.each do |fact| %>
+                <li class="list-group-item"> <strong> <%= fact[0] %>: </strong>  <%= truncate(fact[1], :length => 50)%></li>
+                <% @facts.delete(fact[0]) %>
+              <% end %>
```
lzap: Ok this is code logic in view template, not good. Can you move all the code logic in the controller? Pass the data as hash of hashes and in views just blindly render them.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -25,8 +26,19 @@ def index
 
   def show
     # filter graph time range
+    @general =    {:Highlights => [ "productname", "memorysize", "interface", "manufacturer", "architecture", "macaddress", "processorcount", "physicalprocessorcount", "discovery_boot"]}
+    @categories = {:Storage => ["blockdevice"],
+                   :Hardware => ["hardw", "manufacturer", "memo", "process"],
+                   :Software => ["bios", "os", "discovery"]}
```
lzap: I wonder if you can make this configurable, my idea is that every category could be represented using _single_ regexp. Then you can load these from Settings (create new four settings named `discovery_storage_CATEGORY`) and provide sane default values for them. Example for hardware: `^(hardw|manufacturer|memo|process)".

Please get rid of other extra fact name comarisons in the code. Everything should have been categorized by the four regexps. No other include/match/delete calls in the code.

lzap: In the regexp try to follow the pattern of `^(prefix1|prefix2)` but in case you want to compare by suffixes you can of course do something like `((suffix1|suffix2)$)|(^(prefix1|prefix2))`

lzap: Wait, intstead of sane default, make empty value the default one and if there is no user defined one, use the pre-defined regexp. That's better for future upgrades.

alongoldboim:
@lzap applied the changes.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -24,11 +25,46 @@ def index
...
   def show
-    # filter graph time range
     @range = nil
-
-    # summary report text
     @report_summary = nil
+    regex_hightlights = Setting[:discovery_CATEGORY_highlights].empty? ? /^(productname|memorysize|interface|manufacturer|architecture|macaddress$|processorcount|physicalprocessorcount|discovery_boot|ipaddress$)/ : Regexp.new(eval(Setting[:discovery_CATEGORY_highlights]))
+    regex_storage = Setting[:discovery_CATEGORY_storage].empty? ? /^blockdevice/ : Regexp.new(eval(Setting[:discovery_CATEGORY_storage]))
+    regex_hardware = Setting[:discovery_CATEGORY_hardware].empty? ? /^(hardw|manufacturer|memo|process)/ : Regexp.new(eval(Setting[:discovery_CATEGORY_hardware]))
+    regex_network = Setting[:discovery_CATEGORY_network].empty? ? /^(ipaddress|interfaces|dhcp|fqdn|hostname|link|mtu|net)|(#{@host.facts_hash["interfaces"].gsub(",","|")})/ : Regexp.new(eval(Setting[:discovery_CATEGORY_network]))
```
lzap: Hmmm I'd rather keep all the regexps clean and static. I think if we copy few facts into highlights AND network, it's no issue. Just rewrite it to be simple regexp without this magic.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -24,11 +25,46 @@ def index
...
+        when regex_network
+          @categories[:Network][key] = value
+        when regex_ipmi
+          @categories[:IPMI][key] = value
+        else
+          @categories[:Miscellaneous][key] = value
+      end
+      @facts.delete(key)
+    end
+    @categories.each {|key,val| @categories.delete(key) if @categories[key].empty? }
```
lzap: Why don't you put regexps into an array, so you can iterate over groups. Then you can get rid of lines 41-46 and also get rid of that case statement (iterate through all regexps, if any match then determine group name, put it into the correct group name). Then it will be much shorter.

app/controllers/discovered_hosts_controller.rb
lzap: Reading your logic below, it looks like you pick up first, so there must be a way to achieve the same result without generating regexp from interfaces.

app/controllers/discovered_hosts_controller.rb
lzap: Make sure you process group names array in correct order. Putting it into an array should be sufficient.

app/models/setting/discovered.rb
```diff
@@ -42,6 +48,18 @@ def self.load_defaults
...
+          self.set('discovery_CATEGORY_highlights', N_("The default regex to organize facts for highlights section"), "" ),
+          self.set('discovery_CATEGORY_storage', N_("The default regex to organize facts for storage section"), "" ),
+          self.set('discovery_CATEGORY_software', N_("The default regex to organize facts for software section"), "" ),
+          self.set('discovery_CATEGORY_hardware', N_("The default regex to organize facts for hardware section"), "" ),
+          self.set('discovery_CATEGORY_network', N_("The default regex to organize facts for network section"), "" ),
+          self.set('discovery_CATEGORY_ipmi', N_("The default regex to organize facts for ipmi section"), "" ),
+      ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
+    end
+
+
```
lzap: I apologize for complications, but I like `discovery_facts_highlights` much better than (capitalized) CATEGORY.

lzap:
Ok that looks so much better.

app/controllers/discovered_hosts_controller.rb
alongoldboim: OK, will figure out a way, but just to know whats so bad in generating a regex from the interfaces?

app/controllers/discovered_hosts_controller.rb
alongoldboim: OK sounds good.

alongoldboim:
Thanks it will be done soon :)

alongoldboim:
@lzap fixed, there are two versions one with iterating on the array of regex and the other with the case, tell me what commit you prefer, thanks.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -171,6 +177,27 @@ def auto_provision_all
...
+    hardware = Setting[:discovery_facts_hardware].empty? ? /^(hardw|manufacturer|memo|process)/ : Regexp.new(eval(Setting[:discovery_facts__hardware]))
+    network = Setting[:discovery_facts_network].empty? ? /^(ipaddress|interfaces|dhcp|fqdn|hostname|link|mtu|net|macaddress|wol|port|speed)/ : Regexp.new(eval(Setting[:discovery_facts_network]))
+    software = Setting[:discovery_facts_software].empty? ? /^(bios|os|discovery)/ : Regexp.new(eval(Setting[:discovery_facts_software]))
+    ipmi = Setting[:discovery_facts_ipmi].empty? ? /^ipmi/ : Regexp.new(eval(Setting[:discovery_facts_ipmi]))
+    @regex_array = [hightlights, storage, hardware, network, software, ipmi]
+    @categories = Array.new(6) { Hash.new }
+    @categories_names = [:Hightlights, :Storage, :Hardware, :Network, :Software, :IPMI]
+  end
+
+  def assign_to_category(key, value )
```
lzap: Maybe `assign_fact_to_category` ? Also extra space.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -10,16 +10,22 @@ def discovered_hosts_title_actions(host)
...
-                                method = action[2] if action.size > 1
-                                link_to(action[0] , action[1], method)
-                              end.flatten
-        ),
+      select_action_button( _("Select Action"), {},
+                            actions.map do |action|
+                              method = action[2] if action.size > 1
+                              link_to(action[0] , action[1], method)
+                            end.flatten
+      ),
```
lzap: Please get rid of code reindenting. This makes cherry picking hard.

app/views/discovered_hosts/show.html.erb
```diff
@@ -2,15 +2,52 @@
...
+      <% end -%>
+    </div>
+  </div>
+</div>
+<script type="text/javascript" charset="utf-8">
+  $(document).ready(function(){
+    $('.facts-list > li:nth-child(even)').css('background-color','#f5f5f5')
+  });
+</script>
+
```
lzap, line 95: Extra newline :-)

lzap:
Ok this looks so much better, now please get rid of duplicating of facts and deleting facts. It is not necessary, you don't use @fact variable anymore in the template. No need to delete anything. Also modify your code in a way that if there is no match found, assign to the "Misceleneous" or "Other" section which is currently missing, but should have been there (no regexp, its a "fallback" group).

alongoldboim:
@lzap done.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -65,5 +71,7 @@ def discovery_status_icon(host)
 
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
+
   end
+
 end
```
lzap, line 84: Please remove all extra reformatting and empty lines.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -17,9 +17,15 @@ def discovered_hosts_title_actions(host)
...
+          $('.facts-panel').addClass('collapse in').height('auto');"
+        )
+      ),
+      button_group(
         link_to(_("Delete"), hash_for_discovered_host_path(:id => host),
-                :class => "btn btn-danger", :confirm => _('Are you sure?'), :method => :delete)
+          :class => "btn btn-danger", :confirm => _('Are you sure?'), :method => :delete)
       )
+
     )
```
lzap, line 38: Same here.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -171,6 +174,29 @@ def auto_provision_all
...
   private
 
+  def init_regex_and_categories
+    hightlights = Setting[:discovery_facts_highlights].empty? ? /^(productname|memorysize|interface|manufacturer|architecture|macaddress$|processorcount|physicalprocessorcount|discovery_boot|ipaddress$)/ : Regexp.new(eval(Setting[:discovery_facts_highlights]))
+    storage = Setting[:discovery_facts_storage].empty? ? /^blockdevice/ : Regexp.new(eval(Setting[:discovery_facts_storage]))
+    hardware = Setting[:discovery_facts_hardware].empty? ? /^(hardw|manufacturer|memo|process)/ : Regexp.new(eval(Setting[:discovery_facts__hardware]))
+    network = Setting[:discovery_facts_network].empty? ? /^(ipaddress|interfaces|dhcp|fqdn|hostname|link|mtu|net|macaddress|wol|port|speed)/ : Regexp.new(eval(Setting[:discovery_facts_network]))
+    software = Setting[:discovery_facts_software].empty? ? /^(bios|os|discovery)/ : Regexp.new(eval(Setting[:discovery_facts_software]))
+    ipmi = Setting[:discovery_facts_ipmi].empty? ? /^ipmi/ : Regexp.new(eval(Setting[:discovery_facts_ipmi]))
+    @regex_array = [hightlights, storage, hardware, network, software, ipmi, "misceleneous"]
```
lzap: Use symbol instead. Alternatively, you can use the index, but I kinda like this solution. Another more clever approach is to create a class named `AlwaysMatcher` with one method `def match; true; end` adding it as the last element. Then the loop can be even more simple (it will always match). But this works for me too.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -24,11 +25,13 @@ def index
   end
 
   def show
-    # filter graph time range
     @range = nil
-
-    # summary report text
     @report_summary = nil
```
lzap, line 32: And please keep the comments here, we shoould not remove those important lines.

lzap:
Ok almost there, few more glitches and then I will re-test :-)

app/controllers/discovered_hosts_controller.rb
alongoldboim: changed it a bit, didn't use ur solution but improved the one i had.

alongoldboim:
@lzap finished, I think we have reached the holy land :+1:

app/controllers/discovered_hosts_controller.rb
```diff
@@ -29,6 +30,11 @@ def show
 
     # summary report text
     @report_summary = nil
+    init_regex_and_categories
+    @host.facts_hash.each do |key, value|
+      value = number_to_human_size(value)  if /size$/.match(key)
```
lzap: Double space before `if`.

Can you also proceed with expanding Settings chapter in our manual? http://theforeman.org/plugins/foreman_discovery/4.0/index.html#3.1.4Globalsettings it is this page: https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/foreman_discovery/4.0/index.md

alongoldboim:
Fixed the double space.
the pr for the documentation part https://github.com/theforeman/theforeman.org/pull/434, should be merged after this pr will be merged.

lzap:
Great, thanks. Can only merge after we fix unit tests on develop.

lzap:
Ok we can re [test] now.

alongoldboim:
Cool, thanks for the reviews @lzap .

lzap:
> Cool, thanks for the reviews @lzap .

Ok, I think we can extend this further. Would you mind showing also
interfaces that were discovered?

It's host.interfaces entity, basically in a table with the following
columns:

Primary/Provisioning icons
Identifier
Type
MAC
IP

Very similar to New Host - Interfaces table.

This would be very useful!

## 

Later,
 Lukas #lzap Zapletal

alongoldboim:
@lzap Yea we can do that, we already doing that but we are not centering all of the facts together in a table, ill add it, you want to place it under a "Interfaces" category? or place it under an existing one?
Do we want to keep on using our solution with the regex on settings? seen in the pr you referenced that you think about doing it differently.

alongoldboim:
Current UX.
selection_007

kybaker:
@alongoldboim - You sent a request for me to review. 

The design above looks nice at first glance but it will prove to be cumbersome to use over time. It looks like the goal of this page is to allow users to scan and process data as quickly as possible. Expanding and collapsing adds an extra step to this process. In my experience performing usability testing, users do not want to have to click anything, they want information available to scan as quickly as possible. We are adding an extra step here. I think adding some simple headers above a table for each section would be the most effective user experience.

alongoldboim:
@kybaker thanks for the response, i agree with what you'r saying yet there is a problem with expending all the tabs by default because there are lots of facts (can even get to 5 pages of scrolling), that's why we got the highlights section(expanded by default) which should answer most of the user needs and the other facts are organized in categories to help the user get to the needed fact faster by knowing what its connected to.
In the image you can see the scroller size when all is expanded.
scroller_demo

do you still think we should cancel the expanding?

kybaker:
I understand and parsing the information here is beneficial. I don't think this will be a show stopper but we could always change if there is negative feedback.

alongoldboim:
Sounds good, thanks.

lzap:
Yeah I agree with @alongoldboim - we expand the most important stuff anyway, but big servers with hudreds of cores generates _huge_ output. Its like neverending.

@alongoldboim ok your first version looks good, haven't tested it yet but can you make the table padding the same size as in other panes? Also remove the "Type" column and just add icons as we do on the Interfaces tab on the Edit Host form (primary, provisioning) with mouse hoovers. And that should do it!

alongoldboim:
@lzap Done, just how i'm supposed to decide whether primary, provisioning flags are active?.

lzap:
Ok almost there:
- Can you add 'Back' button that does browser's back action?
- Expand Interfaces pane by default
- Use 'break-me' CSS class I've introduced in Core last week for fact value cells (https://github.com/theforeman/foreman/commit/c687f3a1c9d157b1313070ddc005b518af30acad)
- Let's still use your trimming code, but now for very long strings. I suggest to trim only values longer than 1000 characters, what you think? Or maybe we can get rid of this completely, if it still looks okay with real data.

alongoldboim:
Added the requirements, but some of the facts values are placed in 'li' and not in 'td', if I add this class to those elements it doesn't looks so good(I can either change it all to tables or for first version leave it as it is).
About the trimming it works well with all the data I have checked, i think we'd better use the current trimming because it prevents overflow and if we will have issues we could change it.

lzap:
Discovered hosts have `.subnet` association, can you please add it as an extra fact to Highlight pane? Would be named "discovery_subnet", you can render subnet name plus the subnet address with a link to the subnet:

discovery_subnet: mysubnet (192.168.44.0)

alongoldboim:
Added this fact at the format you wanted.
There is no show page for subnet, i can link it to the edit page but i don't know how valuable it will be, for now i left it as a regular fact without a link, let me know if you think differently.

lzap:
Oh right, I think link is not necessary youre right. Testing.

lzap:
Ok, I still don't like how trimming works, I think we trim too much on larger screens. Isn't there any other way? Could you talk to Shimon who implemented the "eclipse" CSS style and use that at least? This will only trim if there is no free space. Other than that we are good to go! :-)

alongoldboim:
It was Tomer :), added and working, give it a check.

lzap:
Ok two last things:

1) N+1 query

```
| N+1 Query detected
 |   Nic::Managed => [:subnet]
 |   Add to your finder: :includes => [:subnet]
 | N+1 Query method call stack
 |   /home/lzap/work/foreman/app/models/host/base.rb:83:in `subnet'
 |   /home/lzap/work/foreman_discovery/app/controllers/discovered_hosts_controller.rb:304:in `add_custom_facts'
 |   /home/lzap/work/foreman_discovery/app/controllers/discovered_hosts_controller.rb:39:in `show'
 |   /home/lzap/work/foreman/app/controllers/concerns/application_shared.rb:13:in `set_timezone'
 |   /home/lzap/work/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   /home/lzap/work/foreman/lib/middleware/catch_json_parse_errors.rb:9:in `call'
 | 
 | /home/lzap/work/foreman/app/models/host/base.rb:83:in `subnet'
 | /home/lzap/work/foreman_discovery/app/controllers/discovered_hosts_controller.rb:304:in `add_custom_facts'
 | /home/lzap/work/foreman_discovery/app/controllers/discovered_hosts_controller.rb:39:in `show'
 | /home/lzap/work/foreman/app/controllers/concerns/application_shared.rb:13:in `set_timezone'
 | /home/lzap/work/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
```

2) The interfaces table lost it's padding, can you put it back?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -262,4 +293,17 @@ def skip_bullet
...
+
+  def get_interfaces
+    @host.interfaces.each do |interface|
+      @interfaces << {:identifier => interface["identifier"], :type => interface["type"], :mac => interface["mac"], :ip => interface["ip"]? interface["ip"] : "N/A", :primary => interface["primary"], :provision => interface["provision"]}
+    end
+  end
+
+  def add_custom_facts
+    if @host.subnet
+      discovery_subnet = "#{@host.subnet.name + " (" + @host.subnet.network + ")"}"
```
lzap: Weird usage of embedding :-) I'd probably go with `"#{host.subnet.name} (#{host.subnet.network})"`.

lzap:
Actually the pane with no padding is not bad at all, if you can do the same for all that's fine. But I see some problem at the bottom line, there is some unexpected padding still.

9a9af915-7417-4caf-9901-d3431a42c7c5

lzap:
The puzzle solution is:

``` diff
diff --git a/app/controllers/discovered_hosts_controller.rb b/app/controllers/discovered_hosts_controller.rb
index 09e2d24..acef4e6 100644
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -244,7 +244,7 @@ class DiscoveredHostsController < ::ApplicationController
     end
   end

-  def find_by_name(includes = "")
+  def find_by_name(*includes)
     params[:id].downcase! if params[:id].present?
     @host = includes.empty? ? resource_base.find_by_id(params[:id]) : resource_base.includes(includes).find_by_id(params[:id])
     @host ||= includes.empty? ? resource_base.find_by_name(params[:id]) : resource_base.includes(includes).find_by_name(params[:id])
@@ -252,7 +252,7 @@ class DiscoveredHostsController < ::ApplicationController
   end

   def find_by_name_incl_subnet
-    find_by_name(:subnet)
+    find_by_name({:interfaces => :subnet})
   end

   def find_multiple
@@ -306,7 +306,7 @@ class DiscoveredHostsController < ::ApplicationController

   def add_custom_facts
     unless @host.primary_interface.subnet.nil?
-      discovery_subnet = @host.primary_interface.subnet.name + "(" + @host.primary_interface.subnet.network + ")"
+      discovery_subnet = "#{@host.primary_interface.subnet.name} (#{@host.primary_interface.subnet.network})"
       assign_fact_to_category("discovery_subnet", discovery_subnet)
     end
   end
```

lzap:
Ok good work on this one.

app/models/setting/discovered.rb
```diff
@@ -42,6 +48,16 @@ def self.load_defaults
...
     end
 
+    Setting.transaction do
+      [
+          self.set('discovery_facts_highlights', N_("The default regex to organize facts for highlights section"), "" ),
+          self.set('discovery_facts_storage', N_("The default regex to organize facts for storage section"), "" ),
+          self.set('discovery_facts_software', N_("The default regex to organize facts for software section"), "" ),
+          self.set('discovery_facts_hardware', N_("The default regex to organize facts for hardware section"), "" ),
+          self.set('discovery_facts_network', N_("The default regex to organize facts for network section"), "" ),
+          self.set('discovery_facts_ipmi', N_("The default regex to organize facts for ipmi section"), "" ),
```
lzap: Three more nitpicks:

1) Drop the `The default" from the setting name, I'd go with the short "Regex to organize..."

2) Extra space after `""`

3) Rewrite the commit message, it contains some garbage

lzap: Scratch the (4), we can do separate PR for `""`'s.

lzap:
Merged as 8b23c24a667f4e8588f96a346432aafeab63ba85, thank you!

PR #202, opened 2015-07-24
Fixes #10469 - enforced discovery rule taxonomy

lzap:


app/models/discovery_rule.rb
```diff
@@ -37,4 +38,16 @@ def default_int_attributes
    self.max_count ||= 0
    self.priority  ||= 0
   end
+
+  def enforce_taxonomy
+    return if hostgroup.nil?
+    unless (ms = hostgroup.organizations - organizations).empty?
+      names = ms.collect(&:name).inspect
```
domcleal: Prefer .to_sentence over inspect, it's more user-friendly - or leave the names out.

lzap:
Reworded and leverated `to_sentence` call.

@orrabin this goes to develop only and 3.0.1 only if we introducing taxonomies there. I believe we don't.

lzap:
@orrabin I have doublechecked, this belongs only to develop for 4.0 release.

app/models/discovery_rule.rb
```diff
@@ -37,4 +38,16 @@ def default_int_attributes
    self.max_count ||= 0
    self.priority  ||= 0
   end
+
+  def enforce_taxonomy
+    return if hostgroup.nil?
+    unless (ms = hostgroup.organizations - organizations).empty?
+      names = ms.collect(&:name).to_sentence
+      errors.add(:organizations, _("Hostgroup organization(s) %s must also be associated to the discovery rule") % names)
```
domcleal: "Host group" if you can.

lzap: > "Host group" if you can.

Done.

## 

Later,
 Lukas #lzap Zapletal

lzap:
@orrabin can you take a look on this one? We need this for 4.0 discovery.

lzap:
Once this is merged we can start testing develop against 1.9 foreman and go ahead with 4.0 final release.

app/models/discovery_rule.rb
```diff
@@ -37,4 +38,16 @@ def default_int_attributes
    self.max_count ||= 0
    self.priority  ||= 0
   end
+
+  def enforce_taxonomy
+    return if hostgroup.nil?
+    unless (ms = hostgroup.organizations - organizations).empty?
```
orrabin: This should only be checked when organizations are enabled, same below for locations.

lzap:
Fixed.

test/unit/discovery_rule_test.rb
```diff
@@ -1,20 +1,59 @@
 require 'test_helper'
 
 class DiscoveryRuleTest < ActiveSupport::TestCase
+  setup do
+    @hg_org = FactoryGirl.create(:organization, :name => 'hgorg')
+    @hg_org.save!
```
orrabin, line 6: why do you need to save after you create?

orrabin:
@lzap it works well for me

test/unit/discovery_rule_test.rb
```diff
@@ -1,20 +1,59 @@
...
+    @hostgroup = FactoryGirl.create(:hostgroup)
+    @hostgroup.organizations << @hg_org
```
orrabin, line 8: maybe add this to the line above with `:organizations => [@hg_org]` and then you won't need to save again?

test/unit/discovery_rule_test.rb
```diff
@@ -1,20 +1,59 @@
...
+    @hostgroup.save!
+  end
 
   test "should be able to create a rule without entering hosts limit" do
-    rule = DiscoveryRule.new :name => "myrule", :search => "cpu_count > 1", :priority => "1", :hostgroup_id => "1"
+    rule = DiscoveryRule.new :name => "myrule",
```
orrabin, line 13: can this be tested with FactoryGirl?

test/unit/discovery_rule_test.rb
```diff
@@ -1,20 +1,59 @@
...
+    refute_valid rule
+  end
+
+  test "should enforce hostgroup organizations" do
+    skip unless SETTINGS[:organizations_enabled]
+    rule = DiscoveryRule.new :name => "myrule",
+      :search => "cpu_count > 1",
+      :hostgroup_id => @hostgroup.id
     refute_valid rule
+    assert_equal 'Host group organization hgorg must also be associated to the discovery rule', rule.errors[:organizations].first
```
orrabin: I'd use `@hg_org.name` instead of writing `hgorg` in case anyone ever changes the org and also it took me a second to figure out this was the name for the org above.

test/unit/discovery_rule_test.rb
lzap, line 6: Yeah removed.

test/unit/discovery_rule_test.rb
lzap, line 8: Yes but then I'd need to drop my setup helper, I think this is fine.

test/unit/discovery_rule_test.rb
lzap, line 13: Well I am actually testing AR validations, how to do that with FactoryGirl?

orrabin, line 13: I think it's the same but since you aren't saving the rules it should be fine.

orrabin:
[test]

orrabin:
merged as 5cb015eb6ab9be490956af00300bdd1df94c8d18, thanks @lzap

PR #205, opened 2015-08-07
Add the app/views path of rabl view path

phirince:
This is related to this issue on foreman_hooks:

https://github.com/theforeman/foreman_hooks/issues/24

Basically, the foreman_hooks is not able to find the rabl templates of disovered_hosts because it is not being added to the view_path. I think, to fix this issue and to make this a more complete plugin, this should be included in the view path. This also requires a minor change in the foreman code base as well. I'll submit it and update it here.

Please review this.
-PP


domcleal:
A couple of things on this one:
1. Be sure to do `||= []` first and use `<<` in case we have two plugins that want to add to view_paths before Foreman itself.
2. For plugins, we typically write initialisers in lib/*/engine.rb which might be preferable to using config/initializers/ in a plugin.

[test]

domcleal:
> Be sure to do ||= [] first

Correction, as per https://github.com/theforeman/foreman/pull/2596#discussion_r36612886 I don't think this part's needed.  The bit that is important is to use `<<` so as not to overwrite the whole array.

phirince:
Please check my latest commit. 
Will it be better to do this?

config.view_paths += ForemanDiscovery::Engine.paths['app/views'].existent

-PP

domcleal:
Looks good to me.  I've checked it initialises the view_paths correctly, but haven't checked it works with hooks to render yet.

lzap:
Also there are four commits with some extra merge commit that should not be present. Please squash into single one for the final proposal.

PR #209, opened 2015-08-13
Bumped Foreman dependency in engine.rb, extracted strings

lzap:
Forgot to do this for 4.0 release but we can rollout 4.0.1 gem only, it only
affects gem installations.

### DIFF
--- a/locale/de/foreman_discovery.po
+++ b/locale/de/foreman_discovery.po
@@ -1,7 +1,7 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the foreman package.
-# 
+#
 # Translators:
 # Arnold Bechtoldt <mail@arnoldbechtoldt.com>, 2014
 # Christina Gurski <Gurski_christina@yahoo.de>, 2015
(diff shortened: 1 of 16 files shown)


lzap:
Added string extraction and new simple make target that extract via CLI utility available in gettext gem:

```
make -C locale/ extract
```

I also merged/updated PO files:

```
make -C locale/ update-po
```

@domcleal what you think, at least as a temporary solution until https://github.com/theforeman/foreman/pull/2096 is fixed...

locale/de/foreman_discovery.po
```diff
@@ -415,10 +446,61 @@ msgstr "Anzahl der Einträge pro Anfrage"
...
+#~ msgid "The default fact name to use for the MAC of the system"
+#~ msgstr "Der Standard-Faktenname zum Verwenden für die MAC des Systems"
+
+#~ msgid "Total pool size"
+#~ msgid_plural "Total pool size"
+#~ msgstr[0] "Gesamt-Pool-Größe"
+#~ msgstr[1] "Gesamt-Pool-Größe"
+
+#~ msgid "hash containing facts for the host"
+#~ msgstr "Hash, der Fakten für den Host enthält"
```
lzap, line 506: Was not sure how to disable those "backups". I assume these gets overwritten after we pull from Transifex...

lzap:
When merging this please cherry pick into 4.0 as well

lzap:
Can update http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Extracting-strings, if you like this

domcleal:
No, I don't think Foreman's advice should change, we should fix the problem in core.  It's fine if you want to do it in some plugin, but if we encourage others to do it, they may miss extra configuration we keep in core, support for JavaScript etc.

lzap:
Ok made a note in the http://projects.theforeman.org/projects/foreman/wiki/Release_Process_Plugins

@orrabin can you review and merge this? I'd like to ask translators for updating during tomorrows announcement pulling translations in 4.0.1.

orrabin:
@lzap do we really want the version bump in develop? if we are planing another 3.x version maybe put the version bump only in 4.0 for now or at least have it in a different PR so we can cherry-pick the translations to 3.0-stable without needed to fix it?

lzap:
> @lzap do we really want the version bump in develop? if we are planing another 3.x version maybe put the version bump only in 4.0 for now or at least have it in a different PR so we can cherry-pick the translations to 3.0-stable without needed to fix it?

I don't plan any 3.0.x release. I know we've talked about this and it
looks like 3.0 series is not the best when it comes to stability, but
the most annoying bug of facter importer extensions not being effective
is quite a change for a point release.

Please merge into develop and cherry pick into 4.0

## 

Later,
 Lukas #lzap Zapletal

orrabin:
merged as fe7254ea090a5eaec8c625081787d5916bed9ff6, thanks @lzap

PR #212, opened 2015-08-26
Fixes #11574 - rule is deleted with hostgroup

lzap:


lzap:
For the record, to test the nullify statement you need to provision a host via a rule, then change hostgroup of the host to something else, then delete the original hg which should set rule to nil.

orrabin:
@lzap the `:dependent => :destroy` works fine but I tried to follow your steps to test the nullify and when trying to delete the hostgroup I got: `Host(#74808040) expected, got Host::Managed(#78344660)`

part of the stack trace:

```
activerecord (3.2.21) lib/active_record/associations/association.rb:210:in `raise_on_type_mismatch'
activerecord (3.2.21) lib/active_record/associations/collection_association.rb:453:in `block in delete_or_destroy'
activerecord (3.2.21) lib/active_record/associations/collection_association.rb:453:in `each'
activerecord (3.2.21) lib/active_record/associations/collection_association.rb:453:in `delete_or_destroy'
activerecord (3.2.21) lib/active_record/associations/collection_association.rb:231:in `delete'
activerecord (3.2.21) lib/active_record/associations/collection_association.rb:157:in `delete_all'
activerecord (3.2.21) lib/active_record/associations/collection_proxy.rb:46:in `delete_all'
activerecord (3.2.21) lib/active_record/associations/builder/has_many.rb:56:in `block in define_nullify_dependency_method'
activesupport (3.2.21) lib/active_support/callbacks.rb:418:in `_run__4333952607528659708__destroy__206682355595937245__callbacks'
activesupport (3.2.21) lib/active_support/callbacks.rb:405:in `__run_callback'
activesupport (3.2.21) lib/active_support/callbacks.rb:385:in `_run_destroy_callbacks'
activesupport (3.2.21) lib/active_support/callbacks.rb:81:in `run_callbacks'
activerecord (3.2.21) lib/active_record/callbacks.rb:254:in `destroy'
activerecord (3.2.21) lib/active_record/transactions.rb:254:in `block in destroy'
activerecord (3.2.21) lib/active_record/transactions.rb:313:in `block in with_transaction_returning_status'
activerecord (3.2.21) lib/active_record/connection_adapters/abstract/database_statements.rb:192:in `transaction'
activerecord (3.2.21) lib/active_record/transactions.rb:208:in `transaction'
activerecord (3.2.21) lib/active_record/transactions.rb:311:in `with_transaction_returning_status'
activerecord (3.2.21) lib/active_record/transactions.rb:254:in `destroy'
```

lzap:
I will add tests and change both to `nullify`. Note develop branch currently fails because of http://projects.theforeman.org/issues/11708

orrabin:
Tests haven't been added yet and this needs rebase.

lzap:
Rebased, added a test.

orrabin:
@lzap I'm getting: 
`
PG::Error: ERROR: null value in column "hostgroup_id" violates not-null constraint DETAIL: Failing row contains (4, test, 1=0, null, , 0, 0, t, 2016-01-12 07:11:37.175086, 2016-01-12 07:11:37.175086). : UPDATE "discovery_rules" SET "hostgroup_id" = NULL WHERE "discovery_rules"."id" IN (SELECT "discovery_rules"."id" FROM "discovery_rules" WHERE "discovery_rules"."hostgroup_id" = $1 ORDER BY discovery_rules.name)
`
when trying to delete the hostgroup with this patch on core develop branch.

lzap:
Ok `nullify` is not the best solution, we want to avoid rules without hostgroup. So I changed that to `restrict_with_error` which works great - it errors out when you attept to delete a hostgroup with rule associated. The message says it all: _Cannot delete record because dependent discovery rules exist_

The other direction stays `nullify` because that makes sense - it's just a counter basically.

lzap:
@orrabin ping - trying to hit 6.2 with this (customer facing)

orrabin:
@lzap I tested and it looks good to me.
The test is still for nullify - it tests that it won't use the rule without a hostgroup but now you can't get to that situation.
Do you mind adding a test that hostgroup isn't deleted when you have a rule that uses it?
After that I'll merge.

Edit: I see the test has changed but it checks for the exception when trying to use the rule and not that the hostgroup isn't actually deleted.  I think this could use another test but if it's not possible to recreate the situation I'll merge as is.

lzap:
@orrabin added the test, not sure about the place tho, this should perhaps be in the hostgroup test, but we don't have any in the plugin.

orrabin:
@lzap since we aren't extending hostgroup I'm fine with the test location. We may want to add some more tests for rules and hostgroups and then we should probably move all of them.
I'll merge when tests are green :)

orrabin:
looks unrelated but maybe [test] will work the second time

orrabin:
Thanks @lzap! merged as 82d93bfb7748482ca32ab293d8c0b47c38c4c018

PR #213, opened 2015-08-27
Fixes #9490 - Added setting to specify fact to use for hostname

orrabin:


app/models/host/discovered.rb
```diff
@@ -39,10 +39,22 @@ def self.import_host_and_facts facts
...
     hostname_prefix ||= 'mac'
-    hostname = FacterUtils::bootif_mac(facts).try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
+
+    macfact = facts[discovery_fact].try(:downcase)
+    raise(::Foreman::Exception.new(N_("Invalid facts: hash does not contain the required fact '%s'"), discovery_fact)) unless macfact
+    begin
+      macfact = Net::Validations.normalize_mac(macfact)
+    rescue ArgumentError => e
+      macfact = facts['discovery_bootif'].try(:downcase)
+    end
```
lzap, line 53: Looks like I have removed this block in 4c018bc136f7db433b3bbcfbdb97cca53bd3e64a

app/models/host/discovered.rb
```diff
@@ -39,10 +39,22 @@ def self.import_host_and_facts facts
...
+
+    name_fact = facts[Setting[:discovery_hostname]] || macfact
+    hostname = name_fact.try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
```
lzap, line 56: I don't like the prefix find/replace approacha anymore, can you concatenate it the correct way?

lzap, line 56: One problem here, if user sets any fact other than MAC address, we are replacing all comas here. I think we need to actually drop all non-ascii characters which are not accepted in host names. Please make this a helper method and use it also above.

app/models/host/discovered.rb
```diff
@@ -39,10 +39,22 @@ def self.import_host_and_facts facts
...
+    raise(::Foreman::Exception.new(N_("Invalid facts: hash does not contain the required fact '%s'"), name_fact)) unless hostname
```
lzap, line 57: I think lines 56-57 should be swapped, we want to check the content of the fact itself, not the resulting concatenated string (which will likely evaluate to true in most cases).

app/models/setting/discovered.rb
```diff
@@ -10,6 +10,7 @@ def self.load_defaults
     Setting.transaction do
       [
         self.set('discovery_fact', N_("Fact name to use for primary interface detection and hostname"), "discovery_bootif"),
+        self.set('discovery_hostname', N_("The default ERB to evaluate for the discovered host name"), "discovery_bootif"),
```
lzap, line 13: ERB? I don't see ERB templating in the patch.

imriz:
Hi,

My 2 cents - I think it might be better to support a list of facts (separated by a delimiter). We can try them one by one, at the given order, and use the first match. This gives more flexibility. 
If you think this is a good idea, we will contribute the code.

lzap:
I think this is doable and small effort, @orrabin can you incorporate this idea once you get to this? No rush, I know you're busy. Thanks!

imriz:
Any idea when this is going to be merged? This feature is one of the reasons we're maintaining our own fork of the discovery plugin (we've implemented something of this sort there), and I would like to be able to go back to using the vanilla plugin ;)

orrabin:
@imriz why not send your fix as a patch? I don't mind closing this, it's not ready and it will take some time for me to get back to working on it.

imriz:
Sure. However, our patch is very opinionated towards our environment (fact
name hard coded, no support for prefix etc). Mind if I will base my patch
on your patch? If so, do I need to squash your commit, or can I leave it,
and just add a commit on top of it?

orrabin:
@imriz you can squash my commit into yours, update the new pr number here and I'll close this one.

imriz:
@orrabin done

dLobatog:
@orrabin I'll close this as the other PR has been opened - reopen if you want to continue

PR #216, opened 2015-09-11
Bkearney/new downstream strings

bkearney:


lzap:
The other was merged.

PR #219, opened 2015-09-18
Fixes #11837 - optimized fact loading on index

lzap:


lzap:
Rebased @orrabin 

This fixes the unoptimized data fetch and also removed N+1 errors (via the `reload` trick).

Just for the record, the query cannot use pluck in Rails 3 or older versions, this code needs Rails 4+.

```
[10] pry(main)> host.fact_values.joins(:fact_name).where('fact_names.name' => Setting::Discovered.discovery_fact_column_array).pluck(:name, :value)
=> [["ipaddress", "10.84.245.135"], ["discovery_bootif", "52:54:00:fa:06:58"], ["memorysize_mb", "900"]]
[11] pry(main)> ActiveRecord::Base.connection.select_all(host.fact_values.joins(:fact_name).where('fact_names.name' => Setting::Discovered.discovery_fact_column_array).select("name, value")).collect{|x| [x["name"], x["value"]]}

=> [["ipaddress", "10.84.245.135"], ["discovery_bootif", "52:54:00:fa:06:58"], ["memorysize_mb", "900"]]
```

app/models/host/discovered.rb
```diff
@@ -81,6 +81,8 @@ def populate_fields_from_facts(facts = self.facts_hash, ignored_type)
     unless primary_ip.nil?
       subnet = Subnet.subnet_for(primary_ip)
       if subnet
+        # reload subnet to fetch all references
+        subnet.reload
```
unorthodoxgeek: what is this doing? don't understand the trick

lzap: This is to prevent N+1, @ares showed me this trick. Marek, do you remember which code in particular is causing the N+1 errors? I vaugly remember this is the interface detection code

unorthodoxgeek: this will trigger a query which wasn't triggered before. don't understand how that brings number of queries down...

lzap: Well, I removed this line from the patch. It was for something else.

app/models/host/discovered.rb
```diff
@@ -45,7 +45,7 @@ def self.import_host_and_facts facts
     hostname = FacterUtils::bootif_mac(facts).try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
 
     # create new host record
-    h = ::Host::Discovered.find_by_name hostname
```
unorthodoxgeek, line 51: I know this isn't yours, but if you're at it, can you change `h` into `host` ?
I really dislike pointless variable names, and it goes against the Ruby conventions...

app/views/discovered_hosts/_discovered_host.html.erb
```diff
@@ -5,6 +5,10 @@
 <td class="hidden-tablet hidden-xs"><%= number_to_human_size(discovery_attribute(host, :memory, 0) * 1024 * 1024) %></td>
 <td class="hidden-tablet hidden-xs"><%= discovery_attribute(host, :disk_count) %></td>
 <td class="hidden-tablet hidden-xs"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
-<% Setting::Discovered.discovery_fact_column_array.each do |fact_column| %>
-<td class="hidden-tablet hidden-xs"><%= host.facts_hash[fact_column.strip] || 'N/A' %></td>
+<%
+  # can be replaced with pluck in Rails 4+
```
unorthodoxgeek: aren't we already at rails 4+?

lzap: Strange, I had to mess up rebasing, this was already replaced with `host.fact_values.joins(:fact_name).where('fact_names.name' => Setting::Discovered.discovery_fact_column_array).pluck(:name, :value)`. This is the old version from Rail 3 time. Fixed, pushed, also renamed `host` above. I also moved the line to controller, which is a better fit.

lzap:
I rebased, any chance of reviewing @unorthodoxgeek or @orrabin ?

app/controllers/discovered_hosts_controller.rb
```diff
@@ -22,6 +22,9 @@ def index
       :model,
       :discovery_attribute_set
     ], {:interfaces => :subnet}).paginate(:page => params[:page])
+    @host_facts = Hash[@hosts.collect do |host|
```
unorthodoxgeek: mind putting the array into a variable so this line would be somewhat readable?

lzap: Yeah, looks weird. Rebased.

unorthodoxgeek:
seems good to me :+1:

orrabin:
@lzap I fixed the commit message on merge since you divided to 2 prs.

merged as c3b1b2cb06f8db1962b491f76c1f09e8e77fb700, thanks @lzap!
@unorthodoxgeek thanks for reviewing!

PR #225, opened 2015-10-22
Fixes #12256 - removed explicit transactions

lzap:
Causing issues since core orchestration is not ready for them.

### DIFF
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -37,6 +37,10 @@ module Host::ManagedExtensions
     true
   end
 
+  def delReboot
+    # no reboot on orchestration rollback
+  end
+
   def boot_url pxe_file
     raise ::Foreman::Exception.new(N_("Operating system not set for host/hostgroup")) unless operatingsystem
     base = operatingsystem.medium_uri(self)
@@ -56,6 +60,10 @@ module Host::ManagedExtensions
     true
   rescue ::Foreman::Exception
     true
+  end
+
+  def delKexec
+    # no kexec on orchestration rollback
   end
 
   def delete_discovery_attribute_set
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -76,12 +76,10 @@ module Api
       end
 
       def update
-        Host.transaction do
-          @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host)
-          forward_request_url
-          update_response = @host.update_attributes(params[:discovered_host])
-          process_response update_response
-        end
+        @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host)
+        forward_request_url
+        update_response = @host.update_attributes(params[:discovered_host])
+        process_response update_response
       end
 
       api :DELETE, "/discovered_hosts/:id/", N_("Delete a discovered host")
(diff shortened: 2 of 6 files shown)


iNecas:
I've tested the patch and I can confirm it fixes the issue with the post_commit happening while the orchestration fails: I've explicitely raise an error in `setTFTP` orchestration method, without this patch, the discovered host got restarted even when some previous orchestration code failed. With the patch, no restarted happened. I also can confirm the standard workflow with provisioning a discovered host works. I have not tested though all the cases where this PR touches the code.

Also, at least an issue should be raised on the foreman side about the nested transactions issue: this might cause other errors somewhere else then the discovery.

app/models/host/managed_extensions.rb
```diff
@@ -34,6 +34,10 @@ def setReboot
     true
   end
 
+  def delReboot
```
ohadlevy, line 40: not related to this patch?

lzap, line 40: > not related tot his patch?

It is, it removes exception from logs about missing rollback method when
transaction fails.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Rebased @orrabin

lzap:
Rebased again @orrabin

lzap:
And .... rebased once again :-)

dLobatog:
Merged as df4df986ce943e61cac2d71bd5af6e7a2816bb47, thanks @lzap!

lzap:
@iNecas I am evaluating patches for backporting into Satellite 6.2 and it looks like you had some problems which were solved by this patch. What was it?

iNecas:
The problem was, when some orchestration failed (such as setFTF), the host still got restarted even when the transaction failed. It's probably not that big deal, as if everything works, there are no rollbacks. The only issue here is, that there is a delay until the host gets restarted on rollback. The problem is conceptual: this means that if anyone wraps the `save` method with orchestration into transaction, they can expect similar issues as here, so the post_commit orchestration happens regardless success of failure of pre_commit orchestration.

iNecas:
I've filled an issue against Foreman describing what the real problem is http://projects.theforeman.org/issues/14001

lzap:
Ok, for the record, I am keeping this patch out of Satellite 6.2 as it's not blocker. It has been merged already and is available in Discovery 5.0.1 release.

lzap:
For the record, this also fixes known interaction in Rails (< 4.1.6) between ActiveRecord::Base#becomes and ActiveRecord::Base#save!. The issue is that the save! call creates an UPDATE like this: 

```
UPDATE "hosts" SET ... WHERE "hosts"."type" IN ('Host::Managed') AND "hosts"."id" = 18  [...]
```

The WHERE clause there matches nothing, because the record in the database still has a type of Host::Discovered. An UPDATE on an empty set doesn't raise an error, so execution continues, and the transaction is not rolled back. This means, if you convert a host from discovered to managed and then save! it inside a transaction, all of the SQL statements that handle deleting the discovered attributes, adding entries to hosts/puppetclasses/operating systems/hostgroups etc succeed, and at the end you are left with a broken discovered host with no attributes, that can't even be deleted because of database constraints.

This is only an issue in Rails versions <= 4.1.5, it was fixed with this commit in Rails 4.1.6: https://github.com/rails/rails/commit/257d65ed2ceb14dca6b1319355ed4f8a73540498 . Here is the associated issue: https://github.com/rails/rails/issues/14785

https://bugzilla.redhat.com/show_bug.cgi?id=1357583

PR #228, opened 2015-10-27
refs #3157 - rails 4 changes

unorthodoxgeek:
related to https://github.com/theforeman/foreman/pull/2870

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -5,7 +5,7 @@ class DiscoveryRule < ActiveRecord::Base
   include Parameterizable::ByIdName
   include Taxonomix
 
-  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname
+  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname, :organization_ids
 
   validates :name, :presence => true, :uniqueness => true,
     :format => { :with => /\A(\S+)\Z/, :message => N_("can't contain white spaces.") }
--- a/db/migrate/20141223142759_fill_discovery_attribute_sets_for_existing_hosts.rb
+++ b/db/migrate/20141223142759_fill_discovery_attribute_sets_for_existing_hosts.rb
@@ -1,6 +1,6 @@
 class FillDiscoveryAttributeSetsForExistingHosts < ActiveRecord::Migration
   def up
-    Host::Discovered.scoped.each { |host| host.populate_fields_from_facts}
+    Host::Discovered.all.each { |host| host.populate_fields_from_facts}
   end
 
   def down
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -39,7 +39,9 @@ module ForemanDiscovery
 
     # Add any db migrations
     initializer "foreman_discovery.load_app_instance_data" do |app|
-      app.config.paths['db/migrate'] += ForemanDiscovery::Engine.paths['db/migrate'].existent
+      ForemanDiscovery::Engine.paths['db/migrate'].existent.each do |path|
+        app.config.paths['db/migrate'] << path
+      end
     end
 
     initializer 'foreman_discovery.register_plugin', :after=> :finisher_hook do |app|
@@ -145,10 +147,6 @@ module ForemanDiscovery
         # Only available in 1.8, otherwise it has to be in the initializer below
         apipie_documented_controllers ["#{ForemanDiscovery::Engine.root}/app/controllers/api/v2/*.rb"]
       end
-    end
-
-    initializer "foreman_discovery.load_app_instance_data" do |app|
-      app.config.paths['db/migrate'] += ForemanDiscovery::Engine.paths['db/migrate'].existent
     end
 
     initializer "foreman_discovery.apipie" do
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,9 +2,9 @@
 Rails.application.routes.draw do
 
   # Needed to make the hosts/edit form render
-  match 'architecture_selected_discovered_hosts' => 'hosts#architecture_selected'
-  match 'os_selected_discovered_hosts'           => 'hosts#os_selected'
-  match 'medium_selected_discovered_hosts'       => 'hosts#medium_selected'
+  get 'architecture_selected_discovered_hosts' => 'hosts#architecture_selected'
+  get 'os_selected_discovered_hosts'           => 'hosts#os_selected'
+  get 'medium_selected_discovered_hosts'       => 'hosts#medium_selected'
 
   constraints(:id => /[^\/]+/) do
     resources :discovered_hosts, :except => [:new, :create] do


test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -29,6 +29,9 @@ def switch_controller(klass)
                        :name => 'discovery_reboot',
                        :value => true,
                        :category => 'Setting::Discovered')
+
+    #load the power.rb file, as PowerService is not autoloaded from file named power
+    ForemanDiscovery::NodeAPI::Power
```
ohadlevy: should we rename it then instead?

unorthodoxgeek: that file has a few classes defined in it, renaming the file isn't going to do much, but maybe we should split it to one class per file. don't think it's important enough, but if you'd like, I'll do it.

lzap:
Thanks, that is very appreciated! Assuming these changes are not compatible with Rails 3, we will hold off until Core is migrated.

foreman_discovery.gemspec
```diff
@@ -29,4 +29,5 @@ Gem::Specification.new do |s|
   s.summary = %q{MaaS Discovery Plugin for Foreman}
 
   s.add_dependency 'deface', '< 2.0'
+  s.add_dependency 'rails', '>= 4.1'
```
lzap: We already have transitive dependency Discovery -> Core -> Rails, so I'd rather not introduce this one.

app/models/discovery_rule.rb
```diff
@@ -5,6 +5,8 @@ class DiscoveryRule < ActiveRecord::Base
   include Parameterizable::ByIdName
   include Taxonomix
 
+  attr_accessible :name, :search, :hostgroup_id, :hostname, :priority, :organization_ids, :max_count
```
lzap: I'd also add `enabled`, no need to be private for this one. And also `location_ids` for integrity, although we only test against organizations in our unit tests.

db/migrate/20141223142759_fill_discovery_attribute_sets_for_existing_hosts.rb
```diff
@@ -1,6 +1,6 @@
 class FillDiscoveryAttributeSetsForExistingHosts < ActiveRecord::Migration
   def up
-    Host::Discovered.scoped.each { |host| host.populate_fields_from_facts}
+    Host::Discovered.all.each { |host| host.populate_fields_from_facts}
```
lzap, line 3: Hmmm if need to support both migrations (on Rails 3 -> Rails 3 and also Rails 3 -> Rails 4) then let's do this in a transparent way. Can you make an if-statement and test for Rails version here? What is the background for this change btw?

test/functional/api/v2/discovered_hosts_controller_test.rb
lzap: Please split it, good catch. My fault :-( Would you mind doing this in a separate patch? This might be problematic in the future and we might want to backport this.

Looks great, aside of Katello, this was the most complicated plugin! Well done!

lzap: I wonder if we can write a check script for this and run it in our unit tests (in core as well).

foreman_discovery.gemspec
unorthodoxgeek: this will allow versioning of the plugin in a way that bundler can decide which version of this to use. in case of older codebase, bundler will not use this plugin version because of the dependency. I think it should stay.

test/functional/api/v2/discovered_hosts_controller_test.rb
unorthodoxgeek: this isn't the most complicated plugin, try openSCAP :smile:

foreman_discovery.gemspec
lzap: > this will allow versioning of the plugin in a way that bundler can decide which version of this to use. in case of older codebase, bundler will not use this plugin version because of the dependency. I think it should stay.

It is unnecessary, our versions are fixed and defined in:

http://theforeman.org/plugins/foreman_discovery/4.1/

For the next release, I expect the only one combination there. Can you
please be more specific and provide an example for our case?

## 

Later,
 Lukas #lzap Zapletal

db/migrate/20141223142759_fill_discovery_attribute_sets_for_existing_hosts.rb
unorthodoxgeek, line 3: this version should work for both 3 and 4.
scoped is undefined in rails 4.

test/functional/api/v2/discovered_hosts_controller_test.rb
unorthodoxgeek: we can remove this line once https://github.com/theforeman/foreman_discovery/pull/229 is merged

unorthodoxgeek:
@lzap - fixed all of your comments except the one which I believe should stay that way, please re-review

lib/foreman_discovery/engine.rb
```diff
@@ -39,7 +39,9 @@ class Engine < ::Rails::Engine
 
     # Add any db migrations
     initializer "foreman_discovery.load_app_instance_data" do |app|
-      app.config.paths['db/migrate'] += ForemanDiscovery::Engine.paths['db/migrate'].existent
```
unorthodoxgeek, line 42: btw, why is this code duplicated in 151?
I just went with it because I didn't want to dive in, but it seems like this might be an error

lzap, line 42: Oh please remove the block, that's typo.

lzap:
Ok looks good, please rebase and I can test!

lib/foreman_discovery/engine.rb
unorthodoxgeek, line 42: removed the bottom block

unorthodoxgeek:
branch seems up-to-date with develop

app/models/discovery_rule.rb
```diff
@@ -5,7 +5,7 @@ class DiscoveryRule < ActiveRecord::Base
   include Parameterizable::ByIdName
   include Taxonomix
 
-  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname
+  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname, :organization_ids
```
domcleal, line 8: Taxonomix should already provide this.

orrabin:
Started to test this and the dashboard also needs to be changed from `scoped`:
https://github.com/theforeman/foreman_discovery/blob/develop/app/views/dashboard/_discovery_widget.html.erb#L1

orrabin:
@unorthodoxgeek I can remove `:organization_ids` and fix the dashboard on merge but I'm getting `Circular dependency detected while autoloading constant DiscoveredHostsController` in a few places.
I don't have a reproducer that always works yet.

ohadlevy:
required foreman version should be bumped to 1.11

ohadlevy:
when running tests they also fail:

```
seedsTest#test_0001_populates features:
NoMethodError: undefined method `find_or_create_by_name' for #<Class:0x000000050712a0>
foreman_discovery/db/seeds.d/50_discovery_templates.rb:1:in `<top (required)>'
    test/lib/tasks/seeds_test.rb:18:in `seed'
    test/lib/tasks/seeds_test.rb:27:in `block in <class:SeedsTest>'

```

orrabin:
replaced by https://github.com/theforeman/foreman_discovery/pull/238

PR #233, opened 2015-11-20
Fixes #12548 - Allow discovery_rule to be mass assigned

dLobatog:
As part of projects.theforeman.org/projects/foreman/issues/7568 , the
Host and Hostgroup models will become mass-assignment protected.
In that case, we have to whitelist the attributes that come from
discovery that are mass assigned. A few tests use it and will fail
otherwise.

### DIFF
--- a/app/models/hostgroup_extensions.rb
+++ b/app/models/hostgroup_extensions.rb
@@ -3,5 +3,6 @@ module HostgroupExtensions
 
   included do
     has_many :discovery_rules
+    attr_accessible :discovery_rules, :discovery_rule_ids, :discovery_rule_names, :type
   end
 end
--- a/app/models/concerns/discovery_subnet.rb
+++ b/app/models/concerns/discovery_subnet.rb
@@ -2,6 +2,8 @@ module DiscoverySubnet
   extend ActiveSupport::Concern
   included do
     belongs_to :discovery, :class_name => "SmartProxy"
+
+    attr_accessible :discovery_id
   end
 end
 
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -5,6 +5,8 @@ class DiscoveryRule < ActiveRecord::Base
   include Parameterizable::ByIdName
   include Taxonomix
 
+  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname
+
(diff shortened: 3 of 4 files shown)


app/models/hostgroup_extensions.rb
```diff
@@ -3,5 +3,6 @@ module HostgroupExtensions
 
   included do
     has_many :discovery_rules
+    attr_accessible :discovery_rules, :discovery_rule_ids, :discovery_rule_names
```
lzap: Can you add `:type` as well here to fix the tests? Thanks.

lzap:
For the record, this breaks Foreman. Extensions are included in host/hostgroup classes and I think they overwrite what they had. We need to append to existing?

lzap:
Oh wait it appends correctly, we just need to wait until core has accessible attributes (the other ones) because this enables the checking obviously.

dLobatog:
Added :type to hostgroup_extensions, this is waiting on the other PR then

lzap:
Ok re [test]

lzap:
Ok one more please:

``` diff
diff --git a/app/models/discovery_rule.rb b/app/models/discovery_rule.rb
index 8cdb3fd..e18c20e 100644
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -4,6 +4,8 @@ class DiscoveryRule < ActiveRecord::Base
   friendly_id :name
   include Taxonomix

+  attr_accessible :name, :priority, :search, :enabled, :hostgroup, :hostgroup_id, :max_count, :hostname

   validates :name, :presence => true, :uniqueness => true,
     :format => { :with => /\A(\S+)\Z/, :message => N_("can't contain white spaces.") }
;;;
```

lzap:
Oh we need this change too, would you mind? I can take from here if you prefer, but then review will be slower as you don't have commit rights :-)

https://github.com/theforeman/foreman_bootdisk/commit/793648ecfcd87f75d4f953914bfd2b03ee346f33

dLobatog:
@lzap Not anymore, protecting :locked was a mistake apparently, and should work now. [test] :smile:

lzap:
Oh was there a patch in core which fixed it? it's green, great!

lzap:
Merged as 75c6126058efd444b85061827ea2574c31d3c10d, thank you!

PR #236, opened 2015-12-01
adding the require accessible attribute for ProvisioningTemplate

unorthodoxgeek:


config/initializers/discovery_setup.rb
```diff
@@ -1 +1,3 @@
-
+class Subnet < ActiveRecord::Base
+  attr_accessible :discovery_id
+end
```
lzap, line 3: Oh cool thanks, instead of reopening the class, would you mind implementing the same approach as in https://github.com/theforeman/foreman_discovery/blob/develop/app/models/hostgroup_extensions.rb

(Include it in the `engine.rb`) Thanks!

unorthodoxgeek:
seems like this was solved on master, thanks @lzap

PR #238, opened 2015-12-28
Fixes #12951 - rails 4 changes

orrabin:
This replaces: https://github.com/theforeman/foreman_discovery/pull/228
The changed I added:
- removed `organization_ids` from `app/models/discovery_rule.rb`
- changed `app/views/dashboard/_discovery_widget.html.erb` to use `all` instead of `scoped`
- fixed both seeds files that used `find_or_create_by_name` to use `find_or_create_by(:name => )`
- bumped foreman version to  1.11

### DIFF
--- a/db/seeds.d/50_discovery_templates.rb
+++ b/db/seeds.d/50_discovery_templates.rb
@@ -1,9 +1,8 @@
-kind = TemplateKind.find_or_create_by_name('kexec')
+kind = TemplateKind.where(:name => 'kexec').first_or_create
 
 ProvisioningTemplate.without_auditing do
   content = File.read(File.join(ForemanDiscovery::Engine.root, 'app', 'views', 'foreman_discovery', 'redhat_kexec.erb'))
-  tmpl = ProvisioningTemplate.find_or_create_by_name(
-    :name => 'Discovery Red Hat kexec',
+  tmpl = ProvisioningTemplate.where(:name => 'Discovery Red Hat kexec').first_or_create(
     :template_kind_id => kind.id,
     :snippet => false,
     :template => content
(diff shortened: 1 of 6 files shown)


db/seeds.d/50_discovery_templates.rb
```diff
@@ -1,8 +1,8 @@
-kind = TemplateKind.find_or_create_by_name('kexec')
+kind = TemplateKind.find_or_create_by(:name =>'kexec')
```
dLobatog: Do you care about Rails 3 compatibility in this PR? If so, substitute this by `where(conditions).first_or_create` like we did in Foreman core https://github.com/theforeman/foreman/pull/2747/files

db/seeds.d/50_discovery_templates.rb
```diff
@@ -1,8 +1,8 @@
...
 
 ProvisioningTemplate.without_auditing do
   content = File.read(File.join(ForemanDiscovery::Engine.root, 'app', 'views', 'foreman_discovery', 'redhat_kexec.erb'))
-  tmpl = ProvisioningTemplate.find_or_create_by_name(
+  tmpl = ProvisioningTemplate.find_or_create_by(
```
dLobatog: This is not exactly the same, the original one was finding _by name_, and creating it if it doesn't exist. This one is finding _by all of these attributes_, then creating it if doesn't exist.

`.where(:name => 'Discovery Red Hat kexec').first_or_create(:template_kind_id => kind.id, etc...)` would be a more exact equivalent

ohadlevy:
@orrabin can you correct the commit message?

orrabin:
@ohadlevy fixed the commit message.

dLobatog:
http://ci.theforeman.org/job/test_plugin_pull_request/1943/ is testing this PR

ohadlevy:
merged as 74487f6 thanks @unorthodoxgeek and @orrabin

PR #239, opened 2016-01-03
Refs #12951 - Allowing mass assignment of discovery_rule_id

orrabin:
core tests were failing because of discovery_rule_id


ohadlevy:
merged as 1070ab8 thanks @orrabin

PR #242, opened 2016-01-26
Fixes #13390 - Update the zanata mapping file

bkearney:


lzap:
We can only merge once core is merged: https://github.com/theforeman/foreman/pull/3123

There will be some discussion going on I guess.

bkearney:
sounds good... could you set to woc

PR #246, opened 2016-02-04
Fixes #9490 - Added setting to specify fact to use for hostname

imriz:
Supersedes  #213 

Based on the original patch from @orrabin

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -41,8 +41,11 @@ class Host::Discovered < ::Host::Base
     # construct hostname
     prefix_from_settings = Setting[:discovery_prefix]
     hostname_prefix = prefix_from_settings if prefix_from_settings.present? && prefix_from_settings.match(/^[a-zA-Z].*/)
-    hostname_prefix ||= 'mac'
-    hostname = FacterUtils::bootif_mac(facts).try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
+
+    name_fact = return_first_valid_fact(Setting::Discovered.discovery_hostname_fact_array, facts)
+    raise(::Foreman::Exception.new(N_("Invalid facts: hash does not contain a valid value for any of the facts in the discovery_hostname setting: %s"), Setting::Discovered.discovery_hostname_fact_array.join(', '))) unless name_fact && name_fact.present?
+    hostname = normalize_string_for_hostname("#{hostname_prefix}#{name_fact}")
+    Rails.logger.warn "Hostname does not start with an alphabetical character" unless hostname.downcase.match /^[a-z]/
 
     # create new host record
     h = ::Host::Discovered.find_by_name hostname
@@ -184,4 +187,17 @@ class Host::Discovered < ::Host::Base
     "discovery-not-matched"
   end
 
+  def self.normalize_string_for_hostname(hostname)
+    hostname = hostname.to_s.downcase.gsub(/(^[^a-z0-9]*|[^a-z0-9\-]|[^a-z0-9]*$)/,'')
+    raise(::Foreman::Exception.new(N_("Invalid hostname: Could not normalize the hostname"))) unless hostname && hostname.present?
+    hostname
+  end
+
+  def self.return_first_valid_fact(facts_array, facts)
+    return facts[facts_array] if !facts_array.is_a?(Array)
+    facts_array.each do |value|
+      return facts[value] if !facts[value].nil?
+    end
+    return nil
+  end
 end
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -8,6 +8,7 @@ class Setting::Discovered < ::Setting
   BLANK_ATTRS << 'discovery_facts_hardware'
   BLANK_ATTRS << 'discovery_facts_network'
   BLANK_ATTRS << 'discovery_facts_ipmi'
+  BLANK_ATTRS << 'discovery_prefix'
 
   def self.load_defaults
     # Check the table exists
@@ -15,7 +16,8 @@ class Setting::Discovered < ::Setting
 
     Setting.transaction do
       [
-        self.set('discovery_fact', N_("Fact name to use for primary interface detection and hostname"), "discovery_bootif"),
+        self.set('discovery_fact', N_("Fact name to use for primary interface detection"), "discovery_bootif"),
+        self.set('discovery_hostname', N_("List of facts to use for the hostname (separated by comma, first wins)"), "discovery_bootif"),
         self.set('discovery_auto', N_("Automatically provision newly discovered hosts, according to the provisioning rules"), false),
         self.set('discovery_reboot', N_("Automatically reboot discovered host during provisioning"), true),
       ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
@@ -64,8 +66,17 @@ class Setting::Discovered < ::Setting
 
   def self.discovery_fact_column_array
     return [] if !Setting['discovery_fact_column'].present?
+    discovery_comma_to_array Setting['discovery_fact_column']
+  end
+
+  def self.discovery_hostname_fact_array
+    return [] if !Setting['discovery_hostname'].present?
+    discovery_comma_to_array Setting['discovery_hostname']
+  end
+
+  def self.discovery_comma_to_array string
     list = []
-    Setting['discovery_fact_column'].to_s.split(",").each do |value|
+    string.to_s.split(",").each do |value|
       list << value.strip
     end
   rescue => error
--- a/test/unit/host_discovered_test.rb
+++ b/test/unit/host_discovered_test.rb
@@ -7,6 +7,10 @@ class HostDiscoveredTest < ActiveSupport::TestCase
                        :value => 'discovery_bootif',
                        :category => 'Setting::Discovered')
     FactoryGirl.create(:setting,
+                       :name => 'discovery_hostname',
+                       :value => 'discovery_bootif',
+                       :category => 'Setting::Discovered')
+    FactoryGirl.create(:setting,
                        :name => 'discovery_prefix',
                        :value => 'mac',
                        :category => 'Setting::Discovered')
@@ -65,9 +69,19 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert host.refresh_facts
   end
 
+  test "should create discovered host with hostname if a fact was supplied" do
+    raw = parse_json_fixture('/facts.json')
+    Setting[:discovery_hostname] = 'somefact'
+    facts = raw['facts'].merge({"somefact" => "somename"})
+    host = Host::Discovered.import_host_and_facts(facts).first
+    assert_equal 'macsomename', host.name
+    refute_equal 'e4:1f:13:cc:36:5a', host.mac
+  end
+
   test "should create discovered host with fact_name as a name if it is a valid mac" do
     raw = parse_json_fixture('/facts.json')
     Setting[:discovery_fact] = 'somefact'
+    Setting[:discovery_hostname] = 'somefact'
     facts = raw['facts'].merge({"somefact" => "E4:1F:13:CC:36:5A"})
     host = Host::Discovered.import_host_and_facts(facts).first
     assert_equal 'mace41f13cc365a', host.name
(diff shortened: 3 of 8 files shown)


dLobatog:
@orrabin mind to review this one?

[test]

app/helpers/discovered_hosts_helper.rb
```diff
@@ -66,4 +66,15 @@ def discovery_status_icon(host)
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
   end
+
+  def self.normalize_string_for_hostname(hostname)
+    hostname.try(:downcase).try(:gsub,/(^[^a-z0-9]|[^a-z0-9-_]|[^a-z0-9]$)/,'')
```
unorthodoxgeek: hostname.to_s.downcase.gsub(...)

app/helpers/discovered_hosts_helper.rb
```diff
@@ -66,4 +66,15 @@ def discovery_status_icon(host)
...
+  end
+
+  def self.return_first_valid_fact(factsarray,facts)
```
unorthodoxgeek: please use snake_case for multi-word variable names.

unorthodoxgeek:
what does unstable mean? @dLobatog - maybe you know?

imriz:
@unorthodoxgeek PR updated per your comments. Thanks :)

app/helpers/discovered_hosts_helper.rb
```diff
@@ -66,4 +66,15 @@ def discovery_status_icon(host)
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
   end
+
+  def self.normalize_string_for_hostname(hostname)
+    hostname.to_s.try(:downcase).try(:gsub,/(^[^a-z0-9]|[^a-z0-9-_]|[^a-z0-9]$)/,'')
```
unorthodoxgeek: to_s ensures that you have a string. remove the try methods.

imriz:
Updated..

dLobatog:
@unorthodoxgeek Unstable means it's failing, but also that tests in this plugin have not been consistently green before.

unorthodoxgeek:
@imriz - all the tests fail on ActiveRecord::RecordInvalid: Validation failed: Name can't be blank, Name is invalid - should be pretty easy to fix, it's usually a single LOC

imriz:
@unorthodoxgeek I am just not sure how this error is related to the change in this PR.

app/models/host/discovered.rb
```diff
@@ -41,8 +41,10 @@ def self.import_host_and_facts facts
     # construct hostname
     prefix_from_settings = Setting[:discovery_prefix]
     hostname_prefix = prefix_from_settings if prefix_from_settings.present? && prefix_from_settings.match(/^[a-zA-Z].*/)
-    hostname_prefix ||= 'mac'
-    hostname = FacterUtils::bootif_mac(facts).try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
+
+    name_fact = DiscoveredHostsHelper.return_first_valid_fact(Setting::Discovered.discovery_hostname_fact_array,facts)
```
lzap: I think the `discovery_hostname_fact_array` can be hardcoded in the helper.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -66,4 +66,15 @@ def discovery_status_icon(host)
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
   end
+
+  def self.normalize_string_for_hostname(hostname)
+    hostname.to_s.downcase.gsub(/(^[^a-z0-9]|[^a-z0-9-_]|[^a-z0-9]$)/,'')
+  end
```
lzap: Hosts must not start with a number, can you make sure this method checks and fix this?

Also this helper has zero unit test coverage, can you add new unit test and cover both methods?

lzap:
Can you rebase and incorporate my comments please?

We are wrapping up 5.0 release this week.

app/helpers/discovered_hosts_helper.rb
imriz: That is not correct. That was correct in RFC 952, but the restriction was relaxed in RFC 1123 Section 2.1

app/models/host/discovered.rb
imriz: Not sure I am following. Hard code what?

imriz:
@lzap Rebased. Will update soon with the test.

lzap:
Thanks for heads-up. The code looks good, I will test and merge once the test is ready.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -81,4 +81,15 @@ def discovery_status_icon(host)
...
+
+  def self.return_first_valid_fact(facts_array,facts)
+    return if !facts_array.is_a?(Array)
+    facts_array.each do |value|
+      return facts[value] if !facts[value].nil?
+    end
+  end
```
imriz: @lzap looking at this again, I am not sure these methods belongs here. I am thinking of just moving them into the model, as class methods. What do you think?

orrabin:
@imriz tests are still failing with the error `Name can't be blank`. I looked at the test `can search discovered hosts by disks_size` in test/unit/discovery_attribute_set_test.rb and in `import_host_and_facts` got `nil` in hostname_prefix and `[]` in name_fact so hostname will be empty.

imriz:
@orrabin Thanks. I found and resolved it. I am left with the tests which checks the fallback of the prefix to "mac" - I think this is the wrong approach. I think we should allow users to have an empty prefix, and raise error on invalid values for the prefix. What do you think?

orrabin:
@imriz I think we need a prefix in case the name is somehow not resolved so there is never an option for an empty name but this could be a fallback in case of error or maybe we want an error in that case. So maybe the test should be divided to:
- check the setting prefix is really present in the name
- check the host gets a name with no prefix
- check for prefix 'mac' if something goes wrong? or check that if the name was not resolved correctly you get an error and then remove the 'mac' prefix.

@lzap - do we still need the mandatory prefix after https://github.com/theforeman/foreman_discovery/pull/246#discussion_r52906961 ?

imriz:
@orrabin yeah. I am already throwing an error if I can't resolve hostname. If the idea of the mandatory prefix was to make sure we have something for a hostname, we can have a fallback for that, instead of having a mandatory prefix.

orrabin:
@imriz if I'm not mistaken it was to make sure the hostname starts with a letter.

imriz:
@orrabin @lzap any decision on the mandatory prefix?

imriz:
PR updated

lzap:
I prefer to keep the prefix, but I don't think it must be mandatory, no. If user is absolutely sure the suffix will never start with number, then he/she can use empty prefix, so let's just add a test case to support that. But let's keep it for sure.

lzap:
Maybe add a extra check in the code base and issue a `logger.warn` if the result starts with number.

app/models/setting/discovered.rb
```diff
@@ -16,6 +16,7 @@ def self.load_defaults
     Setting.transaction do
       [
         self.set('discovery_fact', N_("Fact name to use for primary interface detection and hostname"), "discovery_bootif"),
+        self.set('discovery_hostname', N_("List of facts to use for the hostname (seperate by comma)"), "discovery_bootif"),
```
lzap: Please add `(separated by comma, first wins)`

app/helpers/discovered_hosts_helper.rb
```diff
@@ -81,4 +81,5 @@ def discovery_status_icon(host)
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
   end
+
```
lzap: Leftover?

app/models/setting/discovered.rb
lzap: If this can be blanked, add a test for this case. But I read this that this does not accept blank value.

LGTM

imriz: It doesn't accept empty value - empty value will trigger the exception at app/models/host/discovered.rb:46

lzap: > It doesn't accept empty value - empty value will trigger the exception at app/models/host/discovered.rb:46

Ok, if you want to allow blank _prefix_, then add it to the ALLOW_BLANK
constant. See Foreman Core for an example, remember to add a test for
that.

## 

Later,
 Lukas #lzap Zapletal

imriz: @lzap Done. Test is already there ("should create discovered host without prefix").

imriz:
PR updated.

app/models/host/discovered.rb
```diff
@@ -41,8 +41,12 @@ def self.import_host_and_facts facts
     # construct hostname
     prefix_from_settings = Setting[:discovery_prefix]
     hostname_prefix = prefix_from_settings if prefix_from_settings.present? && prefix_from_settings.match(/^[a-zA-Z].*/)
-    hostname_prefix ||= 'mac'
-    hostname = FacterUtils::bootif_mac(facts).try(:downcase).try(:gsub,/:/,'').try(:sub,/^/, hostname_prefix)
+
+    name_fact = return_first_valid_fact(Setting::Discovered.discovery_hostname_fact_array,facts)
```
lzap: Ok this is last round of comments. Missing space before `,`

app/models/host/discovered.rb
```diff
@@ -41,8 +41,12 @@ def self.import_host_and_facts facts
...
+    raise(::Foreman::Exception.new(N_("Invalid facts: hash does not contain the required fact '%s'"), name_fact)) unless name_fact && name_fact.present?
```
lzap: This should read: Cannot find any of the facts + Setting::Discovered.discovery_hostname_fact_array.join(', ') or similar to point the user on the correct setting.

app/models/host/discovered.rb
```diff
@@ -41,8 +41,12 @@ def self.import_host_and_facts facts
...
+    hostname = normalize_string_for_hostname("#{hostname_prefix}#{name_fact}")
+    raise(::Foreman::Exception.new(N_("Invalid hostname: Could not compute a valid hostname for the discovered host"))) unless hostname && hostname.present?
```
lzap: Move the exception to the normalize method.

app/models/host/discovered.rb
lzap: Also the i18n is incorrect, see http://projects.theforeman.org/projects/foreman/wiki/Translating

lzap: Oh scratch that, correct form!

app/models/host/discovered.rb
```diff
@@ -41,8 +41,12 @@ def self.import_host_and_facts facts
...
+    Rails.logger.warn "Hostname does not start with an alphabetical character. While this is allowed by the relevant RFC, this might create issues with certian software" unless hostname.downcase.match /^[a-z]/
```
lzap: Let's keep our log clean, just use "Hostname does not start with an alphabetical character" I guess and that's it :-)

app/models/setting/discovered.rb
```diff
@@ -16,6 +17,7 @@ def self.load_defaults
     Setting.transaction do
       [
         self.set('discovery_fact', N_("Fact name to use for primary interface detection and hostname"), "discovery_bootif"),
```
lzap: Remove "and hostname", won't be relevant for all cases now.

test/unit/host_discovered_test.rb
```diff
@@ -148,6 +168,18 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert Token.where(:host_id => h.id).empty?
   end
 
+  test "discovered host hostname normalization" do
+    assert_equal Host::Discovered.normalize_string_for_hostname("90:B1:1C:54:D5:82"),"90b11c54d582"
+    assert_equal Host::Discovered.normalize_string_for_hostname(".-_Test::Host.name_-."),"testhostname"
+    assert_equal Host::Discovered.normalize_string_for_hostname("testhostname"),"testhostname"
+  end
```
lzap, line 181: Please break up these into separate tests.

Also add test for nil (should return nil as well), an empty string and a string with dash character and underscore character. Both should be allowed.

app/models/host/discovered.rb
```diff
@@ -184,4 +188,15 @@ def lookup_value_match
     "discovery-not-matched"
   end
 
+  def self.normalize_string_for_hostname(hostname)
+    hostname.to_s.downcase.gsub(/(^[^a-z0-9]*|[^a-z0-9\-]|[^a-z0-9]*$)/,'')
+  end
+
+  def self.return_first_valid_fact(facts_array,facts)
```
lzap: Missing space again, I like spaces after commas :-) That's all!

test/unit/host_discovered_test.rb
imriz, line 181: Please note that underscores are not allowed in hostnames.
Also, how can the normalize_string_for_hostname return nil or empty string, if we're going to make it throw exception in such cases? I can however make a test that checks it does throw an exception in such cases.

app/models/host/discovered.rb
imriz: and you're right about it :)

imriz:
@lzap PR updated.

imriz:
I think the test failure is unrelated - it only happens with Ruby 2.0 with postgresql and the error seems to be unrelated:

```
Gemset 'test_plugin_pull_request-2' does not exist, 'rvm ruby-2.0.0-p353 do rvm gemset create test_plugin_pull_request-2' first, or append '--create'.
/usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/resolver.rb:347:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'rails (= 4.1.14.1) ruby' in any of the gem sources listed in your Gemfile or available on this machine. (Bundler::GemNotFound)
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/resolver.rb:325:in `each'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/resolver.rb:325:in `verify_gemfile_dependencies_are_found!'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/resolver.rb:198:in `start'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/resolver.rb:182:in `resolve'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/definition.rb:200:in `resolve'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/definition.rb:140:in `specs'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/definition.rb:185:in `specs_for'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/definition.rb:174:in `requested_specs'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/environment.rb:18:in `requested_specs'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/runtime.rb:13:in `setup'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler.rb:127:in `setup'
    from /usr/local/share/gems/gems/bundler-1.10.6/lib/bundler/setup.rb:18:in `<top (required)>'
    from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
A ruby must be selected in order to delete a gemset.
```

imriz:
[test]

lzap:
Closing eyes on those missing spaces in the tests :-) Thanks nice work.

lzap:
Merged as 1f4e985b05d4af9d05e361a0fa9bdcbea3bc8efc, thank you!

imriz:
@lzap dammit! :) I will open a PR for the spaces ;)

Thanks for the reviews - much appreciated!

PR #247, opened 2016-02-08
Fixes #13593 - Preload assets before sprockets

dLobatog:
Currently Foreman core has an initializer (assets_paths) that takes care
of finding the Discovery assets in the engine and load them before
sprockets is loaded. This is happening because sprockets tries to load
the assets before they are defined (:after => finisher_hook is too
    late).

Just changing it to :before finisher_hook is enough to make it load
assets before then and the deprecation error will go away.

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -44,7 +44,7 @@ module ForemanDiscovery
       end
     end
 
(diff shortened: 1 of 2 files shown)


lzap:
Merged as e4a44fc90ca3098ddcef44ce8bdf861c4ef7a648, thank you!

PR #249, opened 2016-02-17
Fixes #13752 - Display dropdown for discovery_taxonomy settings

dLobatog:
Similar to #13721 , discovery_location and discovery_taxonomy should
show all taxonomies available instead of being a text field.


lzap:
Merged as c244242273237c7e2039a7a16b74ed2549bf4841, thank you!

PR #253, opened 2016-02-24
Fixes #13715 - fixed multiple deletion of hosts

lzap:
What an association return is not an Array in Rails 4+. If we want do array
operation on it we should call to_a.


orrabin:
merged as 4a3065ab7c2111a80a3af004d4ad0e91726c326d, thanks @lzap!

PR #254, opened 2016-03-01
Fixes #13217 - rule ordering fixed

lzap:
Nasty bug we missed for some time. We actually process rules in _bad order_! :-(

### DIFF
--- a/test/unit/discovered_extensions_test.rb
+++ b/test/unit/discovered_extensions_test.rb
@@ -60,7 +60,7 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     host = Host::Discovered.import_host_and_facts(facts).first
     r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc",
                             :organizations => [host.organization], :locations => [host.location])
-    FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc",
+    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = x",
                        :organizations => [host.organization], :locations => [host.location])
     assert_equal find_discovery_rule(host), r1
   end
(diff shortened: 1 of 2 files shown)


lzap:
Bump @orrabin or @stbenjam this is blocker for 6.2.

test/unit/discovered_extensions_test.rb
```diff
@@ -70,11 +70,42 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     host = Host::Discovered.import_host_and_facts(facts).first
     FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = x",
                        :organizations => [host.organization], :locations => [host.location])
-    r2 = FactoryGirl.create(:discovery_rule, :priority => 2, :search => "facts.somefact = abc",
+    r2 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc",
                             :organizations => [host.organization], :locations => [host.location])
     assert_equal find_discovery_rule(host), r2
```
orrabin, line 75: Edit: I know understand this, only the second rule fits the host.  So why did you change these 2 tests? it doesn't matter what the priority of the other rule is.

test/unit/discovered_extensions_test.rb
```diff
@@ -70,11 +70,42 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
...
   end
 
+  test "first rule with same priority is found for a discovered host" do
```
orrabin: This has different priorities.

orrabin:
@lzap there are 2 tests with different priorities that say `same priority` so I don't understand how to test this
. From the code change it looks like it picks by priority and then the oldest one if there is a situation with 2 rules with the same priority.
So how can there be a situation where the second rule is chosen if it has the same priority and it was created 2 seconds later?

test/unit/discovered_extensions_test.rb
lzap, line 75: I need to test cases with same priorities. Now we have two tests covering both 1st and 2nd rule. I reused existing test because it was testing wrong case - fact did not match therefore it was a simple case that is already covered above.

test/unit/discovered_extensions_test.rb
lzap: Fixed.

lzap:
I reworded all tests a bit, hope it's clear now.

lzap:
Ok so let me paste the block we are talking about (I hope :-)

``` ruby
  test "first rule out of two is found for a discovered host" do
    r1 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc",
                            :organizations => [host.organization], :locations => [host.location])
    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = x",
                       :organizations => [host.organization], :locations => [host.location])
    assert_equal find_discovery_rule(host), r1
  end

  test "second rule out of two is found for a discovered host" do
    FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = x",
                       :organizations => [host.organization], :locations => [host.location])
    r2 = FactoryGirl.create(:discovery_rule, :priority => 1, :search => "facts.somefact = abc",
                            :organizations => [host.organization], :locations => [host.location])
    assert_equal find_discovery_rule(host), r2
  end
```

These tests (I touched to improve them a bit - they are unrelated to priorities) are here to test case when there are two (different!) rules and second is picked up. Just in case we have a bug that always matches the condition (therefore only 1st is always found).

orrabin:
@lzap lgtm, test failure is unrelated.

orrabin:
merged as b1d5689467aa42c699c5e3a4904a2d278dbf8a31, thanks @lzap!

PR #256, opened 2016-03-02
Fixes #14010 - network facts are properly displayed

lzap:
Some network facts (those ending with _eth0 or similar) are ignored and not
shown on the discovered host page, for example auto_negotiation_XXX. The
current code skips all facts because it present them in the Highlight section,
while we should still show them.

Also there is a pending PR for the image to add LLDAP facts like
lldp_neighbor_portid_XXX which are also completely ignored.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -38,7 +38,7 @@ class DiscoveredHostsController < ::ApplicationController
     get_interfaces
     @host.facts_hash.each do |key, value|
       value = number_to_human_size(value) if /size$/.match(key)
-      assign_fact_to_category(key, value) unless @interfaces.any? {|interface| key.include? interface[:identifier]}
+      assign_fact_to_category(key, value)
     end
     add_custom_facts
   end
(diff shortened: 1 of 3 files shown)


app/controllers/discovered_hosts_controller.rb
```diff
@@ -196,7 +196,11 @@ def init_regex_and_categories
     @categories_names = [:Hightlights, :Storage, :Hardware, :Network, :Software, :IPMI, :Misceleneous]
   end
 
-  def assign_fact_to_category(key, value )
+  def assign_fact_to_category(key, value)
+    if @interfaces.any? {|interface| key.include? interface[:identifier]}
```
dLobatog, line 200: This looks a bit fragile, is there any test for this?

lzap, line 200: That's right, this should be in a class, with unit testing. Created http://projects.theforeman.org/issues/14229

For now, I have added functional test that at least excercise the page if it renders properly. We haven't had one. Rebased.

dLobatog:
Merged as ea2c0f188806f75f55f05b0074f10c13357b62a9, thanks @lzap!

PR #257, opened 2016-03-03
Fixes #13914 - host and fact importing refactored

lzap:
This was refactored in core: https://github.com/theforeman/foreman/pull/3242

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -30,7 +30,7 @@ class Host::Discovered < ::Host::Base
     where(taxonomy_conditions).order("hosts.created_at DESC")
   }
 
-  def self.import_host_and_facts facts
+  def self.import_host facts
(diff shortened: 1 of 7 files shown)


app/models/host/discovered.rb
```diff
@@ -144,7 +144,8 @@ def proxy_url
 
   def refresh_facts
     facts = ::ForemanDiscovery::NodeAPI::Inventory.new(:url => proxy_url).facter
-    self.class.import_host_and_facts facts
+    self.class.import_host facts
+    import_facts facts
```
dLobatog, line 148: Aren't facts imported in line 57 (`host.import_facts(facts)`)?

lzap, line 148: This is explicit call when user clicks on `Refresh facts` button.

lzap:
Ping?

dLobatog:
Merged as 57cd4c76414a6f1dc830a5d2283b0b24a9690319, thanks @lzap!

PR #258, opened 2016-03-09
Fixes #14208 - Allows to list discovered hosts facts from the Api

orrabin:


dLobatog:
@orrabin :+1: works well, can you add a couple of tests for this endpoint? Aside from that, the ticket number is wrong, it shouldn't reference a Hammer CLI ticket. Thanks!

orrabin:
@dLobatog I added a test and opened up a new ticket.

test/functional/api/v2/fact_value_extensions_test.rb
```diff
@@ -0,0 +1,43 @@
...
+                           :value    => 'discovery_bootif',
+                           :category => 'Setting::Discovered')
+        FactoryGirl.create(:setting,
+                           :name     => 'discovery_prefix',
+                           :value    => 'mac',
+                           :category => 'Setting::Discovered')
+      end
+
+      def setup_set_script_name
+        @controller                 = Api::V2::FactValuesController.new
```
dLobatog: I think placing `tests "api/v2/fact_values"` or `tests Api::V2::FactValuesController`below class (line 6) works too, no need to override this method

test/functional/api/v2/fact_value_extensions_test.rb
```diff
@@ -0,0 +1,43 @@
...
+        @request.env["SCRIPT_NAME"] = @controller.config.relative_url_root
+      end
+
+      def test_list_hosts_facts
```
dLobatog: Please use `test 'title' do` syntax

test/functional/api/v2/fact_value_extensions_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+module Api
+  module V2
+    class FactValuesControllerExtensionsTest < ActionController::TestCase
+      setup do
+        User.current = User.find_by_login "admin"
```
dLobatog: `users(:admin)` to access the fixture directly

test/functional/api/v2/fact_value_extensions_test.rb
```diff
@@ -0,0 +1,43 @@
...
+        @facts       = {
+            "interfaces"        => "lo,eth0",
+            "ipaddress"         => "192.168.100.42",
+            "ipaddress_eth0"    => "192.168.100.42",
+            "macaddress_eth0"   => "AA:BB:CC:DD:EE:FF",
+            "discovery_bootif"  => "AA:BB:CC:DD:EE:FF",
+            "memorysize_mb"     => "42000.42",
+            "discovery_version" => "3.0.0",
+        }
+        FactoryGirl.create(:setting,
```
dLobatog, line 19: What do these settings have to do with the test?

test/functional/api/v2/fact_value_extensions_test.rb
orrabin: thanks! I was looking for that, I'll change it.

test/functional/api/v2/fact_value_extensions_test.rb
orrabin: I thought we were avoiding fixtures in new tests but I'll change it.

test/functional/api/v2/fact_value_extensions_test.rb
orrabin, line 19: Since I'm using `import_host_and_facts` it needs the setting for hostname or it will return the error:

```
Foreman::Exception: ERF42-9632 [Foreman::Exception]: Invalid facts: hash does not contain a valid value for any of the facts in the discovery_hostname setting:
```

I'll remove the prefix setting but hostname needs to stay.

lzap:
Merged as 0b12a494c7127bb3ff98d8161ce36596e9e7dabb, thank you!

PR #260, opened 2016-03-10
Fixes #14099, #14140 - removed the use of eval, typos

alongoldboim:


lzap:
Ok couple of issues:

You need to convert `@category_names` to string and use N_("category1") to get extractions. Read http://projects.theforeman.org/projects/foreman/wiki/Translating for more info.

Both setting names are also incorrect, you need to create a migration to fix this. See `20150114141626_actually_rename_smart_proxy_auth_related_settings.rb` as an example. We want to keep the actual value, thus the migration.

Remove the `evals`, let's do this in one PR, I will test it all. Please squash the commit message to

```
Fixes #14099, #14140 - category settings no longer need quoting, typos
```

Thanks!

lzap:
One additional issue. When rendering the category name, you need to use gettext without `N`:

```
  <div class="panel-heading" ><strong> <%= _(@categories_names[0]) %></strong> </div>
```

app/controllers/discovered_hosts_controller.rb
```diff
@@ -184,16 +184,22 @@ def auto_provision_all
 
   private
 
+  def remove_slashes(string)
+    string.slice!(0)
+    string.slice!(string.length-1)
+    string
+  end
```
lzap: What are you doing here? Setting will work as-is, we don't need to remove first and last character.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -184,16 +184,22 @@ def auto_provision_all
...
+    storage = Setting[:discovery_facts_storage].empty? ? /^blockdevice/ : Regexp.new(remove_slashes(Setting[:discovery_facts_storage]))
+    hardware = Setting[:discovery_facts_hardware].empty? ? /^(hardw|manufacturer|memo|process)/ : Regexp.new(remove_slashes(Setting[:discovery_facts__hardware]))
+    network = Setting[:discovery_facts_network].empty? ? /^(ipaddress|interfaces|dhcp|fqdn|hostname|link|mtu|net|macaddress|wol|port|speed)/ : Regexp.new(remove_slashes(Setting[:discovery_facts_network]))
+    software = Setting[:discovery_facts_software].empty? ? /^(bios|os|discovery)/ : Regexp.new(remove_slashes(Setting[:discovery_facts_software]))
+    ipmi = Setting[:discovery_facts_ipmi].empty? ? /^ipmi/ : Regexp.new(remove_slashes(Setting[:discovery_facts_ipmi]))
     @regex_array = [hightlights, storage, hardware, network, software, ipmi, false]
     @categories = Array.new(7) { Hash.new }
-    @categories_names = [:Hightlights, :Storage, :Hardware, :Network, :Software, :IPMI, :Misceleneous]
+    @categories_names = [ N_("Highlights"),  N_("Storage"),  N_("Hardware"),  N_("Network"),  N_("Software"),  N_("IPMI"),  N_("Miscellaneous")]
```
lzap: This has double spaces.

app/controllers/discovered_hosts_controller.rb
lzap: Ok don't remove slashes, the idea is to accept plain regexp. For instance, when doing

```
egrep ^Line
```

you also don't provide these.

lzap:
Merged as 7b8ac717e4f244c285d276c3de0c695621236179, thank you!

PR #261, opened 2016-03-16
Fixes #14227 - fixed reports for hosts without model_name

lzap:


app/views/discovered_mailer/_discovered_host.html.erb
```diff
@@ -2,7 +2,7 @@
 <td style="<%= td_style%>">
   <%= link_to host, discovered_host_path(:id => host, :host => @url.host, :port => @url.port, :only_path => false, :protocol => @url.scheme) %>
 </td>
-<td style="<%= td_style%>"><%= host.model_name %></td>
+<td style="<%= td_style%>"><%= host.try(:model_name) || 'N/A' %></td>
```
domcleal: You should be using .hardware_model_name as it was renamed in Foreman core, otherwise this will always result in 'N/A'.  See http://projects.theforeman.org/issues/13440 and https://groups.google.com/d/msg/foreman-dev/CLhSVaZsi1g/p9yMwJhMHAAJ.

lzap: > You should be using .hardware_model_name as it was renamed in Foreman core, otherwise this will always result in 'N/A'.  See http://projects.theforeman.org/issues/13440 and https://groups.google.com/d/msg/foreman-dev/CLhSVaZsi1g/p9yMwJhMHAAJ.

I want to keep the fix and possible refactoring separate for easier
cherry picking. I already filed an issue for the refactoring task.
Thanks.

## 

Later,
 Lukas #lzap Zapletal

dLobatog:
@lzap I agree with @domcleal 's comment above, not sure how is this fixing much so far if it'll always show 'n/a'. There a few other instances of `model_name` scattered across the plugin that should be changed to work with newer Foreman versions

lzap:
Renamed, I also removed the method from `Host::Discovered` as I don't see why it was there. Tests will fail - we are working on that.

lzap:

This is the last push towards Discovery 6.0 release (http://projects.theforeman.org/versions/100) - please help to review and merge @dLobatog thanks!

dLobatog:
[test]

app/models/host/discovered.rb
```diff
@@ -192,10 +192,6 @@ def kexec json
     raise ::Foreman::WrappedException.new(e, N_("Unable to perform kexec on %{name} via %{url}: %{msg}"), :name => name, :url => proxy_url, :msg => e.to_s)
   end
 
-  def self.model_name
-    ActiveModel::Name.new(Host)
-  end
-
```
lzap, line 198: I accidentally removed this model method. Fixed tests.

dLobatog:
Merged as 790b2acb66ee61b13a86e546aa29c8e5ced96a18, thanks @lzap!

PR #262, opened 2016-03-21
Fixes #9784 - fixed inherited parameters

lzap:
Discovery process had issues with inheritable attributes (Puppet Master/CA,
Realm and few others). It was simply ignored and there was no way of setting
them.

For interactive provisioning, we need to make sure that the new
`apply_inherited_attributes` is called. Also, since discovery uses Edit Host
form and hostgroup is not yet set, all the Inherit buttons are not rendered as
pressed, which is bad UX. As part of this patch, a parameter "inherit_all" is
passed into views to override the inherit button state.

In case of auto-provisioning, inherited attributes were completely ignored.
During the process, all the inheritable attributes are now explicitly set from
the matched hostgroup.

### DIFF
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -54,6 +54,8 @@ module Foreman::Controller::DiscoveredExtensions
     host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
     # fallback to the original if template did not expand
     host.name = original_name if host.name.empty?
+    # explicitly set all inheritable attributes from hostgroup
+    host.attributes = host.apply_inherited_attributes(hostgroup_id: rule.hostgroup_id)
     # save! does not work here
     host.save
   end
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -9,6 +9,10 @@ class ForemanDiscovery::HostConverter
     if set_managed
       host.managed = set_managed
       host.primary_interface.managed = set_managed
+      # render all inherit buttons as selected on Edit Host form
+      #host.define_singleton_method(:force_inherited) do |_|
+        #true
+      #end
     end
     # set build only and only on final save (facts are deleted)
     if set_build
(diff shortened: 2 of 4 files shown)


app/services/foreman_discovery/host_converter.rb
```diff
@@ -9,6 +9,10 @@ def self.to_managed(original_host, set_managed = true, set_build = true)
     if set_managed
       host.managed = set_managed
       host.primary_interface.managed = set_managed
+      # render all inherit buttons as selected on Edit Host form
```
jmontleon, line 12: Is this all supposed to be commented out?

lzap, line 12: Yes this is a leftover from my 1st attempt to fix this. It used to be a method.

lzap, line 12: Why you ask by the way? Depends on this in FUSOR?

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -54,6 +54,12 @@ def perform_auto_provision original_host, rule
     host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
     # fallback to the original if template did not expand
     host.name = original_name if host.name.empty?
+    # explicitly set all inheritable attributes from hostgroup
+    hostgroup = Hostgroup.find(rule.hostgroup_id)
+    host.hostgroup_inherited_attributes.each do |attribute|
```
ShimShtein: You can take the leverage of `#apply_inherited_attributes` here:

``` Ruby
host.attributes = host.apply_inherited_attributes(hostgroup_id: rule.hostgroup_id)
host.save
```

lzap: Changed the code to it:

``` ruby
    host.hostgroup_inherited_attributes.each do |attribute|
      value = hostgroup.send(attribute.to_sym)
      host.send("#{attribute}=".to_sym, value) if value
    end
```

lzap:
Thanks @ShimShtein would you mind testing this?

lzap:
This was fixed with https://github.com/theforeman/foreman_discovery/pull/273 closing.

PR #265, opened 2016-03-30
Fixes #14288 - Lock discovered host into discovery via PXE configuration

imriz:


app/models/host/discovered.rb
```diff
@@ -55,6 +55,18 @@ def self.import_host_and_facts facts
     # and save (interfaces are created via puppet parser extension)
     host.save(:validate => false) if host.new_record?
     state = host.import_facts(facts)
+
+    # lock the host into discovery via PXE, if feature is enabled in settings
+    if Foreman::Cast.to_bool(Setting['discovery_lock']) and !Setting['discovery_lock_template'].blank?
```
orrabin: @imriz I see the discussion on the issue suggested a common snippet to use in both templates, maybe the `discovery_lock_template` should have a default with that snippet?

app/models/host/discovered.rb
```diff
@@ -55,6 +55,18 @@ def self.import_host_and_facts facts
...
+      begin
+        @host = ::ForemanDiscovery::HostConverter.to_managed(host)
+        template = @host.unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
+        Subnet.subnet_for(host.ip).tftp_proxy.set host.mac, :pxeconfig => template
+      rescue Exception => e
+        Rails.logger.error "Could not set tftp_proxy from Proxy: #{e}"
+      end
+    end
```
lzap: Ok looks good, since you use `unattended_renderer`, you can also render a snippet. Can you make a patch to https://github.com/theforeman/community-templates pulling out this block https://github.com/theforeman/community-templates/blob/develop/pxe/PXELinux_default.erb#L20-L24 into separate snippet, which you can use here?

app/models/setting/discovered.rb
```diff
@@ -19,12 +19,19 @@ def self.load_defaults
...
         self.set('discovery_hostname', N_("List of facts to use for the hostname (separated by comma, first wins)"), "discovery_bootif"),
         self.set('discovery_auto', N_("Automatically provision newly discovered hosts, according to the provisioning rules"), false),
+        self.set('discovery_lock', N_("Automatically generate PXE configuration to pin a newly discovered host to discovery"), false),
         self.set('discovery_reboot', N_("Automatically reboot discovered host during provisioning"), true),
       ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
     end
 
     Setting.transaction do
       [
+          self.set('discovery_lock_template', N_("PXE template to be used when pinning a host to discovery"), ''),
```
lzap: Give the snippet a reasonable name (e.g. "pxelinux_discovery") and use it as the default value here.

app/models/setting/discovered.rb
```diff
@@ -19,12 +19,19 @@ def self.load_defaults
...
+      ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
+    end
```
lzap: Why you used a separate transaction?

app/models/host/discovered.rb
imriz: I was wondering if we shouldn't load it via seeds, instead of relying on external resource?
I can still open the PR to the community templates, to use the snippet I will add in the seeds.
What do you think?

app/models/setting/discovered.rb
imriz: No good reason - left over. I will remove it.

app/models/host/discovered.rb
orrabin: @lzap  there is a bug open about snippets in PXELinux global default template: http://projects.theforeman.org/issues/10592. If it still exists it might need fixing before the snippet can be added.

lzap: @imriz community-templates are the correct place to put it there, we pull them into the release as seedscript regularly.

lzap: Oh yeah, I haven't noticed snippets do not work in PXELinux global default. We need to fix this. Is this feasible @imriz ? I could perhaps take a look sometimes next week.

imriz: @lzap NP - I will take this, but this just might take me another week (end of sprint is coming, and still have open commitments :( ).

imriz: @lzap @orrabin the bug is not reproducible.
Also, I've opened the PR for the community templates.

imriz:
@lzap PR to the community templates has been merged, and comments from the review has been implemented.

app/models/host/discovered.rb
```diff
@@ -55,6 +55,18 @@ def self.import_host_and_facts facts
     # and save (interfaces are created via puppet parser extension)
     host.save(:validate => false) if host.new_record?
     state = host.import_facts(facts)
+
+    # lock the host into discovery via PXE, if feature is enabled in settings
+    if Foreman::Cast.to_bool(Setting['discovery_lock']) and !Setting['discovery_lock_template'].blank?
+      begin
+        @host = ::ForemanDiscovery::HostConverter.to_managed(host)
+        template = @host.unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
```
lzap: On systems which haven't synced new template yet (e.g. unreleased Foreman version - develop) this line

```
::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template'])
```

will render to `nil` which will likely cause an exception. Can you add an extra check on line 60 to only attempt to lock when the template is actually present?

lzap:
Can you move the whole block to the end of the `populate_fields_from_facts` and re-use discovered `host.subnet` which is associated already to the host? Because `subnet_for` is not guaranteed to be successful call and our code already handles all special cases.

app/models/host/discovered.rb
```diff
@@ -55,6 +55,18 @@ def self.import_host_and_facts facts
...
+        Subnet.subnet_for(host.ip).tftp_proxy.set host.mac, :pxeconfig => template
```
lzap: And finally, you should test if the tftp_proxy has TFTP feature associated, if not, skip this orchestration because it would fail.

app/models/host/discovered.rb
imriz: OK, just note that this feature is disabled by default

imriz:
@lzap PR updated per comments

lzap:
I still dont see the block moved down to `populate_fields_from_facts`, also getting N+1 which could be I think elmiminated by just moving it down:

```
2016-04-12T11:28:00 [app] [W] user: lzap
 | http://localhost:3000/api/v2/discovered_hosts/facts
 | N+1 Query detected
 |   Subnet => [:tftp]
 |   Add to your finder: :includes => [:tftp]
 | N+1 Query method call stack
 |   /home/lzap/work/foreman/app/models/subnet.rb:132:in `tftp?'
 |   /home/lzap/work/foreman_discovery/app/models/host/discovered.rb:60:in `import_host_and_facts'
 |   /home/lzap/work/foreman_discovery/app/controllers/api/v2/discovered_hosts_controller.rb:97:in `facts'
 |   /home/lzap/work/foreman/app/controllers/api/v2/base_controller.rb:152:in `disable_json_root'
 |   /home/lzap/work/foreman/app/controllers/concerns/application_shared.rb:13:in `set_timezone'
 |   /home/lzap/work/foreman/app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 |   /home/lzap/work/foreman/lib/middleware/catch_json_parse_errors.rb:9:in `call'
```

imriz:
@lzap this was more tricky than I initially thought.
I still think that it might be a good (read: safer) idea to do this in `import_host_and_facts`, but it seems to be working now anyway.
I ended up creating a deep clone with Marshal, and everything seems to be working fine.
I also added a test.
Can you please re-review?

lzap:
That is hacky. Ok do this instead:

``` diff
diff --git a/app/models/host/discovered.rb b/app/models/host/discovered.rb
index 7285a6c..1df13c9 100644
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -1,5 +1,6 @@
+  include Foreman::Renderer

   attr_accessible :discovery_rule_id

@@ -111,11 +112,11 @@ class Host::Discovered < ::Host::Base
     # lock the host into discovery via PXE, if feature is enabled in settings
     if Setting::Discovered.discovery_lock? and self.subnet.tftp?
       begin
-        @host = ::ForemanDiscovery::HostConverter.to_managed(Marshal::load(Marshal.dump(self)))
-        template = @host.unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
-        @host.subnet.tftp_proxy.set @host.mac, :pxeconfig => template
+        Rails.logger.info "Locking discovered host #{self.mac} in subnet #{subnet}"
+        template = unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
+        subnet.tftp_proxy.set mac, :pxeconfig => template
-        Rails.logger.error "Could not set tftp_proxy from Proxy: #{e}"
+        ::Foreman::Logging.exception("Could not set tftp_proxy from proxy", e)
       end
     end
     self.save!
```

imriz:
OK. This is more or less how we do it in our own fork today.
but maybe we should still set `@host = self` so people could refer it in the ERB templates, if needed?

imriz:
@lzap any chance you can re-review?

lzap:
On my TODO, sorry for the delay, got SELinux issue I need to solve which blots all Foreman versions.

lzap:
Yeah, sure `@host = self` is perfectly fine. Go for it, then I will test.

app/models/host/discovered.rb
```diff
@@ -108,6 +109,16 @@ def populate_fields_from_facts(facts = self.facts_hash, ignored_type)
     end
     self.discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => id).first_or_create
     self.discovery_attribute_set.update_attributes(import_from_facts)
+    # lock the host into discovery via PXE, if feature is enabled in settings
+    if Setting::Discovered.discovery_lock? and self.subnet.tftp?
```
lzap: Use `&&` instead `and` here.

app/models/host/discovered.rb
```diff
@@ -108,6 +109,16 @@ def populate_fields_from_facts(facts = self.facts_hash, ignored_type)
...
+      begin
+        Rails.logger.info "Locking discovered host #{self.mac} in subnet #{subnet}"
+        template = unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
```
lzap, line 117: No need to explicitly call `template` method, the renderer will call it for you I think.

app/models/setting/discovered.rb
```diff
@@ -72,6 +74,10 @@ def self.discovery_hostname_fact_array
     from_array Setting['discovery_hostname']
   end
 
+  def self.discovery_lock?
+    Foreman::Cast.to_bool(Setting['discovery_lock']) and !Setting['discovery_lock_template'].blank? and ProvisioningTemplate.exists?(:name => Setting['discovery_lock_template'])
```
lzap: Hmm isn't the setting automatically created with boolean type so no casting is necessary? You may created it with incorrect type (e.g. string) so it does not work for you. Try to delete it and let it recreate - it should automatically create a boolean type from the default value I think. If not, keep it as is.

lzap: Use `&&` everywhere please.

lzap:
The code looks good! Amend the host variable and we are good to go.

app/models/host/discovered.rb
imriz, line 117: Yes, you're right:
`content = template.respond_to?(:template) ? template.template : template` - but isn't explicit better than implicit? ;)

app/models/setting/discovered.rb
imriz: > Hmm isn't the setting automatically created with boolean type so no casting is necessary? You may created it with incorrect type (e.g. string) so it does not work for you. Try to delete it and let it recreate - it should automatically create a boolean type from the default value I think. If not, keep it as is.

I did it as a safety measure. If you think it is not needed, I will remove it.

imriz:
@lzap I've updated the PR based on your review. I did not remove the call to `.template` and the call to `Foreman::Cast.to_bool` - I believe explicit is better than implicit, but if you still feel strongly about it, I will remove it ;)

lzap:
A lot of changes in core recently, fixed tests: https://github.com/theforeman/foreman_discovery/pull/269 Then I will finally test, thanks for patience :-)

lzap:
Ok [test] now!

lzap:
Ok after second fix let's see .... randomized [test] are evill ;-)

lzap:
Tests are still broken, in the meantime, can you add setting description to the newly added settings? It's a last parameter, we don't use it in the plugin yet, but I plan to do it now: Example (add the last param):

```
self.set('discovery_prefix', N_("The default prefix to use for the host name, must start with a letter"), "mac", N_("Hostname prefix")),
```

lzap:
After https://github.com/theforeman/foreman_discovery/pull/271 is merged we are back in shape!

lzap:
Sorry for the delay, develop branch is back in shape, can you rebase please?

imriz:
@lzap Done (+ the settings description) :)

lzap:
Omg again [test]

app/models/setting/discovered.rb
```diff
@@ -28,6 +28,8 @@ def self.load_defaults
         self.set('discovery_facts_hardware', N_("Regex to organize facts for hardware section"), "", N_("Hardware facts")),
         self.set('discovery_facts_network', N_("Regex to organize facts for network section"), "", N_("Network facts")),
         self.set('discovery_facts_ipmi', N_("Regex to organize facts for ipmi section"), "", N_("IPMI facts")),
+        self.set('discovery_lock', N_("Automatically generate PXE configuration to pin a newly discovered host to discovery"), false,N_("Lock PXE")),
+        self.set('discovery_lock_template', N_("PXE template to be used when pinning a host to discovery"), 'pxelinux_discovery',N_("PXE lock template name")),
```
lzap: Oh sorry, folks have changed the API before they merged it. It's now a Proc, can you change it to (an example):

```
... { :collection => Proc.new {Hash[Environment.all.map{|env| [env[:name], env[:name]]}]} }
```

There is a missing space after comma above.

I also dont like the setting name, I suggest "Locked template name" or something similar?

imriz: @lzap not sure I am following you on the Proc - what changed?

lzap: > @lzap not sure I am following you on the Proc - what changed?

Please change the syntax as in the example provided, for performance
reasons (perhaps) the API expects a proc instead of hash.

## 

Later,
 Lukas #lzap Zapletal

imriz:
@lzap Updated.

lzap:
Merged as b7bf3f7f11ef03596c3e451792b2ad3ff593b874, thank you!

PR #266, opened 2016-04-07
Fixes #14527 - buttons are not visible to reader role

lzap:


dLobatog:
Works well here, I'm waiting to merge until I understand https://github.com/theforeman/foreman/issues/3403

app/helpers/discovered_hosts_helper.rb
```diff
@@ -12,27 +12,22 @@ def disc_report_column(record)
...
-        ),
+        select_action_button( _("Select Action"), {}, actions.map { |action| link_to_if_authorized(action[0] , action[1], action[2] || {}) }.flatten ),
       button_group(
         link_to(_("Expand All"),"#",:class => "btn btn-default",:onclick => "$('.glyphicon-plus-sign').toggleClass('glyphicon glyphicon-minus-sign glyphicon glyphicon-plus-sign');
                 $('.facts-panel').addClass('collapse in').height('auto');"
         )
       ),
       button_group(
-        link_to(_("Delete"), hash_for_discovered_host_path(:id => host),
+        link_to_if_authorized(_("Delete"), hash_for_discovered_host_path(:id => host).merge(:auth_object => host, :permission => :destroy_discovered_hosts),
```
tbrisker: why not use `display_delete_if_authorized` here?

app/helpers/discovered_hosts_helper.rb
```diff
@@ -12,27 +12,22 @@ def disc_report_column(record)
...
         button_group(
             link_to(_("Back"), :back, :class => "btn btn-default")
         ),
-        select_action_button( _("Select Action"), {},
-                              actions.map do |action|
-                                method = action[2] if action.size > 1
-                                link_to(action[0] , action[1], method)
-                              end.flatten
-        ),
+        select_action_button( _("Select Action"), {}, actions.map { |action| link_to_if_authorized(action[0] , action[1], action[2] || {}) }.flatten ),
```
tbrisker: why not `display_link_if_authorized`?

app/helpers/discovered_hosts_helper.rb
```diff
@@ -12,27 +12,22 @@ def disc_report_column(record)
...
 
   def discovered_hosts_title_actions(host)
-    actions = [[_('Provision'),  hash_for_edit_discovered_host_path(:id => host)]]
-    actions <<  [_('Auto Provision'), hash_for_auto_provision_discovered_host_path(:id => host), :method => :post]
-    actions <<  [_('Refresh facts') ,hash_for_refresh_facts_discovered_host_path(:id => host)]
-    actions <<  [_('Reboot') ,hash_for_reboot_discovered_host_path(:id => host), :method => :put]
+    actions = [[_('Provision'), hash_for_edit_discovered_host_path(:id => host).merge(:auth_object => host, :permission => :provision_discovered_hosts)]]
+    actions <<  [_('Auto Provision'), hash_for_auto_provision_discovered_host_path(:id => host).merge(:auth_object => host, :permission => :auto_provision_discovered_hosts), :method => :post]
+    actions <<  [_('Refresh facts') ,hash_for_refresh_facts_discovered_host_path(:id => host).merge(:auth_object => host, :permission => :edit_discovered_hosts)]
+    actions <<  [_('Reboot') ,hash_for_reboot_discovered_host_path(:id => host).merge(:auth_object => host, :permission => :edit_discovered_hosts), :method => :put]
```
tbrisker, line 18: Why adding these elements to the array instead of declaring it right away as an array containing all of them?
Unless you want to only add those that are permitted, as we do for example in https://github.com/theforeman/foreman/blob/develop/app/helpers/hosts_helper.rb#L148 and then you don't need to check the permissions again below.

app/helpers/discovered_hosts_helper.rb
lzap: Sure.

app/helpers/discovered_hosts_helper.rb
lzap: Yes, thanks.

app/helpers/discovered_hosts_helper.rb
lzap, line 18: I do want to check for permissions, each one of these have different permission name.

lzap:
I rebased, now I am hiding instead, the button seems to disappear when there are no buttons available which is exactly what I want.

dLobatog:
Merged as fc7a7ba8c9cb933f0015b45266eaea23ad6c472d, thanks @lzap!

PR #270, opened 2016-05-03
More fixes for random unit tests

lzap:
Should pass except the one that fails in core currently:

http://ci.theforeman.org/job/test_plugin_matrix/1393/database=postgresql,ruby=2.3,slave=fast/testReport/junit/(root)/HostJSTest__hosts%20index%20multiple%20actions/test_0001_show_action_buttons/

### DIFF
--- a/test/functional/api/v2/discovered_hosts_controller_test.rb
+++ b/test/functional/api/v2/discovered_hosts_controller_test.rb
@@ -134,6 +134,8 @@ class Api::V2::DiscoveredHostsControllerTest < ActionController::TestCase
     post :auto_provision, { :id => host.id }
     assert_match host.name, @response.body
     assert_response :success
+  ensure
+    SETTINGS[:organizations_enabled] = true
   end
 
   def test_auto_provision_success_and_delete


lzap:
This hopefully fixes the last bunch @dLobatog

test/unit/discovered_extensions_test.rb
```diff
@@ -4,6 +4,8 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
   include Foreman::Controller::DiscoveredExtensions
 
   setup do
+    SETTINGS[:organizations_enabled] = true
```
dLobatog: This isn't needed in other tests, we always test with taxonomies on and don't guarantee tests will pass with taxonomies off afaik. Tests pass on the latest branch for both discovery and foreman without this, any ideas to reproduce the failure?

lzap: I think I found it. One test turned off organizations, looks like some tests are being scheduled after this one.

lzap:
Looks like it, there is only this weird JS error: http://ci.theforeman.org/job/test_plugin_matrix/1395/

```
HostJSTest::hosts index multiple actions.test_0001_show action buttons
```

https://github.com/theforeman/foreman/pull/3476

dLobatog:
Merged as fbd492f915df5b6fdd990678d2373db9ab449c79, thanks @lzap!

PR #271, opened 2016-05-09
Fixes #14964 - improved taxonomy settings

lzap:
Two changes - tests are red after the IPv6 branch merge in core. The IPAM mode
looks useless in this kind of test.

Second, Ruby warnings are useless when our dependencies have millions of them.

Don't think we need an issue for this kind of change.


lzap:
Do not merge, I will also fix settings in this PR.

domcleal:
Please check the Redmine ticket number, this changes an unrelated ticket.

lzap:
@orrabin I also added descriptions to all settings and merged into one transaction

orrabin:
@lzap tests are failing on`NameError: undefined method 'value' for module '#<Module:0x0000000971bae8>` you need to remove `settings_helper_extensions.rb` if we aren't using it anymore and that will get rid of the error.
Otherwise we can wait for: https://github.com/theforeman/foreman/pull/3503 to get merged.

lzap:
Oh youre right, removed.

app/models/setting/discovered.rb
```diff
@@ -16,50 +16,37 @@ def self.load_defaults
...
+        self.set('discovery_facts_network', N_("Regex to organize facts for network section"), "", N_("Network facts")),
+        self.set('discovery_facts_ipmi', N_("Regex to organize facts for ipmi section"), "", N_("IPMI facts")),
       ].compact.each { |s| self.create s.update(:category => "Setting::Discovered")}
     end
 
     if SETTINGS[:locations_enabled]
       Setting.transaction do
         [
-          self.set('discovery_location', N_("The default location to place discovered hosts in"), ""),
+          self.set('discovery_location', N_("The default location to place discovered hosts in"), "", N_("Discovery location"), { :collection => Hash[Location.all.map{|loc| [loc[:title], loc[:title]]}] }),
```
orrabin: Missing the `value` here and in organizations: https://github.com/theforeman/foreman/blob/develop/app/models/setting.rb#L275
This relied on the default nil for value so it should be:

``` ruby
self.set('discovery_location', N_("The default location to place discovered hosts in"), "", N_("Discovery location"), nil, { :collection => Hash[Location.all.map{|loc| [loc[:title], loc[:title]]}] }),
```

lzap: Oh right! Now it works! Rushed out for lunch. Never push when hungry... Sorry. Amended.

lzap:
@orrabin for the record, the failure is caused by our JS driver, it's random

orrabin:
merged as 4ec24b4f5f119a9e56bddcea194a223bb43c7f6f and 10f7e344759ab7c8732de28fb237fc630a9334c4, thanks @lzap!
I changed `Primary fact` to `Interface fact` because the name was unclear.

PR #273, opened 2016-05-31
Fixes #4426 - Add modal wizard to host creation

ShimShtein:
Each discovered host in thelist gets a modal wizard that will ask for organization, location and hostgroup before redirecting to host edit screen.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -50,20 +50,36 @@ class DiscoveredHostsController < ::ApplicationController
 
   def edit
     @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false) unless @host.nil?
-    render :template => 'hosts/edit'
+    @host.attributes = @host.apply_inherited_attributes(params[:host]) unless params[:host].empty?
+    @host.set_hostgroup_defaults
+    setup_host_class_variables(@host)
+    @override_taxonomy = true
+    if params[:quick_submit]
+      perform_update(@host, _('Successfully provisioned %s') % @host.name)
+    else
+      render :template => 'hosts/edit'
+    end
   end
 
   def update
     @host = ::ForemanDiscovery::HostConverter.to_managed(@host)
     forward_url_options
+    @host.attributes = params[:host]
+
+    perform_update(@host)
+  end
+
+  def perform_update(host, success_message = nil)
     Taxonomy.no_taxonomy_scope do
-      if @host.update_attributes(params[:host])
-        process_success :success_redirect => host_path(@host), :redirect_xhr => request.xhr?
+      if host.save
+        success_options = { :success_redirect => host_path(host), :redirect_xhr => request.xhr? }
+        success_options[:success_msg] = success_message if success_message
+        process_success success_options
       else
         taxonomy_scope
         load_vars_for_ajax
         offer_to_overwrite_conflicts
-        process_error :object => @host, :render => 'hosts/edit'
+        process_error :object => host, :render => 'hosts/edit'
       end
     end
   end
@@ -183,6 +199,18 @@ class DiscoveredHostsController < ::ApplicationController
   end
 
   private
+
+  def setup_host_class_variables(host)
+    if host.hostgroup
+      @architecture    = host.hostgroup.architecture
+      @operatingsystem = host.hostgroup.operatingsystem
+      @environment     = host.hostgroup.environment
+      @domain          = host.hostgroup.domain
+      @subnet          = host.hostgroup.subnet
+      @compute_profile = host.hostgroup.compute_profile
+      @realm           = host.hostgroup.realm
+    end
+  end
 
   def init_regex_and_categories
     hightlights = Setting[:discovery_facts_highlights].empty? ? /^(productname|memorysize|manufacturer|architecture|macaddress$|processorcount|physicalprocessorcount|discovery_subnet|discovery_boot|ipaddress$)/ : Regexp.new(Setting[:discovery_facts_highlights])
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -75,4 +75,28 @@ module DiscoveredHostsHelper
     "<span class='glyphicon #{status_glyph}' style='color: #{status_color}'
       title='#{status_message}'/>".html_safe
   end
+
+  def host_taxonomy_select(f, taxonomy)
+    # Add hidden field with taxonomy value to be updated by the form on submit
+    tax_html = super
+    tax_id = "#{taxonomy.to_s.downcase}_id"
+    selected_taxonomy = params[:host][tax_id]
+    if @override_taxonomy && selected_taxonomy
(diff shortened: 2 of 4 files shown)


lzap:
Thanks, this is a great idea that actually solves several issues we have with the edit form while allowing the flexibility to still use it.

Now, the first problem I have is when I click on Provision button and wait about 5 seconds, the modal reloads somehow:

image

lzap:
For the record, I love the idea and I want to merge it for sure.

Another issue is when I click on Cancel the whole screen remains dimmed for some time, then it returns to normal.

lzap:
Tried both Chrome and Firefox stable.

Made log record of the click and extra load: https://gist.github.com/lzap/ef531c2eaaf9330570f7db82fdce24fa

We will need to cover the new dialog with functional tests, but thats later, let's get it working first.

app/helpers/discovered_hosts_helper.rb
```diff
@@ -75,4 +75,16 @@ def discovery_status_icon(host)
...
+    # Add hidden field with taxonomy value to be updated by the form on submit
+    tax_html = super
+    tax_id = "#{taxonomy.to_s.downcase}_id"
+    selected_taxonomy = params[:host][tax_id]
+    if @override_taxonomy && selected_taxonomy
+      hidden_tax = f.hidden_field tax_id.to_sym, :value => selected_taxonomy
+      tax_html += hidden_tax
+    end
+    tax_html
+  end
```
lzap, line 89: Where does this gets called?

ShimShtein, line 89: It's a hack into host edit form, I add a hidden input field to it that will send org and loc properties to host#update process, so it will update them too.

ShimShtein, line 89: They are disabled in the form, so they are not sent to the server, and we cannot set them.

ShimShtein:
Fixed the unwanted reload issue and removed the fade effect from the dialog - it should be more immediate now.

ShimShtein:
[test]
Test failure seems unrelated, let's try to rerun.

lzap:
Ok great, there is one big issue. When I select a Hostgroup in the modal and click on Provision, I get the form with correct HG set, but OS details are not set. I need to switch it back and forth in order to load up details. Looks like we need to trigger 

Can you also make the modal form a little bit narrower and on the other hand, enlarge all the dropdown boxes to the full width? Also I'd perhaps swap Quick Create and Create host buttons to have the blue one on the right side.

There is also a minor issue when using Quick Create - rendering fails after host finishes provisioining:

```
 | Foreman::Exception (ERF42-4516 [Foreman::Exception]: Unknown action name for success message: edit):
 |   app/controllers/application_controller.rb:243:in `process_success'
 |   /home/lzap/work/foreman_discovery/app/controllers/discovered_hosts_controller.rb:73:in `block in perform_update'
```

It also looks like after quick provisitioning, the puppetmaster/ca fields are not inherited (traditional workflow does work fine).

ShimShtein:
Fixed:
1. OS details are shown now
2. Fields are wider
3. Buttons swapped

I have also changed the parameters for `#process_success` so it should give a valid message. Unfortunately on my laptop the provisioning is broken, so couldn't test it completely.

I think that puppetmaster and ca should be fixed in quick provisioning too - it kind of comes with the OS fix. Couldn't test it either.,

app/controllers/discovered_hosts_controller.rb
```diff
@@ -50,20 +50,36 @@ def destroy
 
   def edit
     @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false) unless @host.nil?
-    render :template => 'hosts/edit'
+    @host.attributes = @host.apply_inherited_attributes(params[:host]) unless params[:host].empty?
+    @host.set_hostgroup_defaults
+    setup_host_class_variables(@host)
+    @override_taxonomy = true
+    if params[:quick_submit]
+      perform_update(@host, _('Successfully provisioned %s.') % @host.name)
```
lzap: I'd prefer without dot.

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -38,8 +38,39 @@
       <td class="hidden-tablet hidden-xs"><%= host.primary_interface.try(:subnet) %></td>
       <td class="hidden-tablet hidden-xs"><%= disc_report_column(host) %></td>
       <td>
+        <!-- Modal -->
+      <div class="modal" id="<%= "myModal-#{host.id}" %>" tabindex="-1" role="dialog" aria-labelledby="myModalLabel">
```
lzap: Rename `myModal` to reasonable id.

lzap:
It works great, couple if minor comments above.

Now, we have zero integration tests coverage of the provisioning flow. Can you add at least one test for quick provision case? I expect this to become the most used form now.

ShimShtein:
Fixed small issues and added tests.

lzap:
Javascript issues... [test]

app/views/discovered_hosts/_discovered_hosts_list.html.erb
```diff
@@ -38,8 +38,39 @@
       <td class="hidden-tablet hidden-xs"><%= host.primary_interface.try(:subnet) %></td>
       <td class="hidden-tablet hidden-xs"><%= disc_report_column(host) %></td>
       <td>
+        <!-- Modal -->
```
lzap, line 41: The comment is out of indentation.

One more thing, I realized that the modal dialog is being rendered per each individual host. It looks like it's all same for all hosts, can you create just one instance of it and little bit of javascript that sets required things?

ShimShtein, line 41: > can you create just one instance of it and little bit of javascript that sets required things?

It's doable, but why? It adds more complexity to the view, but I can't see a real benefit except maybe smaller HTML that gets to the client.
From programming point of view, there is only one place to change if we need changes to the modal.
If the amount of the HTML is an issue, we can offload it completely to AJAX.

lzap, line 41: Yeah not big deal, its just I have 100 hosts per page setting and page is quite big.

lzap:
Merged as 2e08637f70d4d9941b6527793bb300b6824741a6, thank you! Yay!

lzap:
Thanks a lot. Let me show you a trick: https://github.com/lzap/bin-public/blob/master/discover-host

Use the script to create fake discovery host, then you can provision them, reboot orchestration step is not taken as an error (this step will obviously fail), so you can easily test discovery in most of cases.

Now with this knowledge, you are more than welcomed to review other patches in this repo :-)

ShimShtein:
Cool script! I will surely endorse it.
I did something similar, but way less effective: I just created an empty `Host::Discovered` entry using console and fixed validation errors until I was able to save it. It's enough to show the host in the list, so I was able to check the basic use cases.
Thanks for the review!

lzap:
I need to test this patch with Katello, I suspect it won't work with other fields like Content Source, Environment or CV. Not sure.

lzap:
Ok that was not the case.

PR #275, opened 2016-06-15
Fixes #15368 - refresh existing discovered host

lzap:
Renamed host was created twice.


lzap:
Oh wrong branch on my side.

PR #277, opened 2016-06-20
Fixes #9784 - autoprovisioning populates from hostgroup

lzap:
Sets ignored flags like Environment, Puppet or CA. This patch is only for
auto-provisioning. For edit host form, you can now use the new modal dialog
which was merged last week.

When testing this, you may hit: http://projects.theforeman.org/issues/15469

That's a different story.

### DIFF
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -54,8 +54,11 @@ module Foreman::Controller::DiscoveredExtensions
     host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
     # fallback to the original if template did not expand
     host.name = original_name if host.name.empty?
+    # explicitly set all inheritable attributes from hostgroup
+    host.attributes = host.apply_inherited_attributes(hostgroup_id: rule.hostgroup_id)
+    host.set_hostgroup_defaults
     # save! does not work here
-    host.save
+    host.save ? host : false
   end
 
   def perform_reboot_all hosts = Host::Discovered.all
--- a/test/unit/discovered_extensions_test.rb
+++ b/test/unit/discovered_extensions_test.rb
@@ -170,6 +170,19 @@ class FindDiscoveryRulesTest < ActiveSupport::TestCase
     assert_equal host.name, "macaabbccddeeff"
   end
 
+  test "attributes from hostgroup are copied after auto provisioning" do
+    facts = @facts.merge({"somefact" => "abc"})
+    host = Host::Discovered.import_host(facts)
+    domain = FactoryGirl.create(:domain)
+    subnet = FactoryGirl.create(:subnet_ipv4, :name => 'subnet_100', :network => '192.168.100.0', :organizations => [host.organization], :locations => [host.location])
+    hostgroup = FactoryGirl.create(:hostgroup, :with_environment, :with_rootpass, :with_os, :subnet => subnet, :domain => domain)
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -54,6 +54,8 @@ def perform_auto_provision original_host, rule
     host.name = host.render_template(rule.hostname) unless rule.hostname.empty?
     # fallback to the original if template did not expand
     host.name = original_name if host.name.empty?
+    # explicitly set all inheritable attributes from hostgroup
+    host.attributes = host.apply_inherited_attributes(hostgroup_id: rule.hostgroup_id)
```
ShimShtein, line 58: Please call the older `set_hostgroup_defaults` too. It will set domain, os, architecture, medium, ptable and subnet.

lzap, line 58: Done, blindly pushing, I hope tests will be green.

ShimShtein:
:+1: That should do the trick. @orrabin please pay attention to fields inherited from hostgroup when you test it.

dLobatog:
@lzap :+1: , please add a test though, so we can close this bug "forever" :smile:

lzap:
I fixed tests, I still see one error locally, not sure if this gets reproduced on jenkins. @ShimShtein any idea why we have this regression:

```
  1) Failure:
DiscoveredHostsControllerTest#test_edit_form_submit_parameters [/home/lzap/work/foreman_discovery/test/functional/discovered_hosts_controller_test.rb:109]:
Expected at least 1 element matching "#host_operatingsystem_id [selected]", found 0..
Expected 0 to be >= 1.
```

ShimShtein:
My first guess (without testing this) will be disconnection between architecture and OS (maybe wrong taxonomy on those objects) or maybe the OS is not copied because the `#managed?` is [not turned on](https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L617).

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -95,19 +105,28 @@ def test_edit_form_quick_submit
...
-        :host => {
-          :hostgroup_id => hostgroup.id
-        } }, set_session_user_default_manager
-
-      assert_select '#host_operatingsystem_id [selected]' do |elements|
-        assert_equal hostgroup.operatingsystem.id.to_s, elements.first[:value]
-      end
+    host = Host::Discovered.import_host(@facts)
+    host.name = "discoveredhost1"
+    host.save!
```
lzap: The name change is not necessary, it's there to be able to set a breakpoint later.

FYI I just pushed my attempt to fix the broken test for traditional provisioning via form. Can't get it working.

lzap:
I am disabling test named `test_edit_form_submit_parameters` as I don't want to block this PR. There is issue with Rails helper method loading which was introduced in https://github.com/theforeman/foreman_discovery/pull/273 and we need another PR and some code changes to fix this.

Once this is merged we can re-test also other pending PR like https://github.com/theforeman/foreman_discovery/pull/261

lzap:
@dLobatog my goal is to close out this PR and #261 in order to move on towards new discovery release.

lzap:

This is the last push towards Discovery 6.0 release (http://projects.theforeman.org/versions/100) - please help to review and merge @dLobatog @orrabin thanks!

dLobatog:
Merged as 8e3bc400b510e3c9d16a200656e80056d5fb62a6, thanks @lzap!

PR #278, opened 2016-06-21
Fixes #15469 - fixed host_taxonomy_select nil exception

ShimShtein:
When editing host, the `params[:host]` is still `nil`


lzap:
Merged as 2a4ad073d59072507b051059ca50b430b27446bc, thank you!

PR #282, opened 2016-07-19
Fixes #15731 - introduced discovery_delete_facts setting

lzap:
It does not make sense not to delete them, initial facts upload will overwrite
everything. This is blocking users from using some facts in templates.

### DIFF
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -17,6 +17,7 @@ class Setting::Discovered < ::Setting
     Setting.transaction do
       [
         self.set('discovery_fact', N_("Fact name to use for primary interface detection"), "discovery_bootif", N_("Interface fact")),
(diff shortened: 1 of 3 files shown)


lzap:

This is the last push towards Discovery 6.0 release (http://projects.theforeman.org/versions/100) - please help to review and merge @dLobatog @orrabin thanks!

lzap:
Renamed to "clean" which sounds a bit better.

orrabin:
I tested before the name change and this works well.
I checked that after clicking `provision` the facts are still there but once I save the host they are deleted. When the setting is false they are still there.

Waiting for the tests to be green again and then this is good to merge.

lzap:
Looks green, thanks!

orrabin:
merged as 4aa96e6960c9634b21c0cbffb651772f89f2f3ad, thanks @lzap!

PR #252, opened 2016-02-22
Fixes #13829 - view hosts permission added to reader

lzap:
SSIA, see the linked BZ for details. I am _not_ adding a seed script to add
this permission to existing roles for security reasons - users must explicitly
add this permission if they need it.

PR #283, opened 2016-07-19
Fixes #9306 - added missing view_* permissions

lzap:
@orrabin reproduced a bug you reported, can you test?

When you were testing similar PR
https://github.com/theforeman/foreman_discovery/pull/252 you verified it
without need of adding the permission. But I believe the permission will not be
added automatically and we need a migration.

For this PR I needed to manually add the permission in the Role via UI to get
it working. Can you confirm? In that case, we need a migration similar to
a6826bee71d7818959aca7dbbba035fb94a8c204

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -95,10 +95,9 @@ module Api
       def facts
         state = true
         @discovered_host = Host::Discovered.import_host(params[:facts])
+        Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning" unless @discovered_host
         if Setting['discovery_auto'] && @discovered_host && rule = find_discovery_rule(@discovered_host)
           state = perform_auto_provision(@discovered_host, rule)
-        else
-          Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning"
         end
         process_response state
       rescue Exception => e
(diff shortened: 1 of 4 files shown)


lzap:
We just found similar issue with OS: https://github.com/theforeman/foreman_discovery/pull/284

I think we either need to name all the permissions in the "Discovery Reader" role.

lzap:
We will likely fix this in #284 - then we can close this one.

lzap:
Looks like we are missing lot of view_ permissions, for this reason Discovery Manager was unable to provision correctly. This fixes both auto and UI process. Finally now we have a good test for UI provisioning - it checks all important fields.

I am including a migration that adds all required view permissions, we need to warn users in the readme notes for 6.0 release, but I don't see big security impact. Opinion @dLobatog ?

I stumbled upon small bug which I reported separately: http://projects.theforeman.org/issues/15735

Ready for review.

lzap:
Forgot to git add the migration file! Fixed.

lzap:
I also moved "Discovered facts import unsuccessful, skipping auto provisioning" message a bit, it was reported in case import _was_ successful but auto provisioning was not triggered.

dLobatog:
Fair enough - please leave a big warning on the release notes or the manual the role permissions scope is much larger than it was, it can catch people by surprise.

dLobatog:
Merged as 1d7465f68110ad249f2b81c540767e22510d3416, thanks @lzap!

PR #284, opened 2016-07-19
Fixes #15732 - Fixed unavailable operating system

ShimShtein:
1. Added default operating system view role to discovery reader.
2. Removed unnecessary helper calls, since all helpers are added [automagically](https://github.com/rails/rails/blob/a4b5582/actionpack/lib/action_controller/metal/helpers.rb#L95)
3. Removed no taxonomies scope from the test.

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -26,10 +26,6 @@ module ForemanDiscovery
       end
     end
 
-    initializer 'foreman_discovery.helper' do |app|
-      ActionView::Base.send :include, DiscoveredHostsHelper
-    end
-
     initializer 'foreman_discovery.register_gettext', :after => :load_config_initializers do |app|
       locale_dir = File.join(File.expand_path('../../..', __FILE__), 'locale')
       locale_domain = 'foreman_discovery'
@@ -107,6 +103,7 @@ module ForemanDiscovery
           :view_organizations,
           :view_locations,
           :view_hosts,
+          :view_operatingsystems,
           # discovered_hosts
           :view_discovered_hosts,
           # discovered_rules
@@ -180,9 +177,6 @@ module ForemanDiscovery
 
       # Include subnet extensions
       ::Subnet.send :include, DiscoverySubnet
-
-      # Include helper for dashboard
-      ::DashboardHelper.send(:include, DiscoveredHostsHelper)
     end
 
     rake_tasks do
--- a/test/functional/discovered_hosts_controller_test.rb
+++ b/test/functional/discovered_hosts_controller_test.rb
@@ -95,19 +95,16 @@ class DiscoveredHostsControllerTest < ActionController::TestCase
   end
 
   def test_edit_form_submit_parameters
-    disable_taxonomies do
-      host = Host::Discovered.import_host(@facts)
-      domain = FactoryGirl.create(:domain)
-      hostgroup = FactoryGirl.create(:hostgroup, :with_subnet, :with_environment, :with_rootpass, :with_os, :domain => domain)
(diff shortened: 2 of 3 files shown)


lib/foreman_discovery/engine.rb
```diff
@@ -107,6 +103,7 @@ class Engine < ::Rails::Engine
           :view_organizations,
           :view_locations,
           :view_hosts,
+          :view_operatingsystems,
```
lzap, line 106: Nice, I need to test if this adds the permission to an existing role, otherwise we'd need a migration like:

```
class AddOrglocPermissions < ActiveRecord::Migration
  def up
    [Role.find_by_name("Discovery Manager"), Role.find_by_name("Discovery Reader")].each do |r|
      next if r.nil?
      r.add_permissions!(["view_operatingsystems"])
    end
  end

  def down
  end
end
```

lzap, line 106: Yeah, it does not add, you need to add manually. Can you add the migration above? There are no security concerns about this one, we can simply add it. If the permission is already present, we can add second instance, no big deal.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -95,19 +95,16 @@ def test_edit_form_quick_submit
...
+    domain = FactoryGirl.create(:domain)
+    hostgroup = FactoryGirl.create(:hostgroup, :with_subnet, :with_environment, :with_rootpass, :with_os, :domain => domain)
+    get :edit, {
+      :id => host.id,
+      :host => {
+        :hostgroup_id => hostgroup.id
+      } }, set_session_user_default_manager
+    assert_select '#host_operatingsystem_id [selected]' do |elements|
+      assert_equal hostgroup.operatingsystem.id.to_s, elements.first[:value]
     end
```
lzap, line 108: Can you also add asserts for:
- puppet environment, ca, master
- architecture, media, ptable

This covers all problematic ones. I think we might have more issues than OS - I just saw we also need view_ permissions for all the others.

lzap, line 108: Let's add them all, let me create the list for you.

lzap, line 108: - :view_architectures
- :view_compute_resources
- :view_domains
- :view_realms
- :view_environments
- :view_hostgroups
- :view_hosts
- :view_images
- :view_locations (already present)
- :view_media
- :view_models
- :view_operatingsystems
- :view_provisioning_templates
- :view_ptables
- :view_puppetclasses
- :view_smart_proxies
- :view_subnets
- :view_organizations (already present)

lzap, line 108: Please add all of these to the migration as well.

lzap:
Since there is already opened ticket on permissions, I will do the above in separate PR #283 thanks!

lzap:
Merged as c3eca9a9960f232ca5eec13f1cc66762f28937b2, thank you!

lzap:
Just for the record (and googlers), if you encounter a test failure:

```
  1) Failure:
DiscoveredHostsControllerTest#test_edit_form_submit_parameters [/home/lzap/work/foreman_discovery/test/functional/discovered_hosts_controller_test.rb:121]:
Expected at least 1 element matching "#host_hostgroup_id [selected]", found 0..
Expected 0 to be >= 1.

125 runs, 332 assertions, 1 failures, 0 errors, 0 skips
```

Make sure you have this patch and Discovery roles have the `view_operatingsystems` permission.

Edit: Scratch that, this is incompatibility between core versions: https://github.com/theforeman/foreman_discovery/pull/291

PR #285, opened 2016-07-20
Fixes #15750 - strong parameters (v1.13+)

lzap:
Do not merge before https://github.com/theforeman/foreman/pull/3659

Also prior merging this try to merge as many other PRs as possible to minimize
cherry-picking for 6.0 plugin release (which is 1.12 compatible).

This refactoring revealed some leaks in few functional tests - I was not
expecting return codes or result contents correctly and it was silently
failing.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -1,6 +1,7 @@
 module Api
   module V2
     class DiscoveredHostsController < ::Api::V2::BaseController
+      include Foreman::Controller::Parameters::DiscoveredHost
       include Foreman::Controller::DiscoveredExtensions
 
       before_filter :find_resource, :except => %w{index create facts auto_provision_all reboot_all}
@@ -40,7 +41,7 @@ module Api
       param_group :discovered_host, :as => :create
 
       def create
-        @discovered_host = Host::Discovered.new(params[:discovered_host])
+        @discovered_host = Host::Discovered.new(discovered_host_params)
         process_response @discovered_host.save
       end
 
@@ -78,7 +79,7 @@ module Api
       def update
         @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host)
         forward_request_url
-        update_response = @host.update_attributes(params[:discovered_host])
+        update_response = @host.update_attributes(discovered_host_params)
         process_response update_response
       end
 
@@ -93,15 +94,16 @@ module Api
       param :facts, Hash, :required => true, :desc => N_("hash containing facts for the host with minimum set of facts: discovery_bootif, macaddress_eth0, ipaddress, ipaddress_eth0, interfaces: eth0 (example in case primary interface is named eth0)")
 
       def facts
+        facts = params['facts']
         state = true
-        @discovered_host = Host::Discovered.import_host(params[:facts])
+        @discovered_host = Host::Discovered.import_host(facts)
         Rails.logger.warn "Discovered facts import unsuccessful, skipping auto provisioning" unless @discovered_host
         if Setting['discovery_auto'] && @discovered_host && rule = find_discovery_rule(@discovered_host)
           state = perform_auto_provision(@discovered_host, rule)
         end
         process_response state
       rescue Exception => e
-        logger.warn "Host discovery failed, facts: #{params[:facts]}"
+        logger.warn "Host discovery failed, facts: #{facts}"
         logger.debug e.message + "\n" + e.backtrace.join("\n")
         render :json => {'message'=>e.to_s}, :status => :unprocessable_entity
       end
--- a/app/controllers/concerns/foreman/controller/parameters/discovered_host.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/discovered_host.rb
@@ -1,0 +1,24 @@
+module Foreman::Controller::Parameters::DiscoveredHost
+  extend ActiveSupport::Concern
+  include Foreman::Controller::Parameters::HostBase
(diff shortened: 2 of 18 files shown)


app/controllers/concerns/foreman/controller/parameters/discovered_host.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+  end
+
+  def discovered_host_params
+    self.class.discovered_host_params_filter.filter_params(params, parameter_filter_context)
+  end
+
+  def discovered_host_search_params
+    params.permit(:search, :order, :page)
+  end
```
lzap, line 23: These are not mass-assigned, I will remove the helper here and in `discovery_rule` as well.

lib/foreman_discovery/engine.rb
```diff
@@ -139,6 +139,11 @@ class Engine < ::Rails::Engine
           :parent => :configure_menu,
           :after=> :hostgroups
 
+        # strong parameters
+        parameter_filter Subnet, :discovery_id
+        parameter_filter Host::Managed, :discovery_rule_id
+        parameter_filter Hostgroup, :discovery_rules, :discovery_rule_ids, :discovery_rule_names, :type
```
domcleal: Should be `:type, :discovery_rules => [], :discovery_rule_ids => [], :discovery_rule_names => []` to permit an array of IDs/names.

lzap: Thanks for review @domcleal !

lzap:
I have changed the parameter_filter and removed non-mass-assigned search params. Tests will fail - the core part need to be merged first.

lzap:
Added also `host.build` test in auto UI provisioning functional tests as described in https://github.com/theforeman/foreman_discovery/pull/289 (could not add them there because these tests were broken - asserts were incorrect).

app/controllers/concerns/foreman/controller/parameters/discovered_host.rb
```diff
@@ -0,0 +1,20 @@
+module Foreman::Controller::Parameters::DiscoveredHost
+  extend ActiveSupport::Concern
+  include Foreman::Controller::Parameters::HostBase
+  include Foreman::Controller::Parameters::HostCommon
+
+  class_methods do
+    def discovered_host_params_filter
+      Foreman::ParameterFilter.new(::Host::Discovered, :top_level_hash => 'host').tap do |filter|
```
domcleal: :top_level_hash isn't accepted here any more, this second arg should be able to be removed.  (And it should now work for your API controller.)

lzap: > :top_level_hash isn't accepted here any more, this second arg should be able to be removed.  (And it should now work for your API controller.)

Thanks for headsup, I will fix that and test your PR tomorrow.

## 

Later,
 Lukas #lzap Zapletal

lzap:
I rebased and removed the unnecessary param. I have some test failures locally but I believe this must be some migration issue. Pushing to see how Jenkins will like that.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -50,7 +51,7 @@ def destroy
 
   def edit
     @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false) unless @host.nil?
-    @host.attributes = @host.apply_inherited_attributes(params[:host]) unless params[:host].empty?
+    @host.attributes = @host.apply_inherited_attributes(discovered_host_params) unless discovered_host_params.empty?
```
domcleal: You may need to add a `:host` argument to discovered_host_params and pass that through as the second arg to `filter_params` to specify the top level hash, as it won't be the standard `:discovered_host`.  Note that the API controller _does_ use :discovered_host, so the arg should default to `nil` and not be set in that controller.

Edit: i.e. `self.class.discovered_host_params_filter.filter_params(params, parameter_filter_context, :host)` from here, but `self.class.discovered_host_params_filter.filter_params(params, parameter_filter_context, nil)` for the API controller.

lzap: Oh that was it and combination of broken tests. The UI test was in bad shape. Updated, ready for re-review!

lzap:
@dLobatog @domcleal guys discovery develop is busted until this is merged: http://ci.theforeman.org/job/test_plugin_matrix/1741/database=postgresql,ruby=2.0.0,slave=fast/console

Thanks!

ohadlevy:
what needs to be done? isnt it simply adding back the protected_attributes
gem to discovery?

lzap:
> what needs to be done? isnt it simply adding back the protected_attributes
> gem to discovery?

Sorry what? Please follow

https://github.com/theforeman/foreman/pull/3659

for more info.

## 

Later,
 Lukas #lzap Zapletal

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -93,15 +94,16 @@ def destroy
       param :facts, Hash, :required => true, :desc => N_("hash containing facts for the host with minimum set of facts: discovery_bootif, macaddress_eth0, ipaddress, ipaddress_eth0, interfaces: eth0 (example in case primary interface is named eth0)")
 
       def facts
+        facts = params['facts']
```
dLobatog, line 97: Does this change anything?

dLobatog:
Looks good to me now, tests pass, PXELess, regular PXE, and auto-provision with rules all worked flawlessly so I'm merging this

dLobatog:
Merged as 34db0a489cbd5e8d1cdb5da5ef079828f829fdf1, thanks @lzap!

PR #286, opened 2016-07-21
Do not run test:discovery on test rake task

lzap:
Everytime I do `rake test` in core, this also runs all plugin tests. This makes
`rake test test/unit/some_test.rb` unusable.

Discovery follows generic patter for all plugins, I would like to discuss a
different possibility in this PR. For now I am removing this, but let's come up
with something better.

### DIFF
--- a/lib/discovery.rake
+++ b/lib/discovery.rake
@@ -1,21 +1,22 @@
 # Setup Tests
 namespace :test do
   desc "Test Discovery plugin"
-  Rake::TestTask.new(:discovery) do |t|
+  Rake::TestTask.new(:foreman_discovery) do |t|
     test_dir = File.join(File.dirname(__FILE__), '..', 'test')
     t.libs << ["test",test_dir]
     t.pattern = "#{test_dir}/**/*_test.rb"
     t.verbose = true
     t.warning = false
   end
+
+  task :discovery => :foreman_discovery
 end
-Rake::Task[:test].enhance do
-  Rake::Task['test:discovery'].invoke
-end
+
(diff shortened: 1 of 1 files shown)


lzap:
So the clear solution is to introduce new target like `test:plugins` and specify it explicitly to test all plugins. We'd need to change Jenkins configurations and file PRs for all plugins we test against. What do you say @domcleal on that?

lzap:
I filed https://github.com/theforeman/foreman/pull/3666 - it's a required patch in core. Let's test it for discovery and then modify plugin template.

lib/discovery.rake
```diff
@@ -1,21 +1,24 @@
...
-  Rake::TestTask.new(:discovery) do |t|
+  Rake::TestTask.new(:foreman_discovery) do |t|
     test_dir = File.join(File.dirname(__FILE__), '..', 'test')
     t.libs << ["test",test_dir]
     t.pattern = "#{test_dir}/**/*_test.rb"
     t.verbose = true
     t.warning = false
   end
+
+  task :discovery => :foreman_discovery
```
lzap, line 12: This is temporary until I change all my scripts and stuff.

lzap:
Test will apparently fail until the core change is merged.

pronix:
can i test plugin isolated from theforeman ? for example just unit tests ?

lzap:
Yes, that's what you do with `be rake test:discovery` now. This patch corrects when running core tests (discovery tests were executed too by default which was time consuming).

rabajaj0509:
@lzap the core PR was closed and now I don think PR makes a lot sense now. Do you want to close this one?

PR #287, opened 2016-07-22
Fixes #15769 - discovery nested taxonomy fixed

lzap:
SSIA test with Org/Suborg.


dLobatog:
Merged as 30a514d69039ae4b7f97b1ada1d41033cf701f5b, thanks @lzap!

PR #288, opened 2016-07-22
Fixes #15788 - Discovery widget crashes for hosts w/o reports

dLobatog:
The widget is calling 'host.last_report < 7.days.ago' - which fails with
"ActionView::Template::Error (undefined method `<' for nil:NilClass')"
if the host has no reports. This crashes the whole dashboard page.


lzap:
Merged as 6e4e02870fc6eb7afc8e73ecb878fdf5504b7005, thank you!

PR #289, opened 2016-07-22
Fixes #15793 - quick provisioning sets build mode

lzap:
Regression introduced by our modal dialog.

Needed for successful testing of

https://github.com/theforeman/foreman_discovery/pull/226

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -71,6 +71,7 @@ class DiscoveredHostsController < ::ApplicationController
 
   def perform_update(host, success_message = nil)
     Taxonomy.no_taxonomy_scope do
+      ::ForemanDiscovery::HostConverter.set_build_clean_facts(host)
       if host.save
         success_options = { :success_redirect => host_path(host), :redirect_xhr => request.xhr? }
         success_options[:success_msg] = success_message if success_message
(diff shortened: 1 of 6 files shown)


lzap:
Found one additional issue, hold on.

app/controllers/discovered_hosts_controller.rb
```diff
@@ -71,6 +71,7 @@ def update
 
   def perform_update(host, success_message = nil)
     Taxonomy.no_taxonomy_scope do
+      host.build = true
```
dLobatog: Please add a test that ensures `host.build` is true when we do this action or something, that way we won't get this regression again

lzap: Will do

lzap:
So the problem was bigger. First, quick provisioning did not set the build flag, this was easy. But the flag can't be set just like `host.build = true`, we need to execute some statements from `to_managed` in ordere to setup things (legacy flag, fact clear override). Therefore I had to extract it into separate method. Added a unit test for that.

But during testing I figured out that our new discovery_clean_facts setting is implemented incorrectly. We want to let users to pick either "delete none" or "delete only non-discovered facts" while it was implemented with choice of "delete all" or "delete only non-discovered facts". So it was not useful at all.

I fixed that as well.

dLobatog:
Tests run fine locally - It's also able to clean up facts & provision fine, so I'm merging

dLobatog:
Merged as 454c6fa0833686587a9afd6a9fbeee01c4927efd, thanks @lzap!

PR #292, opened 2016-08-05
Refs #12634 - PXEGrub 1 and 2 are locked too

lzap:
When https://github.com/theforeman/foreman/pull/3679 is merged, discovery will
no longer work when Lock PXE setting is turned on. This fixes it.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -114,18 +114,21 @@ class Host::Discovered < ::Host::Base
     self.discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => id).first_or_create
     self.discovery_attribute_set.update_attributes(import_from_facts)
     # lock the host into discovery via PXE, if feature is enabled in settings
-    if Setting::Discovered.discovery_lock? && self.subnet.tftp?
-      begin
-        @host = self
-        Rails.logger.info "Locking discovered host #{self.mac} in subnet #{subnet}"
-        template = unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
-        subnet.tftp_proxy.set mac, :pxeconfig => template
-      rescue ::Foreman::Exception => e
-        ::Foreman::Logging.exception("Could not set tftp_proxy from proxy", e)
-      end
-    end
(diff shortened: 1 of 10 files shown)


dLobatog:
[test]

app/models/host/discovered.rb
```diff
@@ -117,9 +117,11 @@ def populate_fields_from_facts(facts = self.facts_hash, ignored_type = nil)
     if Setting::Discovered.discovery_lock? && self.subnet.tftp?
       begin
         @host = self
-        Rails.logger.info "Locking discovered host #{self.mac} in subnet #{subnet}"
-        template = unattended_render(::ProvisioningTemplate.find_by_name(Setting['discovery_lock_template']).template)
-        subnet.tftp_proxy.set mac, :pxeconfig => template
+        TemplateKind::PXE.each do |kind|
+          Rails.logger.info "Locking discovered host #{self.mac} in subnet #{subnet} via #{kind} template"
+          template = unattended_render(::ProvisioningTemplate.find_by_name(Setting["discovery_#{kind}_lock_template"]).template)
```
dLobatog: This is going to fail to find the setting in a test environment, normally we circumvent the issue through fixtures in core. The method has many responsibilities already, if you consider moving this to a separate method I think it'd be nicer.

dLobatog: Stubbing is fine too, probably better as adding fixtures from rails engines is cumbersome

dLobatog:
@lzap notice tests break because of trying to find a Setting. Other than that :+1: , can you create an issue specific for discovery though? Normally projects shouldn't share issue numbers, especially between core and others because the issue numbers are used for releases. 

Thanks!

lzap:
Amended, fixed tests, pulled method out, fixed ref issue number (it was not correct). Should be all fine now.

lzap:
Rebased.

lzap:
Irrelevant test failure, rebased and pushed again.

lzap:
Mind merging @dLobatog ? I refactored tests and I would like to rebase my other PR on top of this one as it touches lots of test files.

lzap:
@dLobatog the reason is http://projects.theforeman.org/issues/16143 (need to rebase it on top of this due to test refactoring in here)

dLobatog:
Thanks @lzap !

PR #293, opened 2016-08-10
Fixes #16041 - Unify accordion styles in show page

dLobatog:
This commit fixes the heading of highlights to look similar to the other
headings (used to violate the 'rule' of no more than 3 font-sizes/types
in the same page)

It also uses Patternfly's accordion carets instead of the +/- glyphicons
which we had to modify via javascript. That way we get rid of the
redundancy, conform to how Patternfly recommends accordions to look, and
we don't need to write any JS code for that.


lzap:
Merged as b6eb17feb938ce93f6791ea59170c46f63baf60f, thank you!

PR #295, opened 2016-08-15
Added missing auto-provisioning asserts

lzap:
The test was actually not stressing puppet and ca flags, environment which is
tested is actually not INHERIT enhanced.


dLobatog:
[test]

dLobatog:
Merged as d81ea562a74e5cbc16deb15653cf6cd06823c0f8, thanks @lzap!

PR #296, opened 2016-08-15
Refs #15750 - added missing strong params for rule

lzap:


app/controllers/concerns/foreman/controller/parameters/discovery_rule.rb
```diff
@@ -5,6 +5,8 @@ module Foreman::Controller::Parameters::DiscoveryRule
     def discovery_rule_params_filter
       Foreman::ParameterFilter.new(::DiscoveryRule).tap do |filter|
         filter.permit :name, :search, :hostname, :priority, :enabled, :hostgroup, :hostgroup_id, :max_count
+        filter.permit :organization_ids => []
```
domcleal: `include Foreman::Controller::Parameters::Taxonomix` and just call `add_taxonomix_params_filter(filter)` instead, as it permits other related parameters automatically.

lzap: Thanks @domcleal amended, it now includes unit tests around these params. RTM @orrabin

dLobatog:
Merged as 543e3eacb939c0f99ba6754f2398f79e3d35d543, thanks @lzap!

PR #297, opened 2016-08-16
Fixes #16105 - Force DNS rebuild when provisioning discovered host

imriz:


imriz:
[test]

lzap:
Will take a look tomorrow, thanks!

app/controllers/discovered_hosts_controller.rb
```diff
@@ -73,6 +73,7 @@ def update
   def perform_update(host, success_message = nil)
     Taxonomy.no_taxonomy_scope do
       ::ForemanDiscovery::HostConverter.set_build_clean_facts(host)
+      @host.primary_interface.discovery_queue_rebuild_dns unless @host.primary_interface.pending_dns_record_changes?
```
lzap: This code will not work with auto-provisioning, make sure it work in both ways. Anyways.

I think instead of calling this explicitly, we should create new `after_validation :queue_dns_discovery` hook. In the hook method, check if we are converting from Discovered, like we do in other hooks, then queue it when necessary.

Add tests for normal and autoprovisioning please. And make the commit message shorter :-)

lzap:
Example:

``` ruby
module MyExtensions
  extend ActiveSupport::Concern

  included do
    after_validation :queue_dns_rebuild
  end

  def queue_dns_rebuild
    return unless errors.empty? && !Setting[:discovery_always_rebuild_dns]
    return if new_record? # Discovered Hosts already exist, and new_records will break `find`
    return unless type_changed? and ::Host::Base.find(self.id).type == "Host::Discovered"

    # do your stuff here
  end

```

lzap:
The code could be perhaps part of discovery's `Host::ManagedExtensions` I think. Just another method I believe.

imriz:
> The code could be perhaps part of discovery's Host::ManagedExtensions I think. Just another method I believe.

Mmm, maybe the same convention as the other orchestration concerns? `Orchestration::Discovery::DnsExtensions` or something like that, or, since it belongs to the Nic model, `Nic::ManagedExtensions`?

imriz:
@lzap updated the PR.

app/models/setting/discovered.rb
```diff
@@ -31,6 +31,7 @@ def self.load_defaults
         self.set('discovery_facts_ipmi', N_("Regex to organize facts for ipmi section"), "", N_("IPMI facts")),
         self.set('discovery_lock', N_("Automatically generate PXE configuration to pin a newly discovered host to discovery"), false, N_("Lock PXE")),
         self.set('discovery_lock_template', N_("PXE template to be used when pinning a host to discovery"), 'pxelinux_discovery', N_("Locked template name"),nil,{ :collection => Proc.new {Hash[ProvisioningTemplate.all.map{|template| [template[:name], template[:name]]}]} }),
+        self.set('discovery_always_rebuild_dns', N_("Force DNS entries creation when provisioning discovered host"), true, N_("Force DNS")),
```
lzap, line 34: Looks great. Why this is enabled by default? Normally, discovered hosts should not be added to DNS, thus I'd set this to `false`.

app/models/nic/managed_extensions.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def discovery_queue_rebuild_dns
+    return unless (dns? || dns6? || reverse_dns? || reverse_dns6?) && errors.empty? && Setting[:discovery_always_rebuild_dns]
+    return if self.host.new_record? # Discovered Hosts already exist, and new_records will break `find`
+    return unless self.host.type_changed? and ::Host::Base.find(self.host.id).type == "Host::Discovered"
+    return if self.pending_dns_record_changes?
+    logger.debug "Queuing DNS rebuild for #{self}"
+    queue.create(:name => _("Rebuild DNS for %s") % self, :priority => 10,
+                 :action => [self, :set_discovery_rebuild_dns])
```
lzap: Priority should be I think 15, because DNS orchestration already have 10 and tasks with same priority have "undefined" order and we don't want to rebuild first.

lzap: I'd not break this line since the lines above are not wrapped too.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -268,9 +272,62 @@ def test_reboot_all_error
...
+  end
+
+  def prepare_hostgroup_for_dns_rebuild(host)
+    hostgroup = setup_hostgroup(host)
+    hostgroup.domain = FactoryGirl.create(:domain,
+                                          :name => 'myorchdomain.net',
+                                          :dns => FactoryGirl.create(:smart_proxy,
+                                                                     :features => [FactoryGirl.create(:feature, :dns)]
+                                          )
+    )
```
lzap: I'd prefer one line really, this is horrible :-)

lzap:
What a coincidence, I just accepted my very first MyHeritage Instant Discovery e-mail when I noticed your domain. I am newcomer there. I was wondering if you could update http://projects.theforeman.org/projects/foreman/wiki/Who_Uses_Foreman (if that's valid and appropriate).

Very nice patch indeed. Please do my smallish comments and I will test and merge it later this week.

test/functional/discovered_hosts_controller_test.rb
imriz: really?? I find it more readable, but I don't care either way - I will change it.

app/models/setting/discovered.rb
imriz: This only affects when you provision the host, not when it enters discovery, and in that case you always expect DNS to be created.

imriz:
> I was wondering if you could update http://projects.theforeman.org/projects/foreman/wiki/Who_Uses_Foreman (if that's valid and appropriate).

Sure! will do :)

app/models/setting/discovered.rb
lzap, line 34: > This only affects when you provision the host, not when it enters discovery, and in that case you always expect DNS to be created. 

Disregard my comment, this should be turned on by default.

## 

Later,
 Lukas #lzap Zapletal

app/models/nic/managed_extensions.rb
imriz: > Priority should be I think 15, because DNS orchestration already have 10 and tasks with same priority have "undefined" order and we don't want to rebuild first.

The rebuild is only fired when there are no DNS changes pending, so it cannot collide. Therefore, I think 10 is the most appropriate priority, since this is the priority DNS changes are executed.

lzap: > The rebuild is only fired when there are no DNS changes pending, so it cannot collide. Therefore, I think 10 is the most appropriate priority, since this is the priority DNS changes are executed. 

Yeah, for some reason I misunderstood the whole motivation. I agree.

## 

Later,
 Lukas #lzap Zapletal

imriz:
@lzap PR updated per your comments, and I've updated the "Who_Uses_Foreman" :)

Thanks!!

lzap:
It work fine, in my environment I see:

```
2016-08-26T14:38:42 2576109c [app] [I] Create DHCP reservation for x525400e94080.local.lan-52:54:00:e9:40:80/192.168.122.101
2016-08-26T14:38:42 2576109c [app] [D] DHCP reservation on net 192.168.122.0 with attrs: {:hostname=>"x525400e94080.local.lan", :mac=>"52:54:00:e9:40:80", :ip=>"192.168.122.101", :network=>"192.168.122.0", :nextServer=>"192.168.122.1", :filename=>"pxelinux.0"}
2016-08-26T14:38:42 2576109c [app] [I] Delete the DNS A record for x525400e94080.local.lan/192.168.122.101
2016-08-26T14:38:42 2576109c [app] [I] Add DNS A record for x525400e94080.local.lan/192.168.122.101
2016-08-26T14:38:42 2576109c [app] [I] Delete the DNS PTR record for 192.168.122.101/x525400e94080.local.lan
2016-08-26T14:38:42 2576109c [app] [D] Fetching DNS reservation for 192.168.122.101/x525400e94080.local.lan
2016-08-26T14:38:42 2576109c [app] [I] Add DNS PTR record for 192.168.122.101/x525400e94080.local.lan
```

The initial DNS deletion failed on proxy and returned 404 but this was not reason for rollback and the transaction continued. Which brings this question - do we want really to _rebuild DNS_ (delete/add operations) or simply _build DNS_ (add operation only)? I lean towards the latter to prevent such 404 errors. Or both maybe (make it a three state flag), because I would like to implement naming discovered hosts using the "deacon" gem we recently added (http://projects.theforeman.org/issues/16330).

Explain me your use case - why you prefer rebuilding instead of building DNS entry? Do you create DNS name during discovery phase somehow?

imriz:
No specific reason. I used it because it is safe, and can overcome stale
records, if such exists. I don't mind changing to add only if you prefer.

lzap:
Merged as fe529727fcea06d52d8ea2b9e8d6da899617e038, thank you!

PR #298, opened 2016-08-25
Fixing gem requirement in README

daviddavis:
Specifying a git location for a gem uses master by default. Looks like master was last updated in 2014 (version 1.2.0).


lzap:
Merged as 22546317d5508d973b14809713b9f6853ba13a7c, thank you!

PR #299, opened 2016-08-26
Fixes #16324 - kexec templates work under safemode

lzap:
Easy one.

### DIFF
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -44,8 +44,7 @@ module Host::ManagedExtensions
     template = provisioning_template(:kind => 'kexec')
     raise ::Foreman::Exception.new(N_("Kexec template not associated with operating system")) unless template
     @host = self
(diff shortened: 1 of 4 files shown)


lzap:
@dLobatog when this is merged I will do new nightly build, regression :-(

lzap:
Since @dLobatog will be busy later this week, mind taking a look @orrabin ? Quick and easy merge. Context: `@kernel, @initrd` wont work in safe mode, plus I renamed them because when doing template preview, these names are already used and gets overwritten.

orrabin:
I tested this but needed https://github.com/theforeman/foreman/pull/3783 to actually get to the point that uses the template for that.
Since this is broken anyway I'll merge it now and it's affect will be noticeable after core pr is merged.

orrabin:
merges as 78261c3e7da85a14bea5100ef7987c01e7149694, thanks @lzap!

PR #302, opened 2016-09-27
Fixes #16657 - max char limit 255 - discovery_rule columns

swapab:


test/unit/discovery_rule_test.rb
```diff
@@ -9,6 +9,57 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
     @hostgroup.save!
   end
 
+  context "#validates attribute" do
+    setup do
+      @huge_name = SecureRandom.hex(128)
```
lzap: Thank you for the patch, this is not too obvious that this generates actually 256 characters, could you change it to something like `"X" * 256`?

lzap (changes requested):
Other than that, WFM! Thanks.

test/unit/discovery_rule_test.rb
swapab: @lzap  Done. Looks more evident.
Thanks for reviewing my PR.

PR #304, opened 2016-09-30
Fixes #16750 - *_selected methods expect params[:host]

dLobatog:
Due to the strong params changes, the methods we override in routes.rb
(architecture_selected, os_selected, medium_selected) do not longer
work.

The reason is that these *_selected methods expect params[:host] to
exist. Other methods like 'taxonomy_scope' also expect params[:host] to
exist.

However, 'discovered_hosts/edit' sends params[:discovered_host], which
is not understood by the HostController.

The discovery plugin should realize that these params will not be
understood by HostController and change them appropriately.

---

This patch merely checks the URL and if it matches 'discovered_hosts' in
some place, it will also set params[:host].

### DIFF
--- a/app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
@@ -1,0 +1,18 @@
+module ForemanDiscovery
+  module Concerns
+    module HostsControllerExtensions
+      extend ActiveSupport::Concern
+      included do
+        before_action :set_discovered_params
+      end
+
+      # Change params to what the hosts controller expects. Certain methods
+      # in the hosts controller like all the _selected methods, taxonomy_scope,
+      # etc.. expect a params[:host] to work.
+      def set_discovered_params
+        return unless request.path.match(/discovered_hosts/).present?
+        params[:host] ||= params[:discovered_host]
+      end
+    end
+  end
+end
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -8,6 +8,7 @@ module ForemanDiscovery
   class Engine < ::Rails::Engine
 
     # support pre-4 Rails versions
+    config.autoload_paths += Dir["#{config.root}/app/controllers/foreman_discovery/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
@@ -200,6 +201,7 @@ module ForemanDiscovery
       ::Nic::Managed.send :include, Nic::ManagedExtensions
 
       # Controller extensions
+      ::HostsController.send :include, ForemanDiscovery::Concerns::HostsControllerExtensions
       ::Api::V2::FactValuesController.send :include, Api::V2::FactValuesControllerExtensions
     end
 
--- a/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
+++ b/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
@@ -1,0 +1,25 @@
+require 'test_plugin_helper'
+require 'test_helper_discovery'
+
+module ForemanDiscovery
+  module Concerns
+    class HostsControllerTest < ActionController::TestCase
+      tests ::HostsController
+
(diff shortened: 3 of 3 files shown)


dLobatog:
@lzap This is a problem with nightly and Foreman 1.13, I think we might want to make a minor relase after this (and other patch I'm making for displaying the templates).

lib/foreman_discovery/engine.rb
```diff
@@ -9,6 +9,7 @@ module ForemanDiscovery
   class Engine < ::Rails::Engine
 
     # support pre-4 Rails versions
+    config.autoload_paths += Dir["#{config.root}/app/controllers/foreman_discovery/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
```
lzap, line 12: Why adding another path? We already have that on line 13. But if you prefer this path, just move all the concerns into the new folder.

lib/foreman_discovery/engine.rb
```diff
@@ -194,6 +195,8 @@ class Engine < ::Rails::Engine
       ::Nic::Managed.send :include, Nic::ManagedExtensions
 
       # Controller extensions
+      ::HostsController.send :include,
+        ForemanDiscovery::Concerns::HostsControllerExtensions
```
lzap: I'd prefer this on one line. No biggie.

app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
```diff
@@ -0,0 +1,18 @@
...
+      extend ActiveSupport::Concern
+      included do
+        before_action :set_discovered_params
+      end
+
+      # Change params to what the hosts controller expects. Certain methods
+      # in the hosts controller like all the _selected methods, taxonomy_scope,
+      # etc.. expect a params[:host] to work.
+      def set_discovered_params
+        return unless request.url.match(/discovered_hosts/).present?
```
lzap: Can you match only path part of url to prevent false execution. To be honest, I'd prefer full comparison of the path rather than regexp.

lzap (changes requested):
LGTM Thanks, Jenkins does not work currently, but I have a failure locally (I rebased on develop everything):

```
1) Error:
ForemanDiscovery::Concerns::HostsControllerTest::hosts controller requests from discovered_hosts url#test_0001_get "host" params from "discovered_hosts" params:
NoMethodError: undefined method `+' for nil:NilClass
    /home/lzap/work/foreman_discovery/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb:17:in `block (2 levels) in <class:HostsControllerTest>'
```

lzap:
Ping, I think this is fairly important one. How do you do? :-)

lib/foreman_discovery/engine.rb
dLobatog: Just because a 'HostsControllerExtensions' without scoping it in ForemanDiscovery seemed too generic and possibly another plugin would use that.

app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
dLobatog: I'm not sure what you mean, do you want me to check if 'discovered_hosts' is included in the URL part after request.host (that is, the path) or what exactly..?

lzap: So my concern is these is an url like `/some/url?param=discovered_hosts` that could match. AFAIR we don't have this one, but we could we.

My proposal is either match only the `path` element or at least add slash: `/\/discovered_hosts\//`.

lib/foreman_discovery/engine.rb
lzap: Ok that's fine, we should move everything in a separate PR perhaps.

dLobatog:
Repushed so that it matches on `request.path` rather than `request.url`. Thanks for the review @lzap 😄

dLobatog:
@lzap looks like integration tests are broken in Discovery?

dLobatog:
@lzap looks like integration tests are broken in Discovery?

dLobatog:
@lzap looks like integration tests are broken in Discovery?

dLobatog:
@lzap looks like integration tests are broken in Discovery?

dLobatog:
@lzap looks like integration tests are broken in Discovery?

dLobatog:
[test]

lzap:
I was just about to start digging to find out why this happens downstream. Apologize for the late reply. Thanks!

PR #305, opened 2016-09-30
Fixes #15548 - Discovered host supports template methods

dLobatog:
Mixes in Hostext::OperatingSystem from core, so that the methods used to
resolve what templates belong to a Host, can be used in Discovered::Host
too.

This fixes the 'Resolve' button on the provision discovered host form.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -2,6 +2,7 @@ class Host::Discovered < ::Host::Base
   include ScopedSearchExtensions
   include Foreman::Renderer
   include BelongsToProxies
+  include ::Hostext::OperatingSystem
 
   belongs_to :hostgroup
   has_one    :discovery_attribute_set, :foreign_key => :host_id, :dependent => :destroy


dLobatog:
Depends on https://github.com/theforeman/foreman/pull/3905 to work

lzap:
How can I help Daniel? The core patch review is stuck, can you rebase all and make updates or comment on the status? This was reported multiple times already, bummer.

lzap:
This showed up on the coffee meeting, what is the status? Bummer bug, we missed 8.0 tho.

timogoebel:
@lzap : Can we think of a solution without needing to patch Forman Core? The core patch will only be in 1.15. That's quite late i think.

dLobatog:
@timogoebel @lzap I will update the patch that adds the methods to Host::Discovered, or research if I can selectively grab methods from Host::Managed to import to Host::Discovered (I doubt that's possible but who knows)

dLobatog:
I tried using the `to_managed` service to see if I could temporarily convert a discovered host (as a hack until 1.15 is out) just to get the templates but it seems to fail 😞

dLobatog:
[test] ok to test

The PR in core is merged so this should work now

dLobatog:
@lzap @timogoebel This works now & tests pass 😄

timogoebel:
Perfect. Let's 🚢 this then, shall we?

dLobatog:
@lzap @orrabin Could you test this out?

lzap:
Thanks @dLobatog awesome, this was pain in the blah.

PR #306, opened 2016-09-30
Fixes #16143 - call unused_ip for discovered host provisioning

lzap:
This patch adds explicit call of `unused_ip` during discovered provisioning which fixes issue when DHCP leases are outside of the DHCP range defined in the subnet.

The code changes required some changes in subnet detection and handling of subnets, which needed to be adapted for IPv6 support. However, this does not aim to add IPv6 support, it will still not work. What works is assigning IPv6 subnet from Hostgroup and assigning a IPv6 address from an IPAM.

Edit: The original PR was stuck (because it was trying to add bunch of other IPv6 features), it was hanging for over three years and rebased (summer 2019) with more lightweight version.


app/models/host/discovered.rb
```diff
@@ -80,41 +80,44 @@ def attributes_to_import_from_facts
...
-      if SETTINGS[:locations_enabled]
-        self.location = Location.find_by_title(Setting[:discovery_location]) ||
-          subnet.try(:locations).try(:first) ||
-          Location.first
-        Rails.logger.info "Assigned location: #{self.location}"
-      end
-      if SETTINGS[:organizations_enabled]
-        self.organization = Organization.find_by_title(Setting[:discovery_organization]) ||
-          subnet.try(:organizations).try(:first) ||
-          Organization.first
```
lzap: Basically, this line was removed in the patch, it is not relevant I was thinking to remove it while touching this part. It does make little sense (sets more or less random organization).

timogoebel:
@lzap : I haven't tried it or looked at the tests, but the IPv4/IPv6-Code looks good to me. Carry on. 👍

ohadlevy:
@lzap what is the status of this PR ?

lzap:
Status: Woudl love to find time to get back to this, this is one of major pains for discovery.

danny-smit:
I assume there isn't any progress since the last comment? I read at http://projects.theforeman.org/issues/16143 that some refactoring was done, does this patch still apply after that?
What would it take to get this one finished?

lzap:
It's bunch of work I currently don't have time for. Anyone can take from here.

lzap:
Guys, I pushed completely new code which implements assigning IP addresses from IPAM. It works both for IPv4 and IPv6 (however discovered hosts IPv6 NICs are not detected - I haven'ŧ tested this yet and this is out of scope).

What works:

* Discover a IPv4 host and provision it in the same subnet - new IPv4 address is acquired via IPAM.
* Discover a IPv4 host and provision it in a different subnet - new IPv4 and/or IPv6 address is acquired via IPAM.
* Works both via autoprovisioning and quick provision form.
* Does not work for Customize Host workflow (Edit Host form hack) - in this case error is thrown IP does not belong to subnet and user must click on Unused IP link to grab a new IP from IPAM.

@timogoebel

.rubocop.yml
```diff
@@ -14,30 +14,12 @@ AllCops:
 Rails:
   Enabled: true
 
-# Don't prefer is_a? over kind_of?
-Style/ClassCheck:
+Style:
```
timogoebel: Somebody seems to hate rubocop, I'd prefer to leave this out of this PR though.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -35,4 +34,33 @@ def self.set_build_clean_facts(host)
     host.build = true
   end
 
+  def self.unused_ip_for_subnet subnet, mac, existing_ip
```
timogoebel: This definitely needs some `()`.

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
```diff
@@ -4,35 +4,48 @@ module ImportHooks
...
-      def set_subnet(ip)
-        host.primary_interface.subnet = suggested_subnet(ip)
+      def set_subnets(ip, ip6)
+        host.primary_interface.subnet = suggested_ipv4_subnet(ip)
+        subnet6 = suggested_ipv6_subnet(ip6) if ip6
+        logger.warn("Detected IPv6 subnet but it's not set for the host, unsupported") if subnet6
       end
 
-      def suggested_subnet(ip)
+      def suggested_ipv4_subnet(ip)
```
timogoebel: This can and should be dryed up more.

test/unit/managed_extensions_test.rb
```diff
@@ -65,8 +66,8 @@ def self.table_name
     end
 
     test "setKexec calls renderer" do
-      Host::Discovered.any_instance.expects(:render_template).with() { |json| JSON.parse(json) }.once
-      Foreman::Renderer.expects(:render).returns({})
+      @host.expects(:render_kexec_template).once.returns({})
```
timogoebel: Why change this? This does not test `render_kexec_template` at all and we don't have any proper tests for this.

test/unit/discovered_extensions_test.rb
```diff
@@ -179,6 +180,33 @@ class DiscoveredExtensionsTest < ActiveSupport::TestCase
     assert_equal hostgroup.puppet_ca_proxy, managed_host.puppet_ca_proxy
   end
 
+  class StubIPAM
+    def suggest_ip
+      "192.168.101.13"
+    end
+  end
+
+  test "subnet is changed and unused_ip called" do
```
timogoebel, line 190: We need tests for the IPv6 path here as well.

test/unit/discovery_attribute_set_test.rb
```diff
@@ -42,6 +43,8 @@ class DiscoveryAttributeSetTest < ActiveSupport::TestCase
 
   test "can search discovered hosts by disks_size" do
     host = discover_host_from_facts(@facts)
+    # test not relevant if fact filter is set to filter out blockdevices
```
timogoebel: Unrelated?

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
```diff
@@ -4,35 +4,48 @@ module ImportHooks
...
-          logger.warn "Subnet could not be detected for #{ip}"
+          logger.info "IPv4 subnet could not be detected for #{ip}"
         end
         subnet
       end
 
+      def suggested_ipv6_subnet(ip6)
+        subnet6 = Subnet.subnet_for(ip6)
+        if subnet6
+          logger.info "Detected IPv6 subnet: #{subnet} with taxonomy #{subnet.organizations.collect(&:name)}/#{subnet.locations.collect(&:name)}"
```
timogoebel: This needs to be `subnet6. organizations`. One more reason to clean this up.

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
```diff
@@ -4,35 +4,48 @@ module ImportHooks
...
       end
 
       private
 
-      def set_subnet(ip)
-        host.primary_interface.subnet = suggested_subnet(ip)
+      def set_subnets(ip, ip6)
+        host.primary_interface.subnet = suggested_ipv4_subnet(ip)
+        subnet6 = suggested_ipv6_subnet(ip6) if ip6
+        logger.warn("Detected IPv6 subnet but it's not set for the host, unsupported") if subnet6
```
timogoebel: Why is this unsupported? What if a host just has solely an IPv6 address?

timogoebel (commented):
> Discover a IPv4 host and provision it in a different subnet

What is the real world use case for this?
I can totally see that users want the discovered host to get a new ip address, maybe the subnet has an "install dhcp range" and the servers are supposed to get ips from a different range. But changing the subnet usually means changing the vlan on the network switch as well. Or is this supposed to work with trunk interfaces where changing the subnet means changing the vlan?

test/unit/discovery_attribute_set_test.rb
lzap: Yes, going to move this into separate PR (separate commit in that PR): https://github.com/theforeman/foreman_discovery/pull/482

test/unit/managed_extensions_test.rb
lzap: Yes, going to move this into separate PR (separate commit in that PR): https://github.com/theforeman/foreman_discovery/pull/482

lzap: FYI this cannot test renderer, this is an orchestration test on a stub. The code does not exist actually :-)

lzap:
> Or is this supposed to work with trunk interfaces where changing the subnet means changing the vlan?

Yes, that's my assumption however I haven't tested this myself. When Subnet changes, also VLAN tag changes therefore provisioning process should be able to leverage that. The issue is PXE obviously (you'd need to change BIOS configuration) however remember that Discovery has PXE-less mode too.

I see complaints all the time when customers select a hostgroup with totally different subnet and tell me that IP address does not fall into the range. I don't ask why they do it tho. Customers do crazy things :-) As long as it does not break existing workflows, why not to add this? 

Frankly the major reason is case number (1) - let the IP to fall into the IPAM-defined range.

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
lzap: Sorry what?

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
lzap: To do what with? We don't support IPv6 provisioning yet, excluding some special cases which are not PXE. On the other hand, it's reasonable to set it properly here. Changing.

lzap:
Resolved all comments except the one. Also added some changes to actually detect IPv6-only interface and subnet correctly. Completely regenerated Rubocop TODO file and cleaned it a bit.

lzap:
Hmm I rebased and pushed into my branch but GH is not picking it up: https://github.com/lzap/foreman_discovery/commit/1c29e79c6eb6ba7acd9c09ba60ec50741be1d4c9 a bug?

lzap:
And it picked it up! Tests are green.

app/views/discovered_hosts/_discovered_host_modal.html.erb
```diff
@@ -17,6 +17,8 @@
           <% if show_location_tab? %>
             <%= select_f f, :location_id, Location.my_locations, :id, :to_label, {}, {:size => 'col-md-10'} %>
           <% end %>
+
+          <div>Changing Subnet is not possible for via Customize Host form, use Create Host or Auto provisioning to set Subnet from a Hostgroup.</div>
```
timogoebel: `Changing the Host's Subnet is not possible via Customize Host form, use Create Host or Auto provisioning and set the desired Subnet at the Hostgroup.`?

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
timogoebel: What I meant was: `suggested_ipv4_subnet` and `suggested_ipv6_subnet` are basically the same methods, just a parameter is different. This might be a little overengineered, but I think it's a good idea to keep the hooks as clean as possible. And everything that increases the code quality in discovery is a good thing.

```ruby
# app/services/foreman_discovery/subnet_suggestion.rb
module ForemanDiscovery
  class SubnetSuggestion
    attr_accessor :ip, :kind

    def self.for(ip:, kind:)
      new(ip: ip, kind: kind).()
    end

    def initialize(ip:, kind:)
     @ip = ip
     @kind = kind
    end

    def call
      return unless ip
      subnet = Subnet.unscoped.subnet_for(ip)
      if subnet
        logger.info "Detected #{kind} subnet: #{subnet} with taxonomy #{subnet.organizations.collect(&:name)}/#{subnet.locations.collect(&:name)}"
      else
        logger.info "#{kind} subnet could not be detected for #{ip}"
      end
      subnet
    end
  end
end
```

So you can just call

```ruby
ForemanDiscovery::SubnetSuggestion.for(ip: ip, kind: 'IPv4')
```

And this can be tested with dedicated unit tests if this get's fancier in the future.

timogoebel (changes requested):

app/views/discovered_hosts/_discovered_host_modal.html.erb
lzap: Sure, I think it should be "in the hostgroup". Let me change it.

lzap:
DONE!

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
```diff
@@ -4,35 +4,47 @@ module ImportHooks
...
-      def set_subnet(ip)
-        host.primary_interface.subnet = suggested_subnet(ip)
+      def set_subnets(ip, ip6)
+        host.primary_interface.subnet = ForemanDiscovery::SubnetSuggestion.for(ip: ip, kind: 'IPv4') if ip
+        host.primary_interface.subnet6 = ForemanDiscovery::SubnetSuggestion.for(ip: ip6, kind: 'IPv6') if ip6
       end
 
-      def suggested_subnet(ip)
-        subnet = Subnet.subnet_for(ip)
+      def suggested_ipv4_subnet(ip)
```
timogoebel: This can be removed now.

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
```diff
@@ -4,35 +4,47 @@ module ImportHooks
...
-          logger.info "Detected subnet: #{subnet} with taxonomy #{subnet.organizations.collect(&:name)}/#{subnet.locations.collect(&:name)}"
+          logger.info "Detected IPv4 subnet: #{subnet} with taxonomy #{subnet.organizations.collect(&:name)}/#{subnet.locations.collect(&:name)}"
         else
-          logger.warn "Subnet could not be detected for #{ip}"
+          logger.info "IPv4 subnet could not be detected for #{ip}"
         end
         subnet
       end
 
+      def suggested_ipv6_subnet(ip6)
```
timogoebel: This can also be removed now.

timogoebel (changes requested):

app/services/foreman_discovery/import_hooks/subnet_and_taxonomy.rb
lzap: Sure removed.

lzap:
Oh right rebased once again.

timogoebel (approved):
Makes sense and looks good.

lzap:
Great, will ping you once tests are green. Upcoming big changes for discovery incoming! I want to start working on https://community.theforeman.org/t/improving-discovery-workflow/15315 next week - please drop your ideas there before I start if possible.

lzap:
All green! :-)

timogoebel:
Thanks, @lzap.

PR #309, opened 2016-10-13
Refs #16899 - core helper is renamed in 1.14

ares:


dLobatog:
Thanks @ares  - I realized after merging it should have a different issue number though...

ares:
you mean a separate issue for the plugin?

dLobatog:
Yeah, refs is fine for fixing an issue again in the same repo, but not across different projects I believe

ares:
I thought that was applicable only in core, plugins (or at least rex) didn't enforce that policy). Anyway next time I can create separate issues.

PR #310, opened 2016-11-07
Fixes #17205 - fixed nil exception for DNS queue

lzap:
I could not reproduce the issue with `develop` but this should help.


dLobatog:
I cannot reproduce either but it makes sense - and it should return early given what's described on Redmine.

dLobatog:
Thanks @lzap !

PR #311, opened 2016-11-11
Fixes #17260 - remove managed flag before deleting

lzap:
Users are reporting "broken" discovered hosts, they are of type "Discovered",
but managed flag is set to true. This must be caused when host fails
provisioning at some point. We had transactions, but removed them since they
were not working with other plugins correctly.

This does not change the root cause, I am still searching how this was
possible, but I would like to create a short-term patch to allow deletion of
these hosts.


dLobatog:
Thanks @lzap !

PR #313, opened 2016-12-05
Fixes #17578 - display nested orgs / locs correctly

sean797:
Small thing but this should display nested orgs & locs correctly


lzap:
Thanks tests are green locally, we had a regression on jenkins. Have a fun weekend!

PR #314, opened 2016-12-06
Fixes #17585 - moved permission migration after override

lzap:
When upgrading Foreman from 1.11 to 1.13, this migration fails with.

```
undefined method `override=' for #<Filter:0x0000000e6f82c8>/opt/rh/sclo-ror42/root/usr/share/gems/gems/activemodel-4.2.5.1/lib/active_model/attribute_methods.rb:433:in `method_missing'
/usr/share/foreman/app/models/filter.rb:216:in `enforce_override_flag'
```

This patch makes sure the migration is triggered *after* the one in core that
is causing this (adding the override flag):

    db/migrate/20160818091420_add_override_flag_to_filter.rb

I considered using fake objects but it's quite complex code for this one and it
is not worth the effort (for an example see https://github.com/theforeman/foreman/blob/develop/db/migrate/20140219183343_migrate_permissions.rb - I'd need to implement all of these).

This will cause the migration to occur twice basically, but it is idempotent.
The permissions API no longer adds duplicities.

### DIFF
--- a/db/migrate/20160719124942_add_missing_view_permissions.rb
+++ b/db/migrate/20160719124942_add_missing_view_permissions.rb
@@ -1,18 +1,9 @@
 class AddMissingViewPermissions < ActiveRecord::Migration
   def up
-    permissions = [
-      "view_architectures", "view_domains", "view_environments", "view_hosts",
-      "view_hostgroups", "view_media", "view_models", "view_operatingsystems",
-      "view_provisioning_templates", "view_ptables", "view_puppetclasses",
-      "view_realms", "view_smart_proxies", "view_subnets"
-    ]
-    role = Role.find_by_name("Discovery Reader")
-    role.add_permissions!(permissions) if role
-    role = Role.find_by_name("Discovery Manager")
-    role.add_permissions!(permissions + ["create_hosts"]) if role
+    # migration renamed to db/migrate/20160818091421_add_missing_view_permissions2.rb
   end
 
   def down
-    # not implemented
+    # migration renamed to db/migrate/20160818091421_add_missing_view_permissions2.rb
   end
 end
--- a/db/migrate/20160818091421_add_permissions_from_default_roles.rb
+++ b/db/migrate/20160818091421_add_permissions_from_default_roles.rb
@@ -1,0 +1,15 @@
+class AddPermissionsFromDefaultRoles < ActiveRecord::Migration
+  def up
+    default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
+    ["Discovery Reader", "Discovery Manager"].each do |role_name|
+      role = Role.find_by_name(role_name) || next
+      default_permissions[role_name].each do |permission|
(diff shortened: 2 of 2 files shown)


lzap:
Don't merge, I screwed it up. Need to push.

lzap:
Now it's correct. If you don't mind quick merge @dLobatog or @orrabin this is 6.3 QA upgrade blocker. Thanks!

lzap:
I believe the failure is not related, develop broken :-(

ehelms:
I wonder if this would be better done in a seed?

lzap:
Plugins do this via lib/foreman_discovery/engine.rb actually and we have it there. The migration is only for existing roles.

db/migrate/20160818091421_add_missing_view_permissions2.rb
```diff
@@ -0,0 +1,21 @@
+class AddMissingViewPermissions2 < ActiveRecord::Migration
+  def up
+    # this migration was renamed from 20160719124942_add_missing_view_permissions.rb
+    return if ActiveRecord::Migrator.get_all_versions.include?(20160719124942)
```
orrabin: why not just check if the permissions exist in the list of permissions?

db/migrate/20160818091421_add_missing_view_permissions2.rb
```diff
@@ -0,0 +1,21 @@
+class AddMissingViewPermissions2 < ActiveRecord::Migration
+  def up
+    default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
```
orrabin: :+1: running this I found the list was missing view_locations which should be added.

db/migrate/20160818091421_add_missing_view_permissions2.rb
```diff
@@ -0,0 +1,21 @@
+class AddMissingViewPermissions2 < ActiveRecord::Migration
```
orrabin: Nitpick: please notice this now adds more then view permissions which is good because I had missing permissions in my discovery role but maybe change the name?

db/migrate/20160818091421_add_missing_view_permissions2.rb
```diff
@@ -0,0 +1,21 @@
...
+  def up
+    default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
+    ["Discovery Reader", "Discovery Manager"].each do |role_name|
+      role = Role.find_by_name(role_name) || next
+      puts "Processing #{role}"
+      default_permissions[role_name].each do |permission|
+        unless role.permission_names.include?(permission.to_sym)
+          puts "Adding permission #{permission}"
```
orrabin: do we want to actually show it when the migration is running or just add it to logs?

lzap:
Renamed, removed the puts statements.

orrabin:
@lzap looks good. Ready to merge when tests are green.

orrabin:
thanks @lzap!

lzap:
Many thanks!

PR #316, opened 2016-12-13
Fixes #17648 - Added support for attributes inheritance

ShimShtein:
Changed HostConverter.to_managed to automatically call attribute inheritance mechanism when a new set of attributes is introduced.

### DIFF
--- a/test/functional/api/v2/discovered_hosts_controller_test.rb
+++ b/test/functional/api/v2/discovered_hosts_controller_test.rb
@@ -28,6 +28,27 @@ class Api::V2::DiscoveredHostsControllerTest < ActionController::TestCase
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.stubs(:reboot).returns(true)
   end
 
+  def setup_hostgroup(host)
+    domain = FactoryGirl.create(:domain)
+    hostgroup = FactoryGirl.create(:hostgroup, :with_subnet, :with_environment, :with_rootpass, :with_puppet_orchestration, :with_os, :domain => domain, :organizations => [host.organization], :locations => [host.location])
+    hostgroup.medium.organizations << host.organization
+    hostgroup.medium.locations << host.location
+    hostgroup.ptable.organizations << host.organization
+    hostgroup.ptable.locations << host.location
+    hostgroup.domain.organizations << host.organization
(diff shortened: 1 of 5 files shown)


lzap:
Ok UI works, tests are passing locally, but can you add new test to check the update methods in both UI and API?

Adding also a test for UI edit method would be great too. This is very fragile code with regressions :-(

lzap:
CLI works too nice!

ShimShtein:
Tests added

test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -54,16 +69,42 @@ def test_delete_discovered_host
...
+    # Get the managed host instance from the DB
+    actual = Host.find(host.id)
+
+    assert_equal hostgroup.id, actual.hostgroup_id
+    assert_equal hostgroup.architecture_id, actual.architecture_id
+    assert_equal hostgroup.operatingsystem_id, actual.operatingsystem_id
+    assert_equal hostgroup.medium_id, actual.medium_id
+    assert_equal hostgroup.ptable_id, actual.ptable_id
+    assert_equal hostgroup.domain_id, actual.domain_id
+  end
```
lzap, line 103: Mind to extend this with three more fields - environment, puppet and ca? We had similar issues in the past with these too. It's optional, I buy it just like that too :-)

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -133,6 +133,29 @@ def test_edit_form_submit_parameters
...
+
+    # Get the managed host instance from the DB
+    actual = Host.find(host.id)
+
+    assert_equal hostgroup.id, actual.hostgroup_id
+    assert_equal hostgroup.architecture_id, actual.architecture_id
+    assert_equal hostgroup.operatingsystem_id, actual.operatingsystem_id
+    assert_equal hostgroup.medium_id, actual.medium_id
+    assert_equal hostgroup.ptable_id, actual.ptable_id
+    assert_equal hostgroup.domain_id, actual.domain_id
```
lzap, line 162: Ditto.

ShimShtein:
Extra checks added

lzap:
You are hero! Thanks a lot.

PR #320, opened 2016-12-30
fixes #17873 - clear association cache when converting host

timogoebel:
I thought I had seen everything, but this really takes the cake.
Details on the issue, to cut it short:

When rendering foreman_url for the PXELinux templates in some cases:

```ruby
proxy = @host.try(:subnet).try(:tftp)
```
and
```ruby
primary_interface = @host.try(:interfaces).try(:find) {|i| i.primary? }
proxy = primary_interface.try(:subnet).try(:tftp)
```

should return the same, but they don't. This is, because the subnet through primary interface association is not up to date.

I wasn't able to come up with a test for this. Suggestions welcome.

### DIFF
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -5,6 +5,7 @@ class ForemanDiscovery::HostConverter
   # Creates shallow copy.
   def self.to_managed(original_host, set_managed = true, set_build = true)
     host = original_host.becomes(::Host::Managed)
+    host.clear_association_cache
     host.type = 'Host::Managed'
     # the following flags can be skipped when parameters are set to false
     if set_managed


lzap:
Yeah, combination of hacky STI and complex New Host javascript form. Honestly I have no idea how to test this and I would much rather invest time in refactoring discovered host (http://projects.theforeman.org/issues/14035 and http://projects.theforeman.org/issues/13848) instead. I suggest to merge this without unit test. @orrabin ?

timogoebel:
@lzap : I'm currently trying to come up with a decent unit test. Let me try one or two more things. Also this currently beaks two tests. I already know why. Will push an updates version later.

lzap:
@timogoebel ok but since caching is also disabled during tests, I would be fine just with doing `expects` on the method you added to ensure this won't get lost in the future...

lzap:
I see you moved the line, what is the status?

timogoebel:
@lzap: Unit tests pass with the line moved. I manually retested it again in our environment with the line moved. Fixed the issue. I gave up with the test for now. I'd say: Merge it, if anyone comes up with a decent test for this we can merge it in a separate PR.

For my reference: Current status of the test is here: https://gist.github.com/timogoebel/0bd8b02ec51dd9903419e20a4ad34570

lzap:
Thanks a lot for the fix, nasty ugly horrible one this is :-)

timogoebel:
Thank you for your help with this, @orrabin and @lzap !

lzap:
Timo, I tried to reproduce with the added line commented out and I did not reproduce it at all.

- Discovered a host with two NICs
- Set the other one as managed, provision, primary
- Set new DNS name
- Submit

TFTP orchestration was performed and foreman_url renders to Smart Proxy. All is fine, this is develop. Can you re-test? I am going to revert it for https://github.com/theforeman/foreman_discovery/pull/346

timogoebel:
@lzap : I fear, I don't have time to test this this week. I'm fine with removing the line for now. I saw the error with 1.13. Maybe a newer rails version or something fixes the issue?

timogoebel:
@lzap: Despite the awesome weather, I did test this. Here are my findings:

I created two discovered hosts with this script:
https://gist.github.com/timogoebel/86b5f333efea150e50d1960e1e7f5de3
I don't know where I stole that from, probably from you. Thanks a lot, anyways.

For both hosts, I changed the network settings like this:


1. Remove IP, DNS-Name, Primary, Provision from `fake1` interface. Managed is still active.
2. Activate managed flag for `fake2`
3. Add a new interface: Type bond, bond0 as identifier, Domain: example.com, Managed true, Primary true, Provision true, Attached devices: `fake1,fake2`.
4. Suggest a random name.

After provisioning, the hosts look like this:

First host, with `host.clear_association_cache` commented out (e.g. without this patch).

image
Note: `fake1` interface still has a subnet and an address set.

Second host, with `host.clear_association_cache` enabled:

image
Note: `fake1` is clean.

Do you see that as well? I'm still investigating further.

timogoebel:
@lzap: I tried to recreate the reproducer I wrote in the bug report for this and wasn't able to reproduce this. I think, we're safe to revert this and I will try this again in a production environment and see if I still see the error.

lzap:
Ok thanks for putting such an effort in this. First, let's identify what is the problem here because I struggle to understand this. So do you expect the subnet and IP to disappear from fake1? Because the latter is known issue which I am about to finally fix, the unused_ip is never called, we simply leave IP address "as is" and don't touch it: https://github.com/theforeman/foreman_discovery/pull/306

The former (subnet) is suspicious, we detect that and set it explicitly during discovery phase, this might confuse hostgroup handling so it thinks this is explicitly set and keeps this. Maybe all we need is to override this from hostgroup. Edit: One thing we need to keep in mind and that is that primary interface always has subnet set (if it was previously discovered) because then the last orchestration step in queue (reboot) won't be performed if Foreman has no direct connection.

We need to store subnet during discovery phase to detect which proxy to use for communication towards proxy, we basically "abused" subnet of a host (which is an alias to primary.subnet). Maybe this was bad design from the day one and we want to introduce new field for this. It just felt natural to use it.

PR #324, opened 2017-01-10
Fixes #16890 - prevent discovery of managed hosts

lzap:
Users were reporting inconsistent records, I implemented dirty workaround the other day:

https://github.com/theforeman/foreman_discovery/pull/311

Now I found the root cause finally, when existing (managed) host was there,
discovery process treated it as discovered host and made some changes
incorrecty. This patch prevents that, adds a test and removes the dirty
workaround.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -30,10 +30,6 @@ class Host::Discovered < ::Host::Base
     where(taxonomy_conditions).order("hosts.created_at DESC")
   }
 
-  before_destroy { |record|
-    record.update_attribute(:managed, false)
-  }
-
   def self.import_host facts
     raise(::Foreman::Exception.new(N_("Invalid facts, must be a Hash"))) unless facts.is_a?(Hash)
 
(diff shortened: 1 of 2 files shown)


app/models/host/discovered.rb
```diff
@@ -54,13 +50,13 @@ def self.import_host facts
...
     hosts = ::Nic::Managed.where(:mac => bootif_mac, :primary => true)
-    if hosts.size == 0
+    if hosts.empty?
       host = Host.new(:name => hostname, :type => "Host::Discovered")
     else
       Rails.logger.warn "Multiple discovered hosts found with MAC address #{name_fact}, choosing one" if hosts.size > 1
       host = hosts.first.host
     end
-    host.type = "Host::Discovered"
+    raise ::Foreman::Exception.new(N_("Host already exists as managed: #{host.name}/#{bootif_mac}")) if host.type != "Host::Discovered"
```
timogoebel, line 59: .new("Host already exists as managed: %s", "#{host.name}/#{bootif_mac}")

to generate the ERF code correctly?

timogoebel (changes requested):

app/models/host/discovered.rb
dLobatog: I'll do this on merge

dLobatog:
Merged as 210f143bc85c58caeb67e8bf9a5cc2edbe764683 thanks @lzap & @timogoebel

PR #303, opened 2016-09-29
fixes #14036,#16722 - use plugin hook for subnet smart proxy

timogoebel:
This needs https://github.com/theforeman/foreman/pull/3904

PR #325, opened 2017-01-17
Fixes #18106 - include host common filters

lzap:
Host common filters were removed due to

http://projects.theforeman.org/issues/17706

but proper fix was to include the `BelongsToProxies` concern. This was done in

http://projects.theforeman.org/issues/16722

So in this patch I am putting the filters back. Katello plugin and facet
building mechanism needs to access attributes.


Seemone:
I will test this tomorrow 2017/01/18

dLobatog:
@lzap Thanks - I'm not understanding why this PR needed if this change was merged https://github.com/theforeman/foreman_discovery/pull/303/files#diff-319f0a9a9c2e1f456b56db13685f460dR4 ?

PR #327, opened 2017-02-24
add API route to rename hostname without triggering conversion to ManagedHost

isratrade:


isratrade:
This feature is needed for the stand-alone [fusor-server](https://github.com/fusor/fusor-server) project that only uses API's and doesn't use a foreman plugin to extend models or controllers.

lzap:
Thanks, we need a test for this.

lzap:
@isratrade if you want this in 1.15 (which is a milestone release) push today please.

isratrade:
@lzap, this PR is no longer needed for the [fusor](https://github.com/fusor/fusor-ember-cli/) project I was working in. If you still feel it is worthwhile to the community, I will write a test.

lzap:
Thanks that'd be helpful if you can get to this. Just one positive integration test is fine.

ohadlevy:
@lzap which usage case does this solve? Can't we always do a put and change the name?

lzap:
We can, closing.

PR #328, opened 2017-02-26
fixes #18677 - adds notification events on host discovery

ohadlevy:


ohadlevy:
requires https://github.com/theforeman/foreman/pull/4337

ohadlevy:
image
image

ohadlevy:
@rohoover anything we can improve to make this notification a bit more interesting? (icons etc?)

ohadlevy:
@rohoover another question - this implementation allows notification to "expire" once the notification condition is no longer set, for example, once we have an unused discovered hosts the notification remains, but once it is consumed, the notification is removed. I didn't want to provide a link to an invalid action (e.g. go do something with the host and then realize there are none).

Rohoover:
@ohadlevy 

I would try using an alternate icon in lieu of the "i". One thought to consider, why do you want this particular type of notification to stand out more than any other. I'm not against the idea, I would just want to have clear reasoning why this type is ranked more important than another. 

What is your second question?

ohadlevy:
@Rohoover 

> I would try using an alternate icon in lieu of the "i". One thought to consider, why do you want this particular type of notification to stand out more than any other. I'm not against the idea, I would just want to have clear reasoning why this type is ranked more important than another.

Discovery, is a means in which unknown /  unconfigured hosts turn on (imagine a company buys 100 new bare metal servers) they auto register to foreman.
The user normally has no way to know discovered hosts are available unless visiting the discovered hosts page.

the idea here is simply to allow users to know they have unused resources that can be "claimed" and start the provisioning/configuration process. 
> 
> What is your second question?

based on our notification implementation, multiple users might receive the notification, however, in the event that one use claimed the resources, we currently remove the notification from all users.

this avoids a user thinking he has a resource he can consume clicking on the link realizing the list of currently avail discovered hosts is empty.

Rohoover:
@ohadlevy 

I imagined that there would be some notifications that would be persistent and others that would not be removed unless read. So in this example, the text would no longer be bold if read, but remain unless the global situation amongst all users is resolved.

ohadlevy:
please do not merge, this will be re-written based on the outcome of https://github.com/theforeman/foreman/pull/4339

ohadlevy:
@dLobatog updated this pr based on the changes in https://github.com/theforeman/foreman/pull/4339

lzap:
I will leave this one on @dLobatog then :-)

ohadlevy:
test

ohadlevy:
foreman pr has been merged.

db/seeds.d/80_discovery_ui_notification.rb
```diff
@@ -0,0 +1,17 @@
+# seeds UI notification blueprints that are supported by Disocvery.
+[
+  {
+    group: _('Hosts'),
+    name: 'new_discovered_host',
+    message: _('One or more Hosts has been discovered'),
```
timogoebel: One ore more hosts _have_ been discovered?

timogoebel (changes requested):
... just one thing I noticed during the demo. Not a full review.

ohadlevy:
@timogoebel updated.

app/services/foreman_discovery/ui_notifications/new_host.rb
```diff
@@ -0,0 +1,30 @@
...
+    # and will update notification expiry when new discovered hosts appear.
+    class NewHost < ::UINotifications::Base
+      private
+
+      def create
+        add_notification if update_notifications.zero?
+      end
+
+      def update_notifications
+        blueprint.notifications.mass_update_expiry
```
ohadlevy: typo: should be blueprint.mass_update_expiry

timogoebel (approved):

app/models/host/managed_extensions.rb
```diff
@@ -61,4 +61,9 @@ def delete_discovery_attribute_set
     DiscoveryAttributeSet.destroy_all(:host_id => self.id) if type_changed?
   end
 
+  def update_notifications
+    return if new_record?
+    return unless type_changed?
+    ForemanDiscovery::UINotifications::DestroyHost.deliver!(self)
```
ohadlevy, line 67: this in theory will break running all core tests with foreman's as this extends Host::Managed yet the seeding happens only in discovery, I'm uncertain if this is an issue or not, but if it is we might need to change the way blueprint seeds is loaded in core. (TBH I don't think it should block this pr).

ohadlevy: Actually, scratch that, I did create a exception handling in core, so this will only warn vs raise.

lzap:
I have one request, the most important notification imho for discovery would be when host was autoprovisioned using a rule. Currently we add a host comment "Autoprovisioned via rule named 'webserver'" (see `perform_autoprovisioning` method).

Could you also add this kind of notifications: "One or more hosts were autoprovisioned via rule 'webserver'"? Because these events trigger either completely on background (auto mode) or if you mass auto provision more than one and we only report "Autoprovisioning was successful" via bubble. Getting this kind of notification where you can check which host were provisioned using what rules would be very helpful.

If you prefer you can only have one kind of notification with summary list of rules and hosts provisioned, that'd work too I think.

ohadlevy:
@lzap :+1.  But maybe in another PR?

lzap:
Oh ok, I mind the PR name. But you could rename, I can still hear your voice "it's super easy to send notification" ;-)

lzap:
So is this ready to have a full review and test now? I see the core PR was merged, any obstacles?

ohadlevy:
@lzap from my side its ready

dLobatog:
Works pretty well - thanks @ohadlevy @lzap @Rohoover

lzap:
Thanks I was planning to do it today but great! :-)

PR #331, opened 2017-03-03
Fixes #18791 - ActiveModel::ForbiddenAttributesError fixed

lzap:
When Katello is installed, calling `apply_inherited_attributes` adds extra
field `content_facet_attributes` which is not permitted for hosts. I think we
should permit it in core, as a whole hash, because this parameter is created in
runtime. This patch is a quick and dirty solution that solves blocking issue
that is in 8.0.0 and 8.0.1 releases - provisioning do not work at all with
Katello plugin installed.

Example how it looks like when added_attributes is {"hostgroup_id"=>"1",
"location_id"=>"2", "organization_id"=>"1"}. Then apply_inherited_attributes
returns {"hostgroup_id"=>"1", "location_id"=>"2", "organization_id"=>"1",
"puppet_proxy_id"=>1, "puppet_ca_proxy_id"=>1, "environment_id"=>1,
"compute_profile_id"=>nil, "realm_id"=>nil,
"content_facet_attributes"=>{"kickstart_repository_id"=>nil,
"content_view_id"=>2, "lifecycle_environment_id"=>2, "content_source_id"=>nil}}

### DIFF
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -8,7 +8,7 @@ class ForemanDiscovery::HostConverter
     host.clear_association_cache
     host.type = 'Host::Managed'
 
-    host.attributes = host.apply_inherited_attributes(added_attributes)
+    host.attributes = host.apply_inherited_attributes(added_attributes.to_h)
     host.set_hostgroup_defaults if host.hostgroup_id.present?
 
     # the following flags can be skipped when parameters are set to false


lzap:
Hey @dLobatog I need some attention here, 8.0.0 and 8.0.1 with Katello installed does not work at all, blocking issue. I am not sure how to *properly* fix this, allow the whole hash or particular fields in core?

lzap:
Daniel pointed out @ShimShtein added it in https://github.com/Katello/katello/blob/master/lib/katello/plugin.rb#L162

Looks like this is only for Host::Managed.

lzap:
I tried to copy that line in plugin.rb but with `Host::Discovered` but it did not help surprisingly. This adds depdndency between plugins, Katello now must load after Discovery in order to set this correctly. Not good.

How about to merge this into 8.0-stable branch to allow me to do fixed 8.0.2 version @dLobatog ?

ShimShtein:
Sorry for piggybacking this PR, I am thinking about more permanent fix to this issue.
@dLobatog, I don't have enough experience with strong params, would it be correct to make the following assumptions for `apply_inherited_attributes`:
1. Reject calls to `apply_inherited_attributes` if the parameters have not passed strong params filter
2. Assume everything added by facets is safe (facet is responsible to filter params if it's relevant)

If it is, I'll change the method accordingly.

lzap:
I don't understand (1) exactly or I don't think it's necessary because Rails errors out the moment inherited parameters are assigned (via `attributes=` method). So I think just (2) might be sufficient, just to mark all attributes added as safe, but only in case the hash is actually StrongParam hash and not ordinary hash (if method exists perhaps).

lzap:
Elegant way of doing it might be to create hash wrapper around StrongAttributes that marks all newly created entries as safe (`permit`)?

ShimShtein:
> I don't understand (1) exactly

I am cloning params hash in `apply_inherited_attributes`, so I am loosing the information about which parameters are permitted.

> So I think just (2) might be sufficient

Assuming I can clone StrongParam hash, will I be able to add more permitted params?
example:
``` ruby
params.permit! #from initial filtering

facet_params = { some_info: 'value' } #added by facet
facet_params.permit!
params[:my_facet_attributes] = facet_params

params.permitted? #should return true!
```

lzap:
> I am cloning params hash in apply_inherited_attributes, so I am loosing the information about which parameters are permitted.

You are not actually, `dup` method is implemented, it does not loose any information I think. The same instance is coming out of `apply_inherited_atributes` to my surprise (I was debugging it). The problem is that these guys are not permitted in this context, it is as simple as that. My patch permits them all.

How about instead modifying existing (strong param protected) hash the method would simply return only attributes that should be added. These can be explicitly permitted and then merged into the protected hash? I can see that the `merge` method is implemented, this should work.

I can perhaps do this in Discovery for now (pass in a fake empty hash and permit), so we have at least clean patch for 8.0.2 and sort out it in core later.

lzap:
@dLobatog I have cleaner solution now, updated. Method to_h returns unsafe hash but with all non-permitted field removed, so it is safe. Then this is passed in the method and it can be mass-assigned without any problem! Tested it, works great.

lzap:
@dLobatog alternatively a fix for core as we already have a report that API is broken: https://github.com/theforeman/foreman/pull/4364 (I prefer to fix it in core)

timogoebel:
I'm in favor of merging the core PR. Shall we add it to discovery as well?

ShimShtein:
If we merge the core PR, we shouldn't need any fixes from discovery at all.

dLobatog:
PR is merged, should this be closed @lzap ?

lzap:
The core patch (http://projects.theforeman.org/issues/18791) was merged and fixes the attribute problem, it is scheduled for 1.14.3 release. If you use Katello plugin and Discovery does not work for you, wait for the minor release or apply the patch manually.

PR #332, opened 2017-03-06
Refs #18801 - append example extra options to kexec

lzap:
This is a patch that adds example extra options to kexec templates. The added
option is actually required to exec RHEL5 and older systems.

Please test with:

https://github.com/theforeman/smart_proxy_discovery_image/pull/5


timogoebel:
Merged, thanks @lzap .

PR #335, opened 2017-03-27
fixes #19032 - widget truncates with ellipsis

timogoebel:
`ellipsis` should be defined on the field, not the whole table. The table needs to be fixed for ellipsis to work. `hidden-xs` works now as well as it hides the whole column.


ohadlevy:
thanks @timogoebel can you please add a before/after screenshot?

timogoebel:
Before:
image

image


After:
image
image

lzap:
Thanks works as expected.

PR #337, opened 2017-03-28
Fixes #19051 - Prevent N+1 on FactValue index

tbrisker:
This depends on changes in https://github.com/theforeman/foreman/pull/4390 - test should fail until that gets merged.

### DIFF
--- a/app/models/concerns/fact_value_extensions.rb
+++ b/app/models/concerns/fact_value_extensions.rb
@@ -4,10 +4,5 @@ module FactValueExtensions
   included do
     belongs_to :discovered_host, :class_name => "Host::Discovered", :foreign_key => :host_id
     scoped_search :relation => :discovered_host, :on => :id, :complete_enabled => false, :only_explicit => true, :rename => :discovered_host
-
-    def host
-      return discovered_host if discovered_host.present?
-      super
-    end
   end
 end
--- a/test/unit/fact_value_extensions_test.rb
+++ b/test/unit/fact_value_extensions_test.rb
@@ -5,7 +5,7 @@ class FactValueExtensionsTest < ActiveSupport::TestCase
     test 'returns discovered_host' do
       discovered_host = FactoryGirl.create(:discovered_host, :with_facts, :fact_count => 1)
       fact_value = discovered_host.fact_values.first
-      assert fact_value.discovered_host, fact_value.host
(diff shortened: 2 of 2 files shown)


lzap:
But they dont! :-)

tbrisker:
That's odd... That means the tests in https://github.com/theforeman/foreman_discovery/commit/f91555774b4856492ffa262a12749882906217bc don't really check that correctly

lzap:
Why this should pass only when CSV was merged, I don't see any link between the two. Can you explain to me?

tbrisker:
@lzap there was a change in a previous version of that PR that moved facts from `Host::Managed` to `Host::Base` which would make this workaround redundant, but that led to other issues with permissions and searching so I took it out for now. 
I am curious as to why the tests pass with this change, maybe @dLobatog has an idea since he created the original workaround?

lzap:
Found it, please change this line

    assert fact_value.discovered_host, fact_value.host

to

    assert_equal fact_value.discovered_host, fact_value.host

(or better assert_nil)

Then it will fail. Can you push the change to see if it works, the patch was merged an hour ago.

tbrisker:
Pushed again with a corrected test. This test should now fail until fix is done in core.

lzap:
Does not work with the patch merged in core... We should consider doing our own facts rather than using Puppet Facts.

lzap:
I believe this is related too: https://github.com/theforeman/foreman/pull/3724

lzap:
I filed new PR in core which changes FactValue belonging to Host::Base rather than Host::Managed. This will enable discovery to drop the extra association that is undeeded. https://github.com/theforeman/foreman/pull/4434

lzap:
I will leave this opened for now, release discovery and then get back to this. But I believe this can't be easily fixed for now.

tbrisker:
Opened another attempt at fixing this in core which will hopefully make this PR work as well.

lzap:
Thanks, core looks good but this is not it, we need to simply drop `discovered_host` association and go with normal `host` association. That's the proper way to fix this. Easy change, the only challenge is to get migration right so existing host associations are changed accordingly. Do you want to work on this? Otherwise I will pick it up next week.

lzap:
Replaced by https://github.com/theforeman/foreman_discovery/pull/339

PR #341, opened 2017-04-27
Fixes #19402 - Correct method for bulk host actions

tbrisker:
In order to prevent url length limits, host bulk actions were changed
from GET to POST. This corrects the routes to use the updated method.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,11 +14,11 @@ Rails.application.routes.draw do
         post 'auto_provision'
       end
       collection do
-        get 'multiple_destroy'
+        post 'multiple_destroy'
         post 'submit_multiple_destroy'
-        get  'select_multiple_organization'
+        post  'select_multiple_organization'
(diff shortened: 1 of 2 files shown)


timogoebel (approved):
@tbrisker : Thanks, looks good to me. How about adding some tests that we at least run the code?

Something like this: https://github.com/theforeman/foreman_expire_hosts/commit/0dc42e4748c058a43772fc439c3eb93fae10c25c

tbrisker:
@timogoebel thanks, added a test for these actions.

timogoebel (approved):
Thanks. Works well. Ready to merge once tests pass.

ohadlevy:
thanks @tbrisker

lzap:
Thanks, so was this changed in core recently? This was working previously.

tbrisker:
yes, see http://projects.theforeman.org/issues/18264

PR #342, opened 2017-04-27
fixes #19409 - auto provision now uses anonymous admin

ohadlevy:
with the introduction of Bug fix #16982: CVE-2016-7078 - User with no organizations or locations can see all resources added.
Discovery queries without a current user fail, this fixes that.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -96,10 +96,12 @@ module Api
       def facts
         facts = params['facts']
         state = true
-        @discovered_host = Host::Discovered.import_host(facts)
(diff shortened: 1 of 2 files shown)


tbrisker (approved):

orrabin:
This fixes auto provisioning and the empty subnet.
👍 once tests are green.

orrabin:
Thanks @ohadlevy!

lzap:
So it was broken only for users without org/loc or for all users? If the latter, why our tests did not fail?

orrabin:
@lzap it was broken for all users, our tests didn't fail because we setup a user in the tests which is why these two lines were added to the test:
```
reset_api_credentials
 User.current = nil
```
to match the real behavior in the controller.

PR #343, opened 2017-04-27
Fixes #19414 - Correct action on host form

tbrisker:
Since the host edit form is reused for provisioning, we need to override
the default action which points to the hosts edit form action.

### DIFF
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -95,4 +95,9 @@ module DiscoveredHostsHelper
             :toggle => 'modal',
             :target => "#fixedPropertiesSelector-#{host.id}"})
   end
+
+  def host_path(host)
+    return super unless controller_name == 'discovered_hosts'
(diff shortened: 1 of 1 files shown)


ohadlevy:
@tbrisker I can confirm this works, would you mind adding a test ?

timogoebel:
I believe, this is a far better approach than the JS mess we had in core. However we should fix this properly which won't be possible without changes to core.
@tbrisker : Do you already have something in mind?

lzap:
Since it's too late for bigger changes in 1.15 now, I am going to test this, merge and create separate refactor ticket for this. Also I will add a ticket to add integration tests, I think we might have one already. What a pain this was.

lzap:
The patch works just fine, I am going to merge tomorrow unless you have a better idea.

timogoebel:
@lzap: 👍 from me. Not nice, but fine enough.

lzap:
Thanks a lot @tbrisker with all the help!

PR #346, opened 2017-05-15
Fixes #19313 - autoprovisioning NIC association fixed

lzap:
After discovered host is converted to Managed, associated NICs still have their associations set to the original record from the database (`Host::Discovered`), which led the `host.managed?` statement fail in `tftp_ready?` method in TFTP orchestration. This led to TFTP orchestration to be skipped completely, so I added a statement to set the appropriate assotiation explicitly during auto provisioning.

This is a regression and I cannot find the root cause of this, it worked previously. We have reports that both 1.14 and 1.15 are broken.


timogoebel:
This is a nasty one. Reminds me a lot of e9c626c7258ce405dbe3d95051a5692ac1d87763. Have you tried if a similar approach helps here?

test/test_helper_discovery.rb
```diff
@@ -1,3 +1,7 @@
+root = File.expand_path(File.dirname(__FILE__))
+require "#{root}/factories/discovery_rule_related"
+require "#{root}/factories/discovery_host_related"
```
lzap: This adds possibility to run individual tests:

    be rake test:discovery TEST=/home/lzap/work/foreman_discovery/test/unit/discovered_extensions_test.rb

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -57,8 +57,15 @@ def perform_auto_provision original_host, rule
...
     host.attributes = host.apply_inherited_attributes(hostgroup_id: rule.hostgroup_id)
     host.set_hostgroup_defaults
+    # interfaces are associated with old record which prevents from TFTP orchestration
+    host.primary_interface.host = host
     # save! does not work here
-    host.save ? host : false
+    if host.save
+      host
+    else
+      Rails.logger.error "Auto provisioning failed: #{host.errors.full_messages.to_sentence}"
```
lzap, line 64: This also closes http://projects.theforeman.org/issues/19405

lzap:
Strange error on Jenkins:

```
NoMethodError: undefined method `split' for nil:NilClass
     (NoMethodError)
/var/lib/workspace/workspace/test_plugin_matrix/database/postgresql/ruby/2.0.0/slave/fast/plugin/test/unit/discovered_extensions_test.rb:176
```

timogoebel:
> Strange error on Jenkins:

I believe, this can be fixed by replacing `assert_equal nil, blubb` with `assert_ni blubb`.

lzap:
Adding `refute_nil` there.

lzap:
Indeed this looks like that was the root cause, I did not notice it. Thank to our weak tests we did not catch it at all. I added tests for autoprovisioning which will verify TFTP and DHCP orchestration.

lzap:
Hmmm re-testing this today, it does not seem to fix the issue. If I comment out the `#host.clear_association_cache` then it works. Any idea how to approach this?

timogoebel:
> Any idea how to approach this?

Not at all. I just remember going nuts to come up with this patch.
Maybe we should post to theforeman-dev and ask for help there. Someone might have an idea.

test/test_helper_discovery.rb
timogoebel: Would `require_relative` be an alternative to this?

lzap: I learned something, thanks. It works yeah.

lzap:
@timogoebel since the regression is fatal and folks cant provision via auto or even API, we might want to remove the line you added for now and try to find better solution later?

To be honest, I don't know why you clear all associations when becomes is overloaded in Host::Base and does exactly that for you:

```ruby
    def becomes(*args)
      became = super
      became.drop_primary_interface_cache
      became.drop_provision_interface_cache
      became.interfaces = self.interfaces
      became
    end
```

timogoebel:
@lzap : I'd be ok with dropping the line. The overloaded becomes method does not help, might even make things worse.

The issue is iirc, that `host != host.primary_interface.host`. Clearing host's cache doesn't help.

lzap:
Ok added more tests including check for subnet.tftp in auto provisioning and fixed factories so they now work with or without `TEST=` option in rake. I cannot reproduce the case you were adding the line for, so I am removing it.

lzap:
I don't understand these, I have `refute_nil` there:

```
NoMethodError: undefined method `split' for nil:NilClass
     (NoMethodError)
/var/lib/workspace/workspace/test_plugin_matrix/database/postgresql/ruby/2.4/slave/fast/plugin/test/unit/discovered_extensions_test.rb:176  
```

lzap:
Any ETA @timogoebel or @tbrisker this is a nasty regression in both 1.14 and 1.15. Thanks!

lzap:
Oh wait tests are still red.

timogoebel:
@lzap : I want to check this out before lunch, hang on...

test/factories/discovery_host_related.rb
```diff
@@ -1,24 +1,26 @@
-FactoryGirl.define do
-  factory :discovered_host, class: 'Host::Discovered' do
-    sequence(:name) { |n| "host#{n}" }
-    sequence(:ip)   { |n| IPAddr.new(n, Socket::AF_INET).to_s }
-    sequence(:mac)  { |n| "02:23:45:67:" + n.to_s(16).rjust(4, '0').insert(2, ':') }
+unless FactoryGirl.factories.registered?(:discovered_host)
```
timogoebel: What is this good for?

lzap: So when you run tests via `TEST=xyz` it loads files in a different way and factory_girl was giving an error - Factory already registered. I could not find any other way than doing this, I want to be able to run all tests and a particular test.

lzap:
Thanks, do you also see a failure like Jenkins?

test/unit/discovery_attribute_set_test.rb
```diff
@@ -1,4 +1,5 @@
-require 'test_helper'
```
timogoebel, line 1: If we do this, we can remove the file `test/test_helper.rb`. It isn't any good anymore.

timogoebel (changes requested):

test/unit/discovery_attribute_set_test.rb
lzap: I actually can't that does not work, this requires `test_helper` from Foreman and not discovery. It's a mess.

lzap: And core `test_helper` required plugins `test_helper`. You need to be this explicit.

timogoebel: Actually, https://github.com/theforeman/foreman_plugin_template doesn't have a `test/test_helper.rb`at all. So this should be possible.

How about doing this in a separate PR to get the actual fix in fast?

lzap:
Incorporated your patch and added refute_nils everywhere.

lzap:
I think I can remove `test_helper.rb` now, let's see what Jenkins does with this version.

lzap:
Might want to split this into two PRs, one is the line change and the other is test improvement.

lzap:
Or just two commits.

timogoebel:
> Or just two commits.

Yep. Makes backporting easier.

What's still missing is the removal of `test/test_helper.rb`

lzap:
Split into two commits, removed the `test_helper.rb` leftover. If it passes locally for you (green for me) let's merge it and if develop is still broken I will create another PR to fix just jenkins tests.

timogoebel (approved):

timogoebel:
Merged, thanks @lzap.

PR #347, opened 2017-05-19
fixes #19594 - replace assert_equal with assert_nil

timogoebel:
@lzap: This should get develop green in ci again.


timogoebel:
Closing in favor of #346.

PR #350, opened 2017-06-01
Fixes #19412,#19732 - kexec path and seed fixed

lzap:
Two issues fixed, both needed patches for 9.1 release.

First, Katello made a change in boot_files_uri method and requires host to be set, otherwise it returns original media URL which is blank. Adding host parameter fixes that.

Second, due to taxonomy changes, multiple seeding was creating multiple templates records as we do not validate the records allowing multiple entries with same name.

These two are remaining items for 9.1: http://projects.theforeman.org/versions/188

@dLobato @ares

### DIFF
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -32,7 +32,8 @@ module Host::ManagedExtensions
     # continue. If flash is implemented for new hosts (http://projects.theforeman.org/issues/10559)
     # we can report the error to the user perhaps.
     true
-  rescue ::Foreman::Exception
+  rescue ::Foreman::Exception => e
+    Foreman::Logging.exception("Unable to reboot", e)
     true
   end
 
@@ -44,11 +45,15 @@ module Host::ManagedExtensions
     template = provisioning_template(:kind => 'kexec')
     raise ::Foreman::Exception.new(N_("Kexec template not associated with operating system")) unless template
     @host = self
-    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture)
-    json = unattended_render(template)
-    old.becomes(Host::Discovered).kexec json
+    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
+    # try to parse JSON and error out early
+    json = JSON.parse(unattended_render(template))
+    ::Foreman::Exception.new(N_("Kernel kexec URL is invalid: '%s'"), json['kernel']) unless json['kernel'] =~ /\Ahttp.+\Z/
+    ::Foreman::Exception.new(N_("Init RAM kexec URL is invalid: '%s'"), json['initrd']) unless json['initrd'] =~ /\Ahttp.+\Z/
+    old.becomes(Host::Discovered).kexec json.to_s
     true
-  rescue ::Foreman::Exception
+  rescue ::Foreman::Exception => e
+    Foreman::Logging.exception("Unable to kexec", e)
     true
   end
 
--- a/test/unit/managed_extensions_test.rb
+++ b/test/unit/managed_extensions_test.rb
@@ -1,0 +1,76 @@
+require 'test_plugin_helper'
+
+class ManagedExtensionsTest < ActiveSupport::TestCase
+  class StubHost < ApplicationRecord
+    include Host::ManagedExtensions
+
+    # prevent from Could not find table exception
+    def self.table_name
+      'hosts'
+    end
+  end
+
+  setup do
+    set_default_settings
+
+    @host = StubHost.new
+    @host.legacy_api = false
+    @host.type = "Host::Discovered"
+    @host.stubs(:old).returns(@host)
+    @facts = {}
+    @host.stubs(:facts).returns(@facts)
+    @post_queue = mock('Post Queue')
+    @host.stubs(:post_queue).returns(@post_queue)
+    @operatingsystem = mock('OS')
+    @host.stubs(:operatingsystem).returns(@operatingsystem)
+    @host.stubs(:provisioning_template).returns('A template')
+    @host.stubs(:medium).returns('http://a_medium')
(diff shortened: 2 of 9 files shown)


lzap:
Tests are failing because of https://github.com/theforeman/foreman/pull/4545 merged in core but I have no idea why @ares @dLobatog ?

lzap:
Example of a failing test: Submit a discovered host:

```ruby
  def test_update_inheritance
    host = Host::Discovered.import_host(@facts)
    hostgroup = setup_hostgroup(host)
    put :update, {
      commit: 'Update',
      id: host.id,
      host: {
        name: 'mytest',
        hostgroup_id: hostgroup.id
      }
    }, set_session_user_default_manager

    # this fails because host was not updated at all - 
    actual = Host.find(host.id)
end
```

The error is: `Failed to save: Owner There is no owner with id 135138680 and type User`. The owner of this host is set to `foreman_admin` account and the validator does not see this associated record.

app/models/host/managed_extensions.rb
```diff
@@ -32,7 +32,8 @@ def setReboot
     # continue. If flash is implemented for new hosts (http://projects.theforeman.org/issues/10559)
     # we can report the error to the user perhaps.
     true
-  rescue ::Foreman::Exception
+  rescue => e
```
dLobatog: What are some common exceptions that could happen here?

app/models/host/managed_extensions.rb
```diff
@@ -44,11 +45,15 @@ def setKexec
     template = provisioning_template(:kind => 'kexec')
     raise ::Foreman::Exception.new(N_("Kexec template not associated with operating system")) unless template
     @host = self
-    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture)
-    json = unattended_render(template)
-    old.becomes(Host::Discovered).kexec json
+    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
+    # try to parse JSON and error out early
+    json = JSON.parse(unattended_render(template))
+    ::Foreman::Exception.new(N_("Kernel kexec URL is invalid: '%s'"), json['kernel']) unless json['kernel'] =~ /^http/
```
dLobatog: Probably matching on `/\Ahttp.+\Z/` is a better idea, to avoid multiline matching:

e.g: json["kernel"] equals to 

```
http://someurl
somerandomcommand
```

is matched by `/^http` whereas `/\Ahttp.+\Z` ensures no extraneous strings

app/models/host/managed_extensions.rb
lzap: I just wanted to catch them all, I had nothing in particular in mind. Shall I revert it?

app/models/host/managed_extensions.rb
lzap: Ok, just for your information, these exception does not rollback transaction, they are more informative.

app/models/host/managed_extensions.rb
dLobatog: Generally I prefer to only handle exceptions that are expected, so that really unexpected things that we should fix and handle in some way are reported to us as soon as they're seen.

lzap: Ok, amended.

dLobatog:
[test] , the job test_plugin_pull_request still installed npm 5.0, I changed it to npm@'<5.0.0'

lzap:
Tests seems allright, added testing for extension, RTM?

app/models/host/managed_extensions.rb
```diff
@@ -44,11 +45,15 @@ def setKexec
...
-    json = unattended_render(template)
-    old.becomes(Host::Discovered).kexec json
+    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
+    # try to parse JSON and error out early
+    json = JSON.parse(unattended_render(template))
+    ::Foreman::Exception.new(N_("Kernel kexec URL is invalid: '%s'"), json['kernel']) unless json['kernel'] =~ /\Ahttp.+\Z/
+    ::Foreman::Exception.new(N_("Init RAM kexec URL is invalid: '%s'"), json['initrd']) unless json['initrd'] =~ /\Ahttp.+\Z/
+    old.becomes(Host::Discovered).kexec json.to_s
     true
   rescue ::Foreman::Exception
```
dLobatog: In order to log the exception this line would need to be `::Foreman::Exception => e`

app/models/host/managed_extensions.rb
dLobatog: The other one was reverted to only rescue `Foreman::Exception`, this one wasn't touched

test/unit/managed_extensions_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    @facts['discovery_kexec'] = "Kexec version X.Y.Z"
+    @host.stubs(:provisioning_template).returns(nil)
+    @post_queue.expects(:create).with(has_entry(:action, [@host, :setReboot])).once
+    @host.queue_reboot
+  end
+
+  test "queue_reboot enques kexec command" do
+    @host.stubs(:type_changed?).returns(true)
+    @host.stubs(:new_record?).returns(false)
+    @host.id = 130513
```
dLobatog, line 59: What is this id needed for, when is Host.find called?

dLobatog (changes requested):
There's a tiny problem with the setKexec method, the exception handling wouldn't work (see inline). 👍 to everything else, thanks!

test/unit/managed_extensions_test.rb
lzap: This is a fake AR class, I called `new` and since it wasn't yet stored, there is no ID yet. I need to set this explicitly because of `self.id` code inside tested class. Fixed exceptions good catch, pushed!

dLobatog (approved):

dLobatog:
Thanks @lzap ! Finally tests are green again!

PR #351, opened 2017-06-05
Fixes #19924 - Consistently use 'default manager' in tests

dLobatog:
Tests were using 'users(:admin)' to create certain objects like Host.
The controller was using a newly created used with 'Default Manager'
permissions to act on these objects, but they are not always visible to
non admin users.

Particularly most tests failed because host.owner is users(:admin)
('secret_admin') which is not visible to a user with Default Discovery
Manager permissions

### DIFF
--- a/test/functional/discovered_hosts_controller_test.rb
+++ b/test/functional/discovered_hosts_controller_test.rb
@@ -127,29 +127,31 @@ class DiscoveredHostsControllerTest < ActionController::TestCase
   end
 
   def test_update_inheritance
-    host = Host::Discovered.import_host(@facts)
-    hostgroup = setup_hostgroup(host)
-    put :update, {
-      commit: 'Update',
-      id: host.id,
-      host: {
-        name: 'mytest',
-        hostgroup_id: hostgroup.id
-      }
-    }, set_session_user_default_manager
+    as_default_manager do
+      host = Host::Discovered.import_host(@facts)
+      hostgroup = setup_hostgroup(host)
+      put :update, {
+        commit: 'Update',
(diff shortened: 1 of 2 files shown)


test/functional/discovered_hosts_controller_test.rb
```diff
@@ -127,6 +127,7 @@ def test_edit_form_submit_parameters
   end
 
   def test_update_inheritance
+    User.current = User.unscoped.find(set_session_user_default_manager[:user])
```
lzap: Can you refactor the helpers and make new `user_default_manager` and use it in `set_session_user_default_manager`? The same for reader, this is little bit clunky :-)

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -201,48 +202,52 @@ def test_reboot_error
   end
 
   def test_auto_provision_success
+    User.current = User.unscoped.find(set_session_user_default_manager[:user])
```
lzap: This needs a helper named `as_default_manager` with block.

lzap: Of course store the old value and put it back, just in case. Alternatively set it in setup and remove in teardown methods, if you like this more.

lzap:
Many thanks for help with this.

dLobatog:
@lzap Updated, I am trying to prove everything's fine with this job : http://ci.theforeman.org/job/test_plugin_matrix/3229/ , locally at least everything is green.

lzap:
All green locally, thanks a lot!

PR #352, opened 2017-06-05
Fixes #19944 - Move default role permissions to seeds

ehelms:
Happy to use a new issue if you feel that's warranted. I still hit this issue even after the migration was moved. Perhaps this has to do with some sort of state reloading but the base cause of the failure can occur again. If ever the Role or Filter models in Foreman are updated, without the migration for them having been run first (which in the current code won't happen) then this will call a function that could call a code path that references non-existent database fields and error out in the same way. I feel these are better served being in the seeds because they are ensuring a consistent data state and not manipulating the tables themselves.

### DIFF
--- a/db/migrate/20160818091421_add_permissions_from_default_roles.rb
+++ b/db/migrate/20160818091421_add_permissions_from_default_roles.rb
@@ -1,12 +1,6 @@
 class AddPermissionsFromDefaultRoles < ActiveRecord::Migration
   def up
-    default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
-    ["Discovery Reader", "Discovery Manager"].each do |role_name|
-      role = Role.find_by_name(role_name) || next
-      default_permissions[role_name].each do |permission|
-        role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
-      end
-    end
+    # Moved to seeds see http://projects.theforeman.org/issues/17585
   end
 
   def down
--- a/db/seeds.d/90_add_permissions_from_default_roles.rb
+++ b/db/seeds.d/90_add_permissions_from_default_roles.rb
@@ -1,0 +1,10 @@
+default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
+
+["Discovery Reader", "Discovery Manager"].each do |role_name|
+  role = Role.find_by_name(role_name) || next
+  default_permissions[role_name].each do |permission|
+    role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+  end
+  role.update_attributes :origin => "discovery", :description => "Discovery plugin built-in role"
+  role.save!
+end


lzap:
Eric, we had issues with these in the past, could you create new ticket for me and also paste whole transaction with the error there? Or tell me how to reproduce, do I need to upgrade from an older version to 1.15? In the meantime we will fix unit tests.

ehelms:
You are not the only plugin that hits this but you hit it in a different way than others. Give (https://github.com/theforeman/foreman/pull/4569#issuecomment-306455098) a look see and I can add more if you need it around a scenario to use. 

The transaction and error are the same as the original ticket intended to solve - it just didn't actually solve the problem. Further, this problem could re-arise at any time IF the code ever includes reference to a database attribute not yet used and someone adds the discovery plugin before migrating the Foreman side. I believe that this should just not be in the migrations as it has more potential to break than just being in seeds.

dLobatog (approved):
👍 from me, it's really seeding data (permissions and roles) so I agree it should be a seed and leave migrations just for schema changes.

@lzap I saw you commented 

> Eric, we had issues with these in the past 

What issues?

@ehelms New ticket number would be appreciated

ehelms:
Updated with new issue

dLobatog:
[test] , the job test_plugin_pull_request still installed npm 5.0, I changed it to npm@'<5.0.0' . The job isn't managed by foreman-infra AFAICT

dLobatog:
[test] , removed Ruby 2.0 from the job

lzap:
There is one change this brings, when this was migration, it was only added once. We decided to do it this way because Discovery permissions were utterly broken and we added an upgrade note we are adding all missing permissions and users should review this.

Your change makes this a seed operation, which means any upgrade or even minor update will trigger adding those back. If user wanted to remove some permission, we would add it back, which can be dangerous - security.

Now since 1.15 includes role locking mechanism (it was introduced earlier than 1.15 I think - does not matter), I suggest we can simple set these roles to be locked the same way as other core roles. Users can still clone these roles and modify the copies at will. This is a precedent, no other plugin does that AFAIK, so I want to hear opinions from @dLobatog or @tbrisker. My proposal is to add this to the seed:

    role.update_attributes :origin => "discovery", :description => "Discovery plugin built-in role"
    role.save!

The role model code is a bit unclear, but basically if you set origin to something, it will make the role locked and users are unable to edit it, except cloning. Since introduction of the Plugin Roles API (see http://projects.theforeman.org/issues/17954) the layer is now capable of doing such a thing automatically for each individual plugin but since we are not there yet, I suggest to go with explicit seed for now.

xprazak2:
Yes, it is exactly like that - if role has origin, then it is locked. We will be locking plugin roles anyway, so if doing it in seeds helps you, go ahead.

dLobatog:
I agree with that too, it makes sense that if the role is given to Foreman users - by Foreman core or a plugin - it should be locked for modifications, users can still clone their roles to make their own.

lzap:
@ehelms would you mind adding the two lines, then I am merging! Thanks.

ehelms:
@lzap Added the two lines from your comment please check they are correct and in the right place

lzap:
Yeah thanks a lot.

PR #358, opened 2017-06-30
Fixes #20170 - Fix the UI for delete modal

rabajaj0509:


ohadlevy (commented):
Thanks, can you please provide a before/after screenshot?

dLobatog:
[test]

rabajaj0509:
Before :  
delete_modal

After:
deletemodalfix

rabajaj0509:
i guess, this looks good but as the number of entries will increase the 'organization' might collide with the submit button. Can someone give a feedback on this ?

dLobatog (approved):
@rahulbajaj0509 If try to delete many, the modal will become larger but it will not collide with the Submit button. You can reproduce the situation already in the Hosts page, trying to remove many hosts at the same time.

dLobatog:
Thanks @rahulbajaj0509 !

Rohoover:
Can we use the Inline messaging warning instead of the current style? The current style doesn't really exist anywhere else with any consistency. 

http://www.patternfly.org/pattern-library/communication/inline-notifications/

rabajaj0509:
@Rohoover thanks for the input, would surely make the changes and raise a pr :)

PR #361, opened 2017-07-10
Fixes #20294,#20259 - taxonomy via fact fixed, scripts

lzap:
I have an improved script to do ad-hoc fake hosts discovery. It will become part of the main git repository.

Old scripts can be removed, they are not used anymore (this was from first generation discovery).


lzap:
You will find this useful @rahulbajaj0509 :-) Please review.

rabajaj0509:
@lzap , the script was indeed informative :)

Concerned with one thing here:
The fake discovered host is created in the organization that is alphabetically first among all the organizations created and does not get created in the organization i want, is this the correct functionality ?

lzap:
Yeah, I can perhaps add custom org/loc option, lemme do that. Unfortunately the way it works is super-clunky, read this: https://theforeman.org/plugins/foreman_discovery/9.1/index.html#3.1.3OrganizationsandLocations

lzap:
Try now! New options org/loc are present. Oh when I was testing this it was not working, reported a bug and fixed it here:

http://projects.theforeman.org/issues/20294

rabajaj0509:
@lzap Thanks 👍  looks good to me !

lzap:
Adding a test, @dLobatog this would be also in 9.1.2 if possible.

test/unit/host_discovered_test.rb
```diff
@@ -42,6 +42,18 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert_equal subnet, host.primary_interface.subnet
   end
 
+  test "should setup subnet with org and loc set via settings" do
```
dLobatog: Please rename this one, two tests have the same name so only one runs I think

dLobatog (changes requested):

test/unit/host_discovered_test.rb
lzap: Done

dLobatog (approved):

dLobatog:
Thanks @lzap !

PR #363, opened 2017-07-12
Fixes #20290 - kexec now produces valid JSON output

lzap:
This is unfortunate, kexec is not working at all in 1.15/9.1.1 because I added `to_s` method call which creates ruby string representation instead of `to_json` which is implicit. I did not spot this via Template preview feature because kexec templates are not rendered using normal stack, we call renderer explicitly - this needs improval in the future.

This one liner fixes it. To test this, you need to initiate provisioning via kexec and validate JSON or let a VM to really kexec. I will do minor release for this later on.

### DIFF
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -50,7 +50,7 @@ module Host::ManagedExtensions
     json = JSON.parse(unattended_render(template))
     ::Foreman::Exception.new(N_("Kernel kexec URL is invalid: '%s'"), json['kernel']) unless json['kernel'] =~ /\Ahttp.+\Z/
     ::Foreman::Exception.new(N_("Init RAM kexec URL is invalid: '%s'"), json['initrd']) unless json['initrd'] =~ /\Ahttp.+\Z/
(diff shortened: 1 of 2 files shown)


ohadlevy:
@lzap can you please add a test?

lzap:
Done.

lzap:
@orrabin this is the blocker :-)

lzap:
To see the JSON I recommend this patch as well: https://github.com/theforeman/foreman_discovery/pull/365

lzap:
Blocker, must merge for 10.0 please @dLobatog @ares thanks.

orrabin:
👍 works for me.
This can't be applied with https://github.com/theforeman/foreman_discovery/pull/365 without conflicts so I added the relevant part that wrote to the log to see the json.
This could be checked without even seeing the json, without the pr the logs show:
` Unable to perform kexec on mac52540055568a | RestClient::InternalServerError: 500 Internal Server Error
`
With this pr the error is gone.

orrabin:
Thanks @lzap!

PR #366, opened 2017-07-24
Fixes #20387 - Clean the redundant code for auto complete search

rabajaj0509:
This commit shall improve the usability by removing the extra spaces prefixed to the search string.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -145,21 +145,6 @@ class DiscoveredHostsController < ::ApplicationController
     redirect_to(discovered_hosts_path)
   end
 
-  def auto_complete_search
-    begin
-      @items = Host::Discovered.complete_for(params[:search])
-      @items = @items.map do |item|
-        category = (['and','or','not','has'].include?(item.to_s.sub(/^.*\s+/,''))) ? 'Operators' : ''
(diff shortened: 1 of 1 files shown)


app/models/discovery_rule.rb
```diff
@@ -26,6 +27,10 @@ class DiscoveryRule < ApplicationRecord
   scoped_search :on => :enabled
   scoped_search :relation => :hostgroup, :on => :name, :complete_value => true, :rename => :hostgroup
 
+  def remove_extra_spaces
+     self.search = search.strip
```
ohadlevy: how about to fix it on the search autocompleter, so its fixed for all screens that use search?

rabajaj0509:
related pr : https://github.com/theforeman/foreman/pull/4693

lzap:
Tests finally fixed [test].

lzap:
Yeah we had it there because `model_of_controller` was not yet available in core, we added it later.

rabajaj0509:
Not sure why the tests are failing here, when i see the test results it says - `no failures`.

lzap:
Looks like an issue with CI - "Step ‘Publish JUnit test result report’ failed: no workspace left". One more try and then I will merge [test]

lzap:
I can confirm search works for both discovered index and rule edit form, but I still see spaces in the dropdown. When you save or confirm the search, the extra space is trimmed tho. Merging thanks.

PR #370, opened 2017-08-11
Fixes #20564 - ignore locking in migration

lzap:
Until there is a way for plugin API to do this for us, we need to ignore locking to prevent this error:

Validation failed: Filters.role is locked for user modifications

### DIFF
--- a/db/seeds.d/90_add_permissions_from_default_roles.rb
+++ b/db/seeds.d/90_add_permissions_from_default_roles.rb
@@ -1,12 +1,14 @@
+# frozen_string_literal: true
 default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
 
 ["Discovery Reader", "Discovery Manager"].each do |role_name|
-  role = Role.find_by_name(role_name) || next
-  default_permissions[role_name].each do |permission|
-    role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+  Role.ignore_locking do
+    role = Role.unscoped.find_by_name(role_name) || next
+    default_permissions[role_name].each do |permission|
+      role.add_permissions(permission) unless role.permission_names.include?(permission.to_sym)
+    end
+    role.origin = 'discovery'
+    role.description = 'Discovery plugin built-in role'
+    role.save
   end
-  role.ignore_locking do |r|
-    r.update_attributes :origin => "discovery", :description => "Discovery plugin built-in role"
-  end
-  role.save!
 end


lzap:
Random failures hmmm [test]

lzap:
Test [test]

lzap:
So previous [test] failed with:

	Test Result (3 failures / +3)
HostTest::location or organizations are not enabled.test_0081_non-admin user with edit_hosts permission can update interface
HostsControllerTest.test_0077_select multiple action with not exists host_ids should redirect to hosts page
HostIntegrationTest::edit page.test_0002_user without edit_params permission can save host with params

Trying again.

lzap:
One [test] failed, a different one of course.

Test Result (1 failure / +1)
HostJSTest::edit page.test_0002_choosing a hostgroup does not override other host attributes

lzap:
If this turns out to be random failure again, let's merge this one then? We have couple of others that would need to be in 10.0.0 @dLobatog @ares ...

ares:
LGTM, not tested yet. Note that this won't benecessary soon, plugin DSL will always update role permissions according to what's defined in engine.rb.

lzap:
@dLobatog mind quick merge here? this only adds `Role.ignore_locking` to prevent "Template is locked" errors during migration. Foreman 1.16 is no longer affected, but we need this for 1.15 which I will be cherry picking into, thanks!

timogoebel (approved):

db/seeds.d/90_add_permissions_from_default_roles.rb
```diff
@@ -1,12 +1,14 @@
...
 ["Discovery Reader", "Discovery Manager"].each do |role_name|
-  role = Role.find_by_name(role_name) || next
-  default_permissions[role_name].each do |permission|
-    role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+  Role.ignore_locking do
+    role = Role.find_by_name(role_name) || next
+    default_permissions[role_name].each do |permission|
+      role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+    end
+    role.ignore_locking do |r|
```
tbrisker: You don't need to ignore locking twice - if you do it at the class level it will be ignored for all instances.

db/seeds.d/90_add_permissions_from_default_roles.rb
```diff
@@ -1,12 +1,14 @@
 default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
 
 ["Discovery Reader", "Discovery Manager"].each do |role_name|
-  role = Role.find_by_name(role_name) || next
-  default_permissions[role_name].each do |permission|
-    role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+  Role.ignore_locking do
+    role = Role.find_by_name(role_name) || next
```
dLobatog: `Role.unscoped.` no user is set here so I think `Role.find_by_name` will be scoped to not show everything in all orgs/locs

dLobatog: Role.unscoped. no user is set here so I think Role.find_by_name will be scoped to not show everything in all orgs/locs

db/seeds.d/90_add_permissions_from_default_roles.rb
```diff
@@ -1,12 +1,14 @@
...
+    default_permissions[role_name].each do |permission|
+      role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
```
dLobatog: Nitpick, since it's calling `.save!` on line 12, use `add_permissions(permission)` here - `add_permissions!` will save it once, then twice at the end of the loop.

db/seeds.d/90_add_permissions_from_default_roles.rb
```diff
@@ -1,12 +1,14 @@
 default_permissions = Foreman::Plugin.find("foreman_discovery").default_roles
```
dLobatog, line 2: nitpick, all strings in here can safely be frozen, you can add

```
# frozen_string_literal: true
```

at the beginning of the file for better performance & thread safety (I know it's a little ridiculous for a loop with 2 objects but it's easy enough)

dLobatog (changes requested):
@lzap Check out the few comments inline that are easy to fix, 👍 if tests pass.

lzap:
Ok amended all remarks thanks for such a triple review that's nice :-)

db/seeds.d/90_add_permissions_from_default_roles.rb
```diff
@@ -1,12 +1,13 @@
...
 ["Discovery Reader", "Discovery Manager"].each do |role_name|
-  role = Role.find_by_name(role_name) || next
-  default_permissions[role_name].each do |permission|
-    role.add_permissions!(permission) unless role.permission_names.include?(permission.to_sym)
+  Role.ignore_locking do
+    role = Role.unscoped.find_by_name(role_name) || next
+    default_permissions[role_name].each do |permission|
+      role.add_permissions(permission) unless role.permission_names.include?(permission.to_sym)
+    end
+    role.update_attributes :origin => "discovery", :description => "Discovery plugin built-in role"
```
dLobatog: `update_attributes` already saves the role, no need to call `.save!` after. Set all attributes then save to ensure it's saved only once:

```ruby
role.add_permissions(permission)
role.origin = 'discovery'
role.description = 'discovery plugin builtin'
role.save
```

(sorry GitHub saved the comment but didn't publish it with the review)

lzap:
@dLobatog just like that?

dLobatog (approved):

timogoebel:
Merged, thanks @lzap, @dLobatog and @tbrisker.

PR #373, opened 2017-09-08
Refs #20564 - Ignore locking in migration

bastilian:
`rake db:migate` still failed with "Validation failed: Filters.role is locked for user modifications".


lzap:
Thanks, looks like I was testing this with already populated db. We have no tests for this, so merging right away.

PR #377, opened 2017-09-17
Fixes #19448 - Show notification if new host discovered

tbrisker:
after marking discovery notification as read, new hosts discovered do
not show a new notification since the read marker is kept.

### DIFF
--- a/app/services/foreman_discovery/ui_notifications/new_host.rb
+++ b/app/services/foreman_discovery/ui_notifications/new_host.rb
@@ -7,11 +7,14 @@ module ForemanDiscovery
       private
 
       def create
-        add_notification if update_notifications.zero?
+        User.as_anonymous_admin do
+          blueprint.notifications.any? ? update_notifications : add_notification
+        end
       end
 
       def update_notifications
         blueprint.mass_update_expiry
+        blueprint.mass_set_seen(false)
       end
 
       def add_notification
--- a/test/unit/ui_notifications/new_host_test.rb
+++ b/test/unit/ui_notifications/new_host_test.rb
@@ -18,6 +18,15 @@ class NewHostNotificationTest < ActiveSupport::TestCase
     assert_not_equal expired_at, blueprint.notifications.first.expired_at
   end
 
+  test 'new discovered hosts should show new notification' do
+    host1 = FactoryBot.create :discovered_host
+    ForemanDiscovery::UINotifications::NewHost.deliver!(host1)
+    blueprint.notifications.first.notification_recipients.update_all(seen: true)
+    host2 = FactoryBot.create :discovered_host
+    ForemanDiscovery::UINotifications::NewHost.deliver!(host2)
+    refute blueprint.notifications.first.notification_recipients.first.seen
+  end
+
   private
 
   def blueprint


app/services/foreman_discovery/ui_notifications/new_host.rb
```diff
@@ -7,7 +7,13 @@ class NewHost < ::UINotifications::Base
       private
 
       def create
-        add_notification if update_notifications.zero?
+        if update_notifications.zero?
+          add_notification
+        else
+          Notification.where(notification_blueprint: blueprint).each do |notification|
+            notification.notification_recipients.update_all(seen: false)
```
ohadlevy: What about the expiry? They could be expired already.

IMHO you should update them both (seen and expiry), and then maybe the right fix is in core

ohadlevy (changes requested):

app/services/foreman_discovery/ui_notifications/new_host.rb
tbrisker: update_notifications already does the mass_update_expiry. I don't know if we should always also mark as unseen when updating the expiry?

ohadlevy: Then why loop twice? Imho just change it at the same query / scope.

tbrisker: I don't mind moving this into `mass_update_expiry`, but do you think we should always mark all as unread when updating expiry?

ohadlevy: Can we chain update_all? Eg is it a scope?

tbrisker: what do you mean by chain it? it return the number of rows modified, and we can't do something like `blueprint.notifications.notification_recipients`. We could add `has_many_through` on the blueprint to get all of the recipients but we'd still need to mark them as unread.

timogoebel: How about we add a "mark as unread" call to core?

lzap:
We can hopefully [test] now.

app/services/foreman_discovery/ui_notifications/new_host.rb
lzap: I think the notification stack should have `add_or_update_notification` call. Anyway, there is a random failure, please rebase on develop.

tbrisker: Opened https://github.com/theforeman/foreman/pull/4970 to fix this in core, once that is merged i will update this pr accordingly.

lzap:
Core was merged, what is the status please?

tbrisker:
Tests were failing, should be fixed now.

lzap:
I am testing this but I see no notifications when I discover a host or even delete one. All I see in logs is `Cache Hit: notification, reading cache for notification-4` can you take a look please? Does it work on your end? Might be misconfiguration of notifications.

tbrisker:
sounds like this has to do with the cache not being refreshed - @ohadlevy any suggestion on best approach? 
I think it's outside the scope of this PR, but perhaps we should add another method in core to invalidate the cache?

lzap:
Ping @ohadlevy @tbrisker what is the status?

tbrisker:
@lzap sorry, haven't had much time to look into this... I'll try to put it back on my list.

rabajaj0509:
@tbrisker any updates? if you are too busy for this, please guide me I can take this over.

tbrisker:
@rahulbajaj0509 feel free to take over, i don't see myself getting to it soon. Basically, the issue is that if you mark discovered hosts notification as read, then another host is discovered, the notification will stay marked as read.

lzap:
I am gonna close the PR then, please reopen if you get back to it @tbrisker thanks!

PR #379, opened 2017-09-20
Fixes #20215 - Fix the accordian for Expand All button

rabajaj0509:
This patch ensures that :
a) The `Expand All` button also performs collapse.
b) The accordian sign changes when `Expand All` button is clicked.
c) Also, the button text is changed from `Exapand All` to `Collapse All` as required.


lzap:
Works like a charm, no tests for this, merging. Thanks!

PR #380, opened 2017-09-26
Fixes #20549 - Displays readable message in reboot API action

rabajaj0509:


rabajaj0509:
Looks like :

screenshot from 2017-09-26 14-11-49

lzap:
LGTM [test]

lzap:
We can hopefully [test] now.

lzap:
[test]

rabajaj0509:
@lzap we need to merge https://github.com/theforeman/foreman/pull/4875 for the tests to pass for this pr :)

lzap:
I am getting this failure when I test this with the core patch:

```
  1) Failure:
DiscoveredHostsControllerTest#test_reboot_all_failure [/home/lzap/work/foreman_discovery/test/functional/discovered_hosts_controller_test.rb:287]:
--- expected
+++ actual
@@ -1 +1,2 @@
-"Errors during reboot: macaabbccddeeff: failed to reboot"
+# encoding: US-ASCII
+"{:message=&gt;&quot;Errors during reboot: macaabbccddeeff: failed to reboot&quot;, :host_details=&gt;[{:name=&gt;&quot;macaabbccddeeff&quot;, :error=&gt;&quot;macaabbccddeeff: failed to reboot &quot;}]}"
```

lzap:
Also please rebase onto develop, then tests should pass once you fix the issue there.

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -197,9 +197,7 @@ def reboot_all
         if error_message
           render_error :custom_error,
```
lzap: After discussion with @tstrachota we should create our own discovery RABL which can extend using `extends` keywoard from `custom_error` and then you can use this here on this line to render it. No core change would be required then.

lzap (changes requested):
Please rewrite using custom RABL, @tstrachota might be able to help if you don't know how to do this, I don't know RABL at all :-)

app/controllers/api/v2/discovered_hosts_controller.rb
rabajaj0509: awesome :) will do that ! closing -  theforeman/foreman#4875

rabajaj0509:
@lzap i just realized that there is no need to use `extends`. The `message` field is what we get by `extends` which is an un-formatted form of the error message but since we have `host_details` now, which is a formatted way, we do not need the `message`. 

looks something like:
screenshot from 2017-11-08 14-25-32

lzap:
@orrabin noticed you are doing hammer part review, so taking this as well?

tstrachota:
@rahulbajaj0509 could you please make the error messages more readable? Or add another node with user-friendlier messages. The main point of this issue was to provide clearer info of what hosts failed to reboot and for what reason. I think we should present only the original error. The wrapping exception bears no information here and makes the error less readable. Please see the issue where I describe the expected output.

rahulbajaj0509:
@tstrachota this pr's output looks like this https://github.com/theforeman/foreman_discovery/pull/380#issuecomment-332222116, I think this is what we should do in discovery. The formatting part can be done in hammer-cli-discovery. Is the approach correct?

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -70,19 +70,23 @@ def perform_reboot_all hosts = Host::Discovered.all
...
             overall_errors << error
-            logger.error error
+            host_details << {:name => "#{discovered_host.name}", :error => "#{error}"}
           end
         rescue Exception => e
-          error = "#{discovered_host.name}: #{e.to_s} "
+          error = "#{e.to_s} "
+          if error.include?"[Foreman::WrappedException]:" and error.include?"([Errno"
+            error = error[/]: (.*?) \(\[/, 1]
+          end
```
rabajaj0509: @tstrachota alone `error = error[/]: (.*?) \(\[/, 1]` could have solved our problem of making the error message more readable but I thought it would be good to have the condition as well. The output of the command now looks like:
screenshot from 2018-03-12 19-05-33

rabajaj0509: @tstrachota on thought though, do we really need the `message` node?,  we show the output in hammer-cli-foreman-disocvery using the `host_details`.

tstrachota: I don't think we need the main `message` node in this case. We can keep it there just for the sake of consistency, but I'm also ok with leaving it out.

rabajaj0509: @lzap @tstrachota Ohkay, keeping consistency in mind I have made changes to the pr and the tests are passing locally.

lzap: It's important to keep it for backward compatibility, people have scripts hooked onto this V2 API, we don't change API until V3, so glad you kept it.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -70,19 +70,23 @@ def perform_reboot_all hosts = Host::Discovered.all
...
     overall_errors = ""
+    host_details  = []
 
     if hosts.count > 0
       hosts.each do |discovered_host|
         begin
           unless discovered_host.reboot
             error =  "#{discovered_host.name}: failed to reboot "
             overall_errors << error
-            logger.error error
```
lzap: You are effectively deleting logging of the exception, we want to keep all the errors in log for each individual host which failed to be able to debug problems, Please keep it, or use Foreman exception logger helper method to format it properly.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
lzap: Hmmm can't you do just:

```ruby
msg = if e.respons_to?(:wrapped_exception)
  e.wrapped_exception.message
else
 e.message
end
```

Or similar? This parsing is really bad, I'd rather fix how exceptions are being throwed rather than doing this magic.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -92,6 +96,9 @@ def perform_reboot_all hosts = Host::Discovered.all
 
     if overall_errors.present? || !result
       error_message % overall_errors.strip
+      { :message => error_message % overall_errors.strip,
+        :host_details => host_details
+      }
     end
```
lzap: Weird formatting.

lzap (changes requested):
Nice, please also add a V2 API test to verify it returns proper data back. I don't see any ATM.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
rabajaj0509: @lzap the code snippet that you have provided is really helpful. thanks!

app/controllers/concerns/foreman/controller/discovered_extensions.rb
rabajaj0509: my bad, sorry about that.

rabajaj0509:
@lzap thanks for your valuable input. Once, tests are written, this pr will be ready.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -78,11 +79,18 @@ def perform_reboot_all hosts = Host::Discovered.all
             error =  "#{discovered_host.name}: failed to reboot "
             overall_errors << error
             logger.error error
+            host_details << {:name => "#{discovered_host.name}", :error => "#{error}"}
```
lzap: Useless interpolation.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -78,11 +79,18 @@ def perform_reboot_all hosts = Host::Discovered.all
...
           end
         rescue Exception => e
-          error = "#{discovered_host.name}: #{e.to_s} "
-          overall_errors << error
+          error = "#{e.to_s} "
+          msg = if e.respond_to?(:wrapped_exception)
+                  e.wrapped_exception.message
+                else
+                  e.message
+                end
```
lzap, line 89: This does not appear to be correct, you convert Exception class to String and then for `wrapped_exception` method presence. Lemme explain, in Foreman we have Foreman::Exception and WrappedException defined. These are just exception wrappers with error codes (ERF-XXX-XXX) which allows us to maintain list of codes. Now, WrappedException is used when you want to store the original exception inside, that's why it's useful.

https://github.com/theforeman/foreman/blob/develop/lib/foreman/exception.rb

So if you catch a generic exception, it will either contain message or wrapped_exception.message does it make any sense? So you want to call respond_to? on `e` directly rather than converting it to string.

https://projects.theforeman.org/projects/foreman/wiki/ErrorCodes

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -78,11 +79,18 @@ def perform_reboot_all hosts = Host::Discovered.all
...
+          overall_errors << msg
           logger.error error
+          host_details << {:name => "#{discovered_host.name}", :error => "#{msg}"}
```
lzap: Ditto - interpolation.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -275,7 +275,8 @@ def test_reboot_all_failure
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
     post :reboot_all, params: { }, session: set_session_user_default_manager
     assert_redirected_to discovered_hosts_url
-    assert_equal "Errors during reboot: #{host.name}: failed to reboot", flash[:error]
+    assert_equal "message", eval(flash[:error]).keys[0].to_s
+    assert_equal "host_details", eval(flash[:error]).keys[1].to_s
```
lzap: I'd assert whole hash if possible.

lzap (changes requested):
This is going the right direction.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
rabajaj0509: makes sense. thanks for explaining.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -78,11 +79,17 @@ def perform_reboot_all hosts = Host::Discovered.all
...
-          error = "#{discovered_host.name}: #{e.to_s} "
-          overall_errors << error
-          logger.error error
+          msg = if e.respond_to?(:wrapped_exception)
+                  e.wrapped_exception.message
+                else
+                  e.message
+                end
+          overall_errors << msg
+          logger.error msg
```
lzap: Please use: `Foreman::Logging.exception` helper, see in core. This only puts the message in logs, we want message and backtrace.

test/functional/discovered_hosts_controller_test.rb
lzap: Do I see `eval`? :-)

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -275,7 +275,8 @@ def test_multiple_reboot_failure
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_equal "Errors during reboot: #{host.name}: failed to reboot", flash[:error]
+    require 'pry'; binding.pry;
```
bastilian: Debug code? :) (also on line 289)

mmoll: wat? :)

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -285,7 +286,8 @@ def test_multiple_reboot_error
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).raises("request failed")
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_match(/ERF50-4973/, flash[:error])
+    require 'pry'; binding.pry;
```
mmoll: wat? :)

test/functional/discovered_hosts_controller_test.rb
rabajaj0509: sorry. forgot to remove it after testing.

test/functional/discovered_hosts_controller_test.rb
rabajaj0509: sorry. forgot to remove it after testing. my bad!

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -275,7 +275,7 @@ def test_multiple_reboot_failure
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_equal "Errors during reboot: #{host.name}: failed to reboot", flash[:error]
+    assert_equal "{:message=>\"Errors during reboot: macaabbccddeeff: failed to reboot\", :host_details=>[{:name=>\"macaabbccddeeff\", :error=>\"macaabbccddeeff: failed to reboot \"}]}", flash[:error]
```
bastilian: Still use `#{host.name}`, otherwise these tests would need to be updated again with fixture changes.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -285,7 +285,7 @@ def test_multiple_reboot_error
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).raises("request failed")
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_match(/ERF50-4973/, flash[:error])
+    assert_equal "{:message=>\"Errors during reboot: request failed\", :host_details=>[{:name=>\"macaabbccddeeff\", :error=>\"request failed\"}]}", flash[:error]
```
bastilian: Same as above.

test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -201,4 +201,14 @@ def test_auto_provision_all_no_rule_success
     assert_response :success
   end
 
+  def test_reboot_all_failure
+    disable_orchestration
+    host = discover_host_from_facts(@facts)
+    ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
+    post :reboot_all, params: { }
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert_equal "Errors during reboot: macaabbccddeeff: failed to reboot", show_response["error"]["message"]
```
bastilian: And here. :)

bastilian (changes requested):

test/functional/discovered_hosts_controller_test.rb
rabajaj0509: ah! good catch thanks @bastilian  :)

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -92,6 +99,10 @@ def perform_reboot_all hosts = Host::Discovered.all
 
     if overall_errors.present? || !result
       error_message % overall_errors.strip
+      {
+        :message => error_message % overall_errors.strip,
+        :host_details => host_details
+      }
     end
```
lzap, line 106: Line 101 is likely a left-over. Quite confusing :-)

test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -201,4 +201,14 @@ def test_auto_provision_all_no_rule_success
...
 
+  def test_reboot_all_failure
+    disable_orchestration
+    host = discover_host_from_facts(@facts)
+    ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
+    post :reboot_all, params: { }
+    show_response = ActiveSupport::JSON.decode(@response.body)
+    assert_equal "Errors during reboot: #{host.name}: failed to reboot", show_response["error"]["message"]
+    assert_equal "macaabbccddeeff: failed to reboot ", show_response["error"]["host_details"][0]["error"]
+  end
```
lzap, line 212: Can you rather assert whole JSON, so we can see how it looks like? Getting overall picture would be beneficial.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -275,7 +275,7 @@ def test_multiple_reboot_failure
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).returns(false)
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_equal "Errors during reboot: #{host.name}: failed to reboot", flash[:error]
+    assert_equal "{:message=>\"Errors during reboot: #{host.name}: failed to reboot\", :host_details=>[{:name=>\"#{host.name}\", :error=>\"macaabbccddeeff: failed to reboot \"}]}", flash[:error]
     assert_nil flash[:success]
```
lzap, line 279: Is it correct to display JSON via flash hash? Flash goes directly to page via notification box, so it won't render correctly. I think the controller should format this into plain string in some way.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -285,7 +285,7 @@ def test_multiple_reboot_error
     ::ForemanDiscovery::NodeAPI::PowerService.any_instance.expects(:reboot).raises("request failed")
     post :submit_multiple_reboot, params: {:host_ids => host.id}, session: set_session_user(User.current)
     assert_redirected_to discovered_hosts_url
-    assert_match(/ERF50-4973/, flash[:error])
+    assert_equal "{:message=>\"Errors during reboot: request failed\", :host_details=>[{:name=>\"#{host.name}\", :error=>\"request failed\"}]}", flash[:error]
```
lzap, line 288: The code is actually useful, I'd keep using it. Let me explain in different comment.

lzap (changes requested):
I think this needs a redesign, the reboot method is utterly complex. It was before your patch, not your fault! I think it's now the time to make it clean.

I suggest creating a helper exception class (maybe in core - can be very helpful) with unit tests. Something like:

```ruby
  class MultiException < ::Foreman::Exception
    def initialize(exceptions, message, *params)
      super(message, *params)
      raise "Can't create multi exception - not an array: #{exceptions}" unless exceptions.respond_to?(:to_a)
      @exceptions = exceptions.to_a.dup.freeze
    end

    attr_reader :exceptions

    def message
      super if @wrapped_exception.empty?
      super + @exceptions.map { |e| e.message }
    end

    def as_json
      {
        message: message
        errors: @exceptions.map { |e| e.message }
      }.as_json
    end
  end
```

And redesign the reboot method to utilize this immutable exception - create an array of wrapped exceptions and everytime you enconter an exception, create new `WrappedException("Unable to reboot host %s", host, original_exception)` and add it to the array, then create the `MultiException("Errors during rebooting")` instance and raise it. The UI controller should catch it and format the output as plain text via "message" method, API controller should use `as_json` to render output as JSON.

The ultimate goal of this:

* Proper message in plain and JSON
* Proper use of Ruby exceptions
* ERF codes for all error messages
* ERF code for the overall message "Errors during rebooting"

rabajaj0509:
@lzap The redesign is really needed indeed. Thanks for the review, will make the changes :)

rabajaj0509:
@lzap, @tstrachota 
Based on [this] (https://github.com/theforeman/foreman_discovery/pull/380#pullrequestreview-114678818) comment do we still want our outputs as discussed [here](https://projects.theforeman.org/issues/20549) (pasted below) 
```
{ "error" => {
    "message" => "Errors during reboot:\n
Unable to reboot mac525400d8c85d via https://192.168.122.25:8443: Failed to open TCP connection to 192.168.122.25:8443\n
Unable to reboot mac52540022ab12 via https://192.168.122.156:8443: Failed to open TCP connection to 192.168.122.156:8443" 
    "host_details" => [{
      "name" => "mac525400d8c85d",
      "error" => "Failed to open TCP connection to 192.168.122.25:8443" 
    },{
      "name" => "mac52540022ab12",
      "error" => "Failed to open TCP connection to 192.168.122.156:8443" 
    }]
  }
}
```

From how much I understand, the Multiexception class is concatenating message of each exception in the `exceptions` parameter and rendering it as `errors` field (in case of json) or as `message`  incase of html. 

Assuming `perform_reboot_all` method returns us an array of `WrappedException` objects or `false` (incase of no errors), doing the following will render us a concatenated string of messages as we are getting right now. Is this what you intended?
```
      def reboot_all
        exceptions = perform_reboot_all
        return process_success _("Discovered hosts are rebooting now") unless error_message
       ::Foreman::MultiException.new(exceptions, _N("Errors during rebooting")).to_json
      end
```

lzap:
If we can simplify the output to be just an array of messages, that could be much easier and also generic solution for other cases. If not, then you can create something like `MultiHostException` and render JSON in require format, but I'd prefer generic form.

rabajaj0509:
@lzap ++ i think the generic exception makes sense.

rabajaj0509:
created a new [pr](https://github.com/theforeman/foreman_discovery/pull/457), to refer the context here correctly. Closing this pr as is, hope it is not an issue.

PR #381, opened 2017-09-28
Fixes #21123 - Changed populate_fields_from_facts signature

ShimShtein:
To reflect changes from https://github.com/theforeman/foreman/pull/3649

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -67,11 +67,13 @@ class Host::Discovered < ::Host::Base
     host
   end
 
-  def import_facts facts
+  def import_facts(facts)
     # Discovered Hosts won't report in via puppet, so we can use that field to
     # record the last time it sent facts...
     self.last_report = Time.now
-    super
+    # Set the correct facts type for new foreman facts importing code.
+    facts[:_type] = :foreman_discovery if SETTINGS[:version].short > '1.16'
+    super(facts)
   end
 
   def setup_clone
@@ -84,10 +86,23 @@ class Host::Discovered < ::Host::Base
     super
   end
 
-  def populate_fields_from_facts(facts = self.facts_hash, ignored_type = nil)
-    # detect interfaces and primary interface using extensions
-    parser = super(facts, :foreman_discovery)
+  def populate_fields_from_facts(*params)
+    if SETTINGS[:version].short > '1.16'
+      parser, type, source_proxy = params
+      facts = parser.facts
+      # detect interfaces and primary interface using extensions
+      super(parser, type, source_proxy)
+    else
+      # backwards compatibility
+      facts = params[0] || self.facts_hash
+      parser = super(facts, :foreman_discovery)
+    end
 
+    populate_discovery_fields_from_facts(facts)
+    parser
+  end
+
+  def populate_discovery_fields_from_facts(facts)
     # set additional discovery attributes
     primary_ip = self.primary_interface.ip
     unless primary_ip.nil?
@@ -122,7 +137,6 @@ class Host::Discovered < ::Host::Base
     # lock the host into discovery via PXE, if feature is enabled in settings
     lock_templates if Setting::Discovered.discovery_lock? && self.subnet.tftp?
     self.save!
-    parser
   end
 
   def lock_templates


orrabin:
@ShimShtein This may fix the call to `super` but the code is called from host base so `populate_fields_from_facts` in discovery complains of too many arguments and it doesn't reach your code yet:

```
Api::V2::FactValuesControllerExtensionsTest#test_0001_list discovered host facts:
ArgumentError: wrong number of arguments (given 3, expected 0..2)
    /home/git/foreman/foreman_discovery/app/models/host/discovered.rb:87:in `populate_fields_from_facts'
    app/models/host/base.rb:157:in `parse_facts'
    app/models/host/base.rb:146:in `import_facts'
    /home/git/foreman/foreman_discovery/app/models/host/discovered.rb:74:in `import_facts'
    /home/git/foreman/foreman_discovery/app/models/host/discovered.rb:66:in `import_host'
    /home/git/foreman/foreman_discovery/test/test_helper_discovery.rb:127:in `block in discover_host_from_facts'
    app/models/concerns/foreman/thread_session.rb:93:in `as'
    app/models/concerns/foreman/thread_session.rb:99:in `as_anonymous_admin'
    /home/git/foreman/foreman_discovery/test/test_helper_discovery.rb:126:in `discover_host_from_facts'
    /home/git/foreman/foreman_discovery/test/functional/api/v2/fact_value_extensions_test.rb:29:in `block in <class:FactValuesControllerExtensionsTest>'
```

ShimShtein:
Fixed the code now. At least locally the tests are passing.

I think this fix is a patch and we need to refactor the whole interface and taxonomies handling out of `Host::Discovered` into `ForemanDiscovery::FactParser` where this kind of work belongs.
As an additional bonus, it will call `save` less times.

app/models/host/discovered.rb
```diff
@@ -84,10 +86,23 @@ def attributes_to_import_from_facts
...
-    parser = super(facts, :foreman_discovery)
+  def populate_fields_from_facts(*params)
+    if SETTINGS[:version].version > '1.16'
+      parser, type, source_proxy = params
+      facts = parser.facts
+      # detect interfaces and primary interface using extensions
+      super(parser, type, source_proxy)
+    else
+      # backwards compatibility
+      facts = params[0] || self.facts_hash
```
orrabin, line 97: If there are only 3 parameters I'd rather see another parameter with a default in the signature here and see what each parameter is then use params[0].

app/models/host/discovered.rb
```diff
@@ -84,10 +86,23 @@ def attributes_to_import_from_facts
     super
   end
 
-  def populate_fields_from_facts(facts = self.facts_hash, ignored_type = nil)
-    # detect interfaces and primary interface using extensions
-    parser = super(facts, :foreman_discovery)
+  def populate_fields_from_facts(*params)
+    if SETTINGS[:version].version > '1.16'
```
orrabin: Backward compatibility is great but there might have been more changes that make the next version of discovery compatible only with 1.17 anyway.
I'll leave it up to @lzap to decide if this is needed.

app/models/host/discovered.rb
ShimShtein: Previous signature was 
``` ruby
populate_fields_from_facts(facts = self.facts_hash, type = 'puppet')
```
It was 2 parameters with completely different meaning. For example the first one in 1.17 is mandatory and it's the parser, but in 1.16 it was facts hash and was optional. So now I mimic an optional parameter - use passed value and if it does not exist, use default (`self.facts_hash`) the same should go for the second parameter, but it's ignored anyway.

orrabin: Thanks for clarifying.

mmoll:
I'd suggest to branch the current master to 9.2-stable now for Foreman 1.16 and bring in breaking 1.17 changes in master then.

lzap:
We need to merge some amount of patches first, let's merge this and then branch off and remove this from stable branch.

lzap:
Getting errors on 1.16 branch, if this PR is supposed to work with both versions, can you fix this please?

ShimShtein:
Fixed the version check mechanism. Tested locally for both 1.6 and develop branches.

lzap:
Thanks tested as well.

PR #383, opened 2017-09-28
Fixes #21153 - Passing conditions to destroy_all is deprecated

mmoll:


mmoll:
this is compatible with Rails 4.2

lzap:
We can hopefully [test] now.

mmoll:
Not yet. But Rails 4 in develop is green now. After merging this, the 22 failures with Rails 5 will come down to about 6, which will need more analysis.

lzap:
Thanks.

PR #385, opened 2017-10-09
Fixes #21244 - Rails 5 upgrade test failures

lzap:
Remaining 2/3 R5 errors fixed, but there is one which wrote @dLobato and I don't understand. The assertion in the test looks incorrect to me, assert takes one parameter, should that be `assert_equal`? In any case, architecture is not being set there. Can you take a look? It's this one:

```ruby
module ForemanDiscovery
  module Concerns
    class HostsControllerTest < ActionController::TestCase
      tests ::HostsController

      context 'hosts controller requests from discovered_hosts url' do
        test 'get "host" params from "discovered_hosts" params' do
          architecture = FactoryGirl.create(:architecture)
          discovered_host_params = {
            'discovered_host' => { 'architecture_id' => architecture.id }
          }

          @request.stubs(:path).
            returns(architecture_selected_discovered_hosts_path)

          post :architecture_selected, discovered_host_params, set_session_user
          assert assigns('architecture'), architecture # THIS FAILS, 1st param is nil
        end
      end
    end
  end
end
```

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -97,7 +97,8 @@ module Api
       param :facts, Hash, :required => true, :desc => N_("hash containing facts for the host with minimum set of facts: discovery_bootif, macaddress_eth0, ipaddress, ipaddress_eth0, interfaces: eth0 (example in case primary interface is named eth0)")
 
       def facts
-        facts = params['facts']
+        # creating a host from facts is not a mass assignment - we store them individually
+        facts = params['facts'].to_unsafe_h
         state = true
         User.as_anonymous_admin do
           @discovered_host = Host::Discovered.import_host(facts)
--- a/test/functional/api/v2/discovery_rules_controller_test.rb
+++ b/test/functional/api/v2/discovery_rules_controller_test.rb
@@ -38,10 +38,11 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
         :name => "foo",
         :search => "bar",
         :hostgroup_id => hostgroup.id,
-        :organization_ids => [organization_one],
+        :organization_ids => [organization_one.id],
         :location_ids => [location_one.id],
         :hostname => "",
         :priority => 1}}
+      refute_match /error/, response.body
     end
     assert_response :success
   end
@@ -54,7 +55,6 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
 
   test "should update taxonomy for discovery rule" do
     rule = FactoryGirl.create(:discovery_rule)
-    organization = FactoryGirl.create(:organization)
     put :update, { :id => rule.to_param, :discovery_rule => { :organization_ids => [rule.organizations.first.id] } }
     assert_response :success
   end
--- a/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
+++ b/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
@@ -7,16 +7,15 @@ module ForemanDiscovery
 
       context 'hosts controller requests from discovered_hosts url' do
         test 'get "host" params from "discovered_hosts" params' do
-          architecture = FactoryGirl.create(:architecture)
+          os = FactoryGirl.create(:operatingsystem, :with_associations)
+          arch_id = os.architectures.first.id
           discovered_host_params = {
-            'discovered_host' => { 'architecture_id' => architecture.id }
+            'discovered_host' => { 'architecture_id' => arch_id }
           }
 
-          @request.stubs(:path).
-            returns(architecture_selected_discovered_hosts_path)
-
+          @request.stubs(:path).returns(architecture_selected_discovered_hosts_path)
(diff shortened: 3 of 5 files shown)


app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -97,7 +97,8 @@ def destroy
       param :facts, Hash, :required => true, :desc => N_("hash containing facts for the host with minimum set of facts: discovery_bootif, macaddress_eth0, ipaddress, ipaddress_eth0, interfaces: eth0 (example in case primary interface is named eth0)")
 
       def facts
-        facts = params['facts']
+        # creating a host from facts is not a mass assignment - we store them individually
+        facts = params['facts'].to_unsafe_h
```
lzap, line 101: This is not a mass assignement, so it is safe to do. We do the same in core:

app/controllers/api/v2/hosts_controller.rb
289:        state = @host.import_facts(params[:facts].to_unsafe_h, detected_proxy)

test/functional/api/v2/discovery_rules_controller_test.rb
```diff
@@ -38,10 +38,11 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
         :name => "foo",
         :search => "bar",
         :hostgroup_id => hostgroup.id,
-        :organization_ids => [organization_one],
+        :organization_ids => [organization_one.id],
```
lzap, line 41: Looks like ID parser in Rails 5 did not catch this anymore, it was a typo anyway. Integer is expected here.

All the rest was just cleaning up Ruby compiler warnings - missing parentheses and unused variable.

dLobatog (commented):
@lzap It should be either:

- To make sure `@architecture` is set to something (not nil) after the controller call.

```ruby
assert assigns('architecture')
```

To make sure `@architecture` and `architecture` are the same:

```ruby
assert_equal architecture, assigns('architecture')  
```

lzap:
Right, that's what I though. Any clue why `assigns('architecture')` is `nil` with Rails 5? I can't figure it out.

dLobatog:
@lzap I think here's the culprit: 

https://github.com/rails/rails/issues/18950

If I'm not wrong you can check `assert_equal architecture, @architecture`

lzap:
Unfortunately, the variable is really nil:

--- expected
+++ actual
@@ -1 +1 @@
-#<Architecture id: 501905020, name: "x86_64-27", created_at: "2017-10-09 11:46:46", updated_at: "2017-10-09 11:46:46">
+nil

lzap:
Finally solved it, spend two days digging in core but this was all about controller attributes not available in unit tests anymore. You can still access it via `@controller` tho. Tests are green for both R4 and R5.

test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
```diff
@@ -12,11 +12,11 @@ class HostsControllerTest < ActionController::TestCase
...
 
-          @request.stubs(:path).
-            returns(architecture_selected_discovered_hosts_path)
-
+          @request.stubs(:path).returns(architecture_selected_discovered_hosts_path)
           post :architecture_selected, discovered_host_params, set_session_user
-          assert assigns('architecture'), architecture
+
+          # normally we would test response.body but in this case it contains OS entries
+          assert_equal architecture, @controller.instance_variable_get("@architecture")
```
ares: in Rails 5 they made functional tests equal to integration tests, we should perhaps test that response contains architecture.name or something and what was the response code

ares (changes requested):
one nit, otherwise looks good to me, thanks for inline comments, it saved me some time

test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
lzap: Agree but see the comment - the response actually does not contain architecture but operating systems associated with this architecture. It's confusing. I could however do some factorygirl magic to create some associations and test that if you insist.

lzap: Ok changed it to integration test.

lzap:
Anything else?

ares:
LGTM, I haven't tested it, if you could trigger the jenkins job manually to confirm, that would be great (or let me know if you don't know how). I don't have merge permissions but after jenkins confirms I think this can be merged.

lzap:
I think I have permissions but how do I do that? Which job is that?

lzap:
Lemme try [test]

timogoebel:
This looks good. If tests pass on Jenkins ([test] does not seem to help, but I don't know any other way as well), I'll happily merge this.

ares:
test won't help until develop branch is green, you need to trigger test_plugin_pull_request, triggered at http://ci.theforeman.org/job/test_plugin_pull_request/4210/

ares:
the result is green, this can (and should be merged)

lzap:
http://ci.theforeman.org/job/test_plugin_pull_request/4210/ ALL GREEN I am skipping our rule and merging this time.

PR #386, opened 2017-10-09
Fixes #21087 - Correct Api doc for discovery rules

rabajaj0509:


rabajaj0509:
@lzap looks like : 

required

lzap:
Thanks. Not sure about the description, I'd say "Organization ID for provisioned hosts. It must match with hostgroup." The same for location. Once you rebase tests should be finally green.

PR #387, opened 2017-10-20
Fixes #21407 - fix broken foreign keys

ares:
this fixes the issue for users who have not yet upgraded, for users who ran this migration it worked anyway :-)

to reproduce
1) create a smart proxy with discovery feature
2) create a subnet that you link to it
3) remove the smart proxy, check the subnet has still discovery_id set to that deleted smart proxy in db
4) run this migration


orrabin:
Edit: I was able to reproduce the error even with this fix.

orrabin:
Adding the `unscoped` fixed this.
Tests are broken and this needs to get in fast, I made sure all tests were passing locally.

orrabin:
Thanks @ares!

PR #388, opened 2017-10-20
Fixes #21417 - Host update propagates facets

parthaa:
Discovery Host update call would filter out Facet parameters on update.
This caused things ContentFacet an SubscriptionFacet used by Katello to
be not propagated over when one provisioned the Host with updates on the
synced media.

This commit fixes that issue by making the host provisioning call use
managed host parameters for updates so that those facets can go through.


lzap:
Thanks I roughly tested this, unit [test] run is green, it works fine on my system.

PR #391, opened 2017-11-27
Fixes #21653 - render delegated nil correctly

lzap:
Fixes RABL issue when delegate is nil. Will return nil, but should not fail.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -9,7 +9,7 @@ class Host::Discovered < ::Host::Base
 
   validates :discovery_attribute_set, :presence => true
 
-  delegate :memory, :cpu_count, :disk_count, :disks_size, :to => :discovery_attribute_set
+  delegate :memory, :cpu_count, :disk_count, :disks_size, :to => :discovery_attribute_set, :allow_nil => true
   after_destroy :delete_notification
 
   scoped_search :on => :name, :complete_value => true


app/models/host/discovered.rb
```diff
@@ -190,7 +190,7 @@ def refresh_facts
     raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url, :error => e)
   end
 
-  def reboot 
+  def reboot
```
lzap: This is just a whitespace, unrelated.

lzap:
Tests failed because of core, this patch is green. Feel free to re-test once develop is green again. [test]

lzap:
Aaand rebased.

ares:
Before I merge, how it happens that host does not have `discovery_attribute_set`?

lzap:
We do not use transactions and discovery is two AR save process.

ares:
Is that race condition of two users then? Just trying to figure out, what the impact will be :-) Then, I'm happy to merge.

lzap:
No, when exception happens during host import, attributes are simply not yet set. We need to save because of NICs.

ares:
Do we render rabl template during this operation? That sounds weird to me. I understand the controller action does not run in transaction, but rabl template is rendered after the controller job is done, so both host and discovery_attribute_set should be saved. Did you check why the discovery_attribute_set was not saved at that moment? Perhaps some [validation issues](https://github.com/theforeman/foreman_discovery/blob/develop/app/models/discovery_attribute_set.rb#L4-L6) if some fact were missing?

lzap:
We save those attributes during discovery phase:

https://github.com/theforeman/foreman_discovery/blob/develop/app/models/host/discovered.rb#L125-L127

We can't use transactions for discovery operations - it breaks orchestration in Katello. And to discover a host, multiple save calls are needed.

This patch does not aim to solve this, it only prevents from erroring out when you want to display such a host with no attributes set.

ares:
so after discussion on IRC, I understood this happens if there's some problem with host saving and user later goes to dicsovered host detail page

in order for this to work (even with inconsistent DB), it's reasonable to allow nil attribute set

thanks @lzap for clarification and patience with me :-)

PR #393, opened 2017-12-06
Fixes #21890 - Fact importer transaction check suppress in tests

ShimShtein:
Should fix test failures for theforeman/foreman#5030


lzap:
Thanks a bunch!

PR #394, opened 2017-12-07
Fixes #21908 - Safemode doesn't allow to access 'resbody' on rescue

lboclboc:
This fixes the same problem as was reported in bug Bug #12237 but for the debian template


lzap:
Thanks, looks like we missed to fix debian kexec!

PR #395, opened 2017-12-12
Fixes #21941 - Add validation for Host Group

rabajaj0509:
We are just checking whether the `hostgroup_id` is being passed from hammer but we are not checking whether that hostgroup_id has a corresponding hostgroup in the database or not. 

This patch makes sure, only valid hostgroup_id's are passed which correspond to a valid hostgroup.

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -11,6 +11,7 @@ class DiscoveryRule < ApplicationRecord
   validates :hostname, :format => { :with => /\A[a-zA-Z<]/, :message => N_("must start with a letter or ERB.") },
     :allow_blank => true
   validates :hostgroup_id, :presence => true
+  validates :hostgroup, :presence => {:message => N_("must be present.")}
   validates :max_count, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates :priority, :presence => true, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates_lengths_from_database
--- a/db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
+++ b/db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
@@ -1,0 +1,8 @@
+ class AddConstraintsOnDiscoveryRulesHostgroups < ActiveRecord::Migration[4.2]
+  
+   def change
+    DiscoveryRule.unscoped.where("hostgroup_id IS NULL OR hostgroup_id NOT IN (?)", Hostgroup.unscoped.all.pluck(:id)).delete_all
+    add_foreign_key "discovery_rules", "hostgroups", name: "discovery_rules_hostgroup_id_fk", :column => "hostgroup_id"
+  end
+
+ end
--- a/test/unit/discovery_rule_test.rb
+++ b/test/unit/discovery_rule_test.rb
@@ -72,13 +72,24 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
     assert_valid rule
   end
 
-  test "should not be able to create a rule without a hostgroup" do
+  test "should not be able to create a rule without a hostgroup-id" do
     rule = DiscoveryRule.new :name => "myrule",
       :search => "cpu_count > 1",
       :organization_ids => [organization_one.id],
       :location_ids => [location_one.id]
     refute_valid rule
     assert_equal "can't be blank", rule.errors[:hostgroup_id].first
+  end
+
+  test "should not be able to create a rule without a valid hostgroup" do
+    rule = DiscoveryRule.new :name => "myrule",
+      :search => "cpu_count > 1",
+      :organization_ids => [organization_one.id],
+      :location_ids => [location_one.id],
+      :hostgroup_id => 99999
+    refute_valid rule
+
+    assert_equal "must be present.", rule.errors[:hostgroup].first
   end
 
   test "should not be able to create a rule with priority bigger than 2^31" do


app/models/discovery_rule.rb
```diff
@@ -32,6 +33,12 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def hostgroup_present
+    if hostgroup_id.present? && hostgroup.blank?
+      errors.add(:hostgroup, _("is invalid."))
```
timogoebel: Could be written as: `errors.add(:hostgroup, :invalid)`

app/models/discovery_rule.rb
```diff
@@ -16,6 +16,7 @@ class DiscoveryRule < ApplicationRecord
   validates_lengths_from_database
   before_validation :default_int_attributes
   before_validation :enforce_taxonomy
+  before_validation :hostgroup_present
```
timogoebel: Why not `validates_presence_of :hostgroup`?

timogoebel (changes requested):
Thanks. Some comments inline. Please also add a test for this:

https://github.com/thoughtbot/shoulda-matchers/blob/5d9f8da8c6611c591813a57417edff4bcaf9bc24/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb#L20-L22

app/models/discovery_rule.rb
rabajaj0509: @timogoebel hey, `validates_presence_of` just validates that the specified attributes are not blank (as defined by Object#blank?). Here, i am trying to make a call to the database and check whether the value is present or not? also making sure a `dummy` value does not pass through.

you might want to run - `hammer -u admin -p changeme discovery-rule create --hostgroup-id 999 --locations="Default Location" --name "rabajaj" --search "last_report = Today" --organizations="Default Organization"` and check the output :)

timogoebel: @rahulbajaj0509: If I read the rails source code correctly, it just calls `hostgroup.blank?` internally. Isn't that the same thing you are doing?

lzap: Well the standard practice is to use `validates_presence_of`. Can you provide an example when the standard approach would not work and yours would please?

lzap: Be sure to use `validates_presence_of :hostgroup` and not `validates_presence_of :hostgroup_id`. The latter is wrong.

rabajaj0509: @lzap 
CASE 1: If you use `validates_presence_of` and run: 
$ hammer -u admin -p changeme discovery-rule create --hostgroup-id 999 --locations="Default Location" --name "rabajaj" --search "last_report = Today" --organizations="Default Organization"
Rule will be created(note : 999 hostgroup does not exist but since it is being passed and not blank, it is accepted)

CASE 2: If we use the function provided in this patch:
Rule will not get created.

timogoebel: I'm wondering if we have this bug across the whole app.

lzap: We do use foreign keys for that both in core and in discovery, please add it to this association and that should do it. Example:

```
app/models/host/discovered.rb
8:  has_one    :discovery_attribute_set, :foreign_key => :host_id, :dependent => :destroy
```

lzap: Note you might want to create FK via migration, I don't remember how this is done but http://guides.rubyonrails.org/active_record_migrations.html#foreign-keys

rabajaj0509:
@lzap @timogoebel  looks good? I have added a FK via migration.

lzap:
Thanks, I do not understand why are you actually allowing nils. This is great opportunity to disallow nils completely and enforce it on SQL level. Is this something that Rails can't live without?

rabajaj0509:
@lzap  good idea! removed the nils. My only idea for having the nil values was that if we by any chance had any pre-existing invalid entries made against the hostgroup_id(eg: 99 or something similar) then we might face FK errors, else for future, I guess removing the nil values would be a plus. No tests failing on the local :)

rabajaj0509:
@lzap I guess, irrelevant tests are failing.

lzap:
Why simple

     validates :hostgroup, :presence => true

would not work? I think validating hostgroup_id is incorrect and should have been removed by this patch. Adding foreign key is good idea, but I don't understand why you disable checks during migration, we don't do that in core. If this is to prevent migration to fail if there are rules without hostgroup set, then we shoud fix this first and then add the key.

The only way of fixing this is to delete those rules, or set them to a first (random) hostgroup found which is perhaps more dangerous (it's safer to break provisioning than to let a system to be provisioned with incorrect hostgroup).

app/models/discovery_rule.rb
```diff
@@ -11,6 +11,7 @@ class DiscoveryRule < ApplicationRecord
   validates :hostname, :format => { :with => /\A[a-zA-Z<]/, :message => N_("must start with a letter or ERB.") },
     :allow_blank => true
   validates :hostgroup_id, :presence => true
+  validates :hostgroup, :presence => { :message => N_("does not exist, please check hostgroup-id.") }, :if => :hostgroup_id?
```
rabajaj0509: @lzap the intention here is not to validate the hostgroup_id but to have a custom error message. If we only have `validates :hostgroup, :presence => true`, the error message would be something like - `Hostgroup can't be blank` which I feel will not convey the correct error to the end-users.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
```diff
@@ -0,0 +1,17 @@
+class AddConstraintsOnDiscoveryRulesHostgroups < ActiveRecord::Migration[4.2]
+  def change
+    # turn off Foreign Key checks
```
rabajaj0509: @lzap I feel if we remove the checks, then it might cause a huge issue during `upgrade`. The checks do not affect the success/failure of the migration, the idea here is to be safe during upgrade.
Another option here would be to remove the checks and write code to delete the discovery rules which do not have the appropriate hostgroup, this option again might cause problems during the upgrade process.

app/models/discovery_rule.rb
lzap: I don't understand how "Hostgroup can't be blank" is not correct error. But if you insist, go for "must be present". Also remove the "if => :hostgroup_id?" part, that prevents from validating new records.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
lzap: I vote deleting them instead, it's not big deal, if no hostgroup was set, they are likely some unused leftovers.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
```diff
@@ -0,0 +1,14 @@
+ class AddConstraintsOnDiscoveryRulesHostgroups < ActiveRecord::Migration[4.2]
+  
+   def change
+    execute "DELETE from discovery_rules dr where NOT EXISTS (select NULL from hostgroups hg where hg.id = dr.hostgroup_id);"
```
lzap: Not exactly sure what you are trying to do here, but this should be:

    DiscoveryRule.where("hostgroup_id IS NULL").delete_all

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    add_foreign_key "discovery_rules", "hostgroups", name: "discovery_rules_hostgroup_id_fk", :column => "hostgroup_id"
+
+    # turn on Foreign Key checks in MySQL only
+    if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+      ActiveRecord::Migration.execute "SET FOREIGN_KEY_CHECKS=1;"
+    end
```
lzap: Remove this block please, this does not make any sense. It's a session-based setting.

lzap:
Okay it is getting into shape, thanks. Two small nitpicks, then let's test this.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
rabajaj0509: Here, I have checked in both the tables - `discovery_rules` and `hostgroups` table. I have checked whether the hostgroup_id(which is a column in the discovery_rules table) has an equivalent id(which is a column of the hostgroups table). 
I have tested this query and it works really efficiently :)

The solution that you have provided in the review, say if there is a discovery rule present but the hostgroup is deleted, then, in that case, you will face a `FOREIGNKEY_VIOLATION` error, to avoid that I suggest we use the above query.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
rabajaj0509: Sorry, It just slipped out of my mind. Will do that.

lzap:
Triggered http://ci.theforeman.org/job/test_plugin_foreman_discovery_develop/ failures because of core.

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
lzap: Ah in that case I am in favour of:

    DiscoveryRule.where("hostgroup_id IS NULL OR hostgroup_id NOT IN (?)", Hostgroup.all.pluck(:id)).delete_all

Your query is more efficient but we are dealing with hundreds of records maximum. MySQL does not support subselects and I want to avoid testing this on MySQL to be honest (my tip: won't work). Thanks :-)

lzap: Correction!

    DiscoveryRule.unscoped.where("hostgroup_id IS NULL OR hostgroup_id NOT IN (?)", Hostgroup.unscoped.all.pluck(:id)).delete_all

rabajaj0509: @lzap done.

rabajaj0509:
@lzap the test addressed in https://github.com/theforeman/foreman_discovery/pull/404 is failing, which is a good sign :)

We need to merge https://github.com/theforeman/foreman_discovery/pull/404 for the tests to be green here.

lzap:
Pls rebase once #404 is merged.

rabajaj0509:
@lzap rebased!

lzap:
Thanks, can you rebase once again and get rid of `[Discovery rule]`? I don't like it, we only have couple of components, it would make more sense in core perhaps, but I generate changelog/announcement from git and it does not look good.

rabajaj0509:
@lzap sure! done :)

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
```diff
@@ -0,0 +1,8 @@
+ class AddConstraintsOnDiscoveryRulesHostgroups < ActiveRecord::Migration[4.2]
+  
+   def change
```
lzap, line 3: I fixed extra space.

lzap:
Ok not sure how you rebased it but I still see the `[Discovery rule]` in your commit. Nevermind, I tested, and I amended the commit change, it's now merged as 0f49002 thanks!

db/migrate/20171222120314_add_constraints_on_discovery_rules_hostgroups.rb
```diff
@@ -0,0 +1,8 @@
...
+    DiscoveryRule.unscoped.where("hostgroup_id IS NULL OR hostgroup_id NOT IN (?)", Hostgroup.unscoped.all.pluck(:id)).delete_all
```
lzap, line 4: Damn one more thing, this should be written as:

```ruby
      reversible do |dir|
        dir.up do
          DiscoveryRule.unscoped.where("hostgroup_id IS NULL OR hostgroup_id NOT IN (?)", Hostgroup.unscoped.all.pluck(:id)).    +delete_all
        end
      end
```

I made the change in separate commit: 6efc6cc

PR #398, opened 2017-12-19
Fixes #21935 - Notifications for org admins too

dLobatog:
Before this commit, the notifications for new discovered host were only
seen by admins (global admins). However, the organization admins should
also be able to see these notifications.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -10,7 +10,6 @@ class Host::Discovered < ::Host::Base
   validates :discovery_attribute_set, :presence => true
 
   delegate :memory, :cpu_count, :disk_count, :disks_size, :to => :discovery_attribute_set
-  after_create :create_notification
   after_destroy :delete_notification
 
   scoped_search :on => :name, :complete_value => true
@@ -105,6 +104,7 @@ class Host::Discovered < ::Host::Base
     end
 
     populate_discovery_fields_from_facts(facts)
+    create_notification
     parser
   end
 
@@ -196,7 +196,7 @@ class Host::Discovered < ::Host::Base
     raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url, :error => e)
   end
 
-  def reboot 
+  def reboot
     resource = ::ForemanDiscovery::NodeAPI::Power.service(:url => proxy_url)
     resource.reboot
   rescue => e
@@ -246,5 +246,29 @@ class Host::Discovered < ::Host::Base
 
   def delete_notification
     ForemanDiscovery::UINotifications::DestroyHost.deliver!(self)
+  end
+
+  def notification_recipients_ids
+    org_recipients = find_organization_users if Taxonomy.organizations_enabled
+    org_recipients ||= []
+
+    admins = User.unscoped.only_admin.except_hidden.
+      reorder('').distinct.pluck(:id)
+    (org_recipients + admins).uniq
+  end
+
+  private
+
+  def find_organization_users
+    return [] unless organization.present?
+    organization_users = if organization.ignore_types.include? 'User'
+                           User.unscoped.all
+                         else
+                           organization.users
+                         end
+
+    organization_users.find_all do |user|
+      user.can? :create_hosts
+    end.pluck(:id)
   end
 end
--- a/test/unit/host_discovered_test.rb
+++ b/test/unit/host_discovered_test.rb
@@ -26,7 +26,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
   end
 
   test "should setup subnet" do
-    raw = parse_json_fixture('/facts.json')
     subnet = FactoryBot.create(:subnet_ipv4, :name => 'Subnet99', :network => '10.35.27.0', :organizations => [organization_one], :locations => [location_one])
     Subnet.expects(:subnet_for).with('10.35.27.3').returns(subnet)
     host = discover_host_from_facts(@facts)
@@ -38,7 +37,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     loc = FactoryBot.create(:location, :name => "subnet_loc")
     Setting['discovery_organization'] = org.name
     Setting['discovery_location'] = loc.name
-    raw = parse_json_fixture('/facts.json')
     subnet = FactoryBot.create(:subnet_ipv4, :name => 'Subnet99', :network => '10.35.27.0', :organizations => [org], :locations => [loc])
     Subnet.expects(:subnet_for).with('10.35.27.3').returns(subnet)
     host = discover_host_from_facts(@facts)
@@ -48,7 +46,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
   test "should setup subnet with org and loc set via facts" do
     org = FactoryBot.create(:organization, :name => "subnet_org_via_fact")
     loc = FactoryBot.create(:location, :name => "subnet_loc_via_fact")
-    raw = parse_json_fixture('/facts.json')
     @facts['foreman_organization'] = org.name
     @facts['foreman_location'] = loc.name
     subnet = FactoryBot.create(:subnet_ipv4, :name => 'Subnet99', :network => '10.35.27.0', :organizations => [org], :locations => [loc])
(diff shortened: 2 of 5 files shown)


app/models/host/discovered.rb
```diff
@@ -247,4 +247,14 @@ def create_notification
...
+
+  def notification_recipients_ids
+    org_admins = Role.unscoped.find_by(
+      :name => Role::ORG_ADMIN,
+      :organization_id => organization_id
+    ).users.pluck(:id)
+    admins = User.unscoped.only_admin.except_hidden.
+      reorder('').distinct.pluck(:id)
+    (org_admins + admins).uniq
+  end
```
lzap: Thanks, will this work also for cloned roles from ORG_ADMIN?

lzap (changes requested):
It does not work for me. I think this is because my admin role does not belong to any organization. You search for role and its taxonomy:
 
    Role.unscoped.find_by(
      :name => Role::ORG_ADMIN,
     :organization_id => organization_id
     )

But it's empty for me and I think it is supposed to be empty, these are non-editable default roles. I think you need to join users and search by user organization rather than role.

Also we are missing a test here.

dLobatog (commented):
I think I'll add a method 'admin_users' to Organization which finds these in core, finding org admins isn't as easy as this PR depicts apparently.

app/models/host/discovered.rb
ares: I'm afraid search by `:name => Role::ORG_ADMIN` won't work, the workflow is that the user clones this template role, and rename it to e.g. "Org admin of Default org". They don't assign organizations to this template role directly.

Isn't there a specific permission that we should check? I guess all users in org that have permission to provision the discovered host and are assigned to this organization should be in the audience?

lzap:
Absolutely, both `admins` and `org_admins` queries should have own helper methods as this is fragile code and we should definitely do it.

I still have question of enforcement - since both roles and users have taxonomy, shall we enforce it? Where exactly? In the helpers? @ares

ares:
I guess all users who can create_hosts in the organization might be interested, so we could search for users having this permission? To find org admins, we could look at the `cloned_from` of all roles but that seems weird. Why other people that can create hosts wouldn't receive a notification?

Lukas, I'm not sure what you mean by enforcement. I think following must be true in order to get the notification:
1) user is in org where the org has been discovered
2) user has a permission to create_host in this organization

exceptions I see
for 1) admin is automatically in every organization even without explicit assignment, so all admins should be always notified
for 2) user has a role that gives you create_host permission but is not assign to any specific org/loc, right now such permission gives you ability to create hosts in all orgs so I think such users should also be notified

dLobatog (commented):
@lzap @ares Repushed, I think this is what you were looking for, with the exception that users who are not in the organization but have 'create_users' are not notified (I'm not too sure that case is correct)

app/models/host/discovered.rb
```diff
@@ -247,4 +247,20 @@ def create_notification
...
+
+  def notification_recipients_ids
+    organization_users = if organization.ignore_types.include? 'User'
+                           User.unscoped.all
+                         else
+                           organization.users
+                         end
+
+    org_recipients = organization_users.find_all do |user|
+      user.allowed_to? "create_hosts"
```
ares: I think `user.can?` would be a better function, this only checks permissions for a given controller action

app/models/host/discovered.rb
```diff
@@ -247,4 +247,20 @@ def create_notification
...
+    end
```
ares: the whole login until this line would be probably worth extracting into new method like `Organization#users_with_permission(name)`. While it will also find users with filtered permissions and might result in notifying a user who at the end can't create the host, it's still good guess. I know it would be useful e.g. in Katello to display notification for users who can refresh manifests.

dLobatog:
Updated to use `.can?`

lzap:
I am not seeing any notification for even admin user, @ares do you want to finish your review?

dLobatog (commented):
I'm unable to see notifications now too - please hold off from merging for a moment. I'm not sure why, as debugging shows the Notification gets created but maybe removed right away...

dLobatog (commented):
@lzap This should be fine now, I think I removed the ':subject => subject' line in some rebase. I have also added tests and tested manually, didn't find any problems with it yet 😄

ares (commented):
As I mentioned earlier, `find_organization_users` would be better to extract to `Organization` model so we can easily search for users with certain permission. But I'm fine with keeping it here, ideally open the redmine issue for core.

Sadly I was unable to test as in my latest discovery and foreman checkout, discovery default org does not seem to work and host is always registered to any context. @lzap is this a known issue?

dLobatog:
@ares I'm not seeing that issue, do you see any errors in the logs? My default organization didn't have the subnet where the host is discovered, and it displayed an error "subnet not found" or something similar.

lzap:
Discovery and taxonomy works for me, check your discovered host is in subnet you have assigned your taxonomy you want to have. When testing this I do see notification now by the way.

app/models/host/discovered.rb
```diff
@@ -247,4 +247,29 @@ def create_notification
...
+    organization_users = if organization.ignore_types.include? 'User'
+                           User.unscoped.all
+                         else
+                           organization.users
+                         end
+
+    organization_users.find_all do |user|
+      user.can? :create_hosts
+    end.pluck(:id)
+  end
```
lzap, line 273: This assumes that all hosts are always discovered in `Setting[:discovery_organization]` but that's not the case. We have pretty complex (and weird) mechanism of finding taxonomy described here: https://theforeman.org/plugins/foreman_discovery/9.1/index.html#3.1.3OrganizationsandLocations

I guess you need to use discovered host's organization instead of setting here. This needs to be trigerred tho after host import (we set taxonomy in `populate_fields_from_facts`).

lzap (changes requested):
If I read that correctly, this will only work for default orgs.

app/models/host/discovered.rb
dLobatog: I was using `host.organization` before, but yeah it was empty as this happens prior to `populate_fields_from_facts`. I'll change that.

app/models/host/discovered.rb
```diff
@@ -105,6 +104,7 @@ def populate_fields_from_facts(*params)
     end
 
     populate_discovery_fields_from_facts(facts)
+    create_notification
```
dLobatog, line 107: I moved the creation of the notification *after* facts are populated, so that we can know the Host organization to fetch the org admins. If we do this prior to importing the facts, we would have to guess the organization - replicating the logic that already does that at fact-import time.

test/unit/host_discovered_test.rb
```diff
@@ -26,7 +26,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
   end
 
   test "should setup subnet" do
-    raw = parse_json_fixture('/facts.json')
```
dLobatog, line 29: I removed many of these instances. They're worthless as far as I can tell. Tests pass without them, moreover the `raw` variables are not even referenced later.

test/unit/ui_notifications/destroy_host_test.rb
```diff
@@ -7,20 +7,23 @@ class DestroyHostNotificationTest < ActiveSupport::TestCase
...
+    ForemanDiscovery::UINotifications::NewHost.deliver!(host)
     assert_difference('blueprint.notifications.count', -1) do
       host.destroy
     end
   end
 
   test 'do not remove notification if other discovered hosts exists' do
     assert_difference('blueprint.notifications.count') do
-      FactoryBot.create(:discovered_host)
+      host = FactoryBot.create(:discovered_host)
```
dLobatog, line 20: Since creating the host is not enough to have the notification (we need to parse facts too), I had to add `deliver!` so there are already some pre-existing notifications to delete

dLobatog (commented):
@lzap Thanks for the review! I moved the `create_notification` method to the end of `populate_fields_from_facts`

dLobatog:
[test]
[test default]

dLobatog (commented):
@lzap Test failures have nothing to do with this PR

dLobatog:
[test] [test default]

lzap (approved):
WFM @ares any comments on the taxonomy part here?

ares:
just tested this under org admin and it works fine now, perhaps pebkac :-) I didn't do a full review but my `can?` concern was fixed and the feature works fine :-)

lzap:
I set notification interval to 1 second, discover a host and I am logged as super admin - no notification at all. Is this expected behavior?

lzap:
Also as org admin I do not see any notifications at all.

ares:
@lzap how did you create org admin? that one worked fine in my case

lzap:
Something is wrong with my ntf dev setup, I am merging and I will test nightly tomorrow. You both say it works.

lzap:
For the record, this failed QA in Satellite. Obviously "I will test nightly tomorrow" never happened...

PR #399, opened 2017-12-27
fixes #22083 - replace deprecated am dirty methods

timogoebel:
Full disclosure: This is untested.

PR #400, opened 2018-01-03
Fixes #22083 - update to Rails 5.1 change tracking

mmoll:


timogoebel:
@mmoll: Is this a duplicate of #399?

PR #401, opened 2018-01-03
Refs #22083 - remove config/as_deprecation_whitelist.yaml

mmoll:


timogoebel:
Just to make sure: Will this fallback to the whitelist in core, then?

mmoll:
Yes.

The thing is that the whitelist in core will always be used and plugins are completely whitelisted by default - except they're bring thier own whitelist, which was just an empty file in this case, triggering the deprecation notices to be fatal.

timogoebel:
Merged, danke @mmoll.

PR #404, opened 2018-01-12
Fixes #22257 -Discovery Rule should not be created without Taxonomy

rabajaj0509:


lzap:
Does it mean the test did not work? 

Anyway, do not delete the test but reverse it - if we expect an error, test for that error.

rabajaj0509:
Makes sense.

rabajaj0509:
@lzap I am unable to digest the fact that our build was green with this test case present. Can throw some light on this, I tried to understand the build - http://ci.theforeman.org/job/test_plugin_foreman_discovery_develop/384/console but didn't find it quite useful. Can you please direct me to the page where  I can see the console output running all the test cases for foreman discovery?

rabajaj0509:
@lzap I have made the changes for this test to pass. But if you see closely, a lot of tests on the `test/functional/discovery_rules_controller_test.rb` fail locally. 
Let me know if the tests for `test/functional/discovery_rules_controller_test.rb`  fail for you too so that I can fix it.

rabajaj0509:
Test is passing locally. 

The error shown via Jenkins here(http://ci.theforeman.org/job/test_plugin_pull_request/4548/database=mysql,label=fast,ruby=2.4/testReport/junit/(root)/DiscoveryRulesControllerTest/test_0004_should_not_create_discovery_rule_without_taxonomy/)  is exactly the opposite of what i see locally. 

Locally:
Actual = 0

lzap:
I see the same as Jenkins:

```
Failure:
DiscoveryRulesControllerTest#test_0004_should not create discovery rule without taxonomy [/home/lzap/work/foreman_discovery/test/functional/discovery_rules_controller_test.rb:23]:
"DiscoveryRule.unscoped.count" didn't change by 0.
Expected: 0
  Actual: 1
```

When I debug the test, I actually see that taxonomy is assigned:

```
[4] pry(#<DiscoveryRulesControllerTest>)> DiscoveryRule.unscoped.first.organizations
=> [#<Organization:0x0055abe777bc10
  id: 447626438,
  name: "Organization 1",
  type: "Organization",
  created_at: Thu, 25 Jan 2018 13:23:00 UTC +00:00,
  updated_at: Thu, 25 Jan 2018 13:23:00 UTC +00:00,
  ignore_types: [],
  ancestry: nil,
  title: "Organization 1",
  description: nil>]
[5] pry(#<DiscoveryRulesControllerTest>)> DiscoveryRule.unscoped.first.locations
=> [#<Location:0x0055abe77fc838
  id: 255093256,
  name: "Location 1",
  type: "Location",
  created_at: Thu, 25 Jan 2018 13:23:00 UTC +00:00,
  updated_at: Thu, 25 Jan 2018 13:23:00 UTC +00:00,
  ignore_types: [],
  ancestry: nil,
  title: "Location 1",
  description: nil>]
```

@ares we see taxonomy being automatically added for a UI endpoint (discovery rule creation), is this expected? FYI here is the test which fails, we want to test that Rule without taxonomy is NOT being created:

```ruby
    test "should not create discovery rule without taxonomy" do
      assert_difference('DiscoveryRule.unscoped.count', 0) do
        post :create, params: {:discovery_rule => {
          :name => "foo",
          :search => "cpu_count = 42",
          :hostgroup_id => 1,
          :hostname => "",
          :priority => 1}}, session: set_session_user_default_manager
        assert_empty(extract_form_errors(response))
      end
      assert_redirected_to discovery_rules_path
    end
```

ares:
Taxonomy is being assigned automatically if you're in that context (Organization.current is set) and what's worse, it can't be unassigned even if you set organization_ids => [] explicitly. Based on my testing, it checks that hostgroup is in same orgs, this test only checks for errors in response which in facts are not set. I guess you should check for response to be 302, if taxonomies of hostgroup and this rules mismatches, you get 200. The fact user is not presented any error is another bug.

lzap:
Hmmm in that case test for that behavior, set `current` org/loc to something and then check it.

rabajaj0509:
@ares @lzap At the time of initializing any object, the default taxonomies are being set. This is done by https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L14, which being responsible here in the test case.

One question, do we need to do this every time? https://github.com/theforeman/foreman/blob/develop/app/models/concerns/taxonomix.rb#L14

rabajaj0509:
@lzap irrelevant test is failing from core!

lzap:
Thanks!

PR #405, opened 2018-01-17
Fixes #22297 - [Discovery rules]Fix the pagination

rabajaj0509:
By default 20 items must be displayed on the current page but if you have say 20 items and
want to show only 5 on page, then the pagination does not work correctly.
This patch makes sure that correct number of rows are displayed per page.

### DIFF
--- a/app/controllers/discovery_rules_controller.rb
+++ b/app/controllers/discovery_rules_controller.rb
@@ -8,7 +8,7 @@ class DiscoveryRulesController < ApplicationController
 
   def index
     base = resource_base.search_for(params[:search], :order => (params[:order]))
-    @discovery_rules = base.paginate(:page => params[:page]).includes(:hostgroup)
+    @discovery_rules = base.paginate(:page => params[:page], :per_page => params[:per_page]).includes(:hostgroup)
   end
 
   def new


lzap:
Cool thanks, can you also fix this UI thing:

image

Looks broken, is it just for me?

lzap:
Oh I noticed that the other patch fixes UI.

But I struggle to understand what this patch does fix. Without it, I can paginate just fine, what is wrong here? (It's just UI what is broken).

app/controllers/discovery_rules_controller.rb
```diff
@@ -8,7 +8,7 @@ class DiscoveryRulesController < ApplicationController
 
   def index
     base = resource_base.search_for(params[:search], :order => (params[:order]))
-    @discovery_rules = base.paginate(:page => params[:page]).includes(:hostgroup)
+    @discovery_rules = base.paginate(:page => params[:page], :per_page => params[:per_page]).includes(:hostgroup)
```
tbrisker, line 11: why not use `resource_base_search_and_page` helper which handles all of these parameters?

tbrisker (changes requested):
to get the pagination to display correctly in the UI, use `will_paginate_with_info` helper for the table.

app/controllers/discovery_rules_controller.rb
lzap: We have another patch that touches this part https://github.com/theforeman/foreman_discovery/pull/406

lzap: I remember we had issues because discovery rules page contains search in different context. It did not work out.

rabajaj0509: @lzap yes, the search field disappears when using `resource_base_search_and_page`. @tbrisker can we think of something to avoid the same?

tbrisker: That's very strange, since in core we have:
```ruby
  def resource_base_with_search
    resource_base.search_for(params[:search], :order => params[:order])
  end

  def resource_base_search_and_page(tables = [])
    base = tables.empty? ? resource_base_with_search : resource_base_with_search.eager_load(*tables)
    base.paginate(:page => params[:page], :per_page => params[:per_page])
  end
```
which is exactly the same...

lzap:
Tests are green locally.

PR #406, opened 2018-01-17
Fixes #20450 - Update pagination to patternfly style

rabajaj0509:


lzap:
Nice can you update also discovery rules? That's all we have, works just fine.

lzap:
[test]

app/controllers/discovered_hosts_controller.rb
```diff
@@ -26,7 +26,7 @@ def index
       :organization,
       :model,
       :discovery_attribute_set
-    ], {:interfaces => :subnet}).paginate(:page => params[:page])
+    ], {:interfaces => :subnet}).paginate(:page => params[:page], :per_page => params[:per_page])
```
tbrisker: please use `@hosts = resource_base_search_and_page.includes(....` so if something changes in core we won't have to update this as well.

tbrisker (changes requested):

rabajaj0509:
@tbrisker, works perfectly now.

Looks like:
screenshot from 2018-01-31 17-02-45

tbrisker (approved):
LGTM, Thanks @rahulbajaj0509 !

PR #411, opened 2018-03-06
Fixes #22783 - Audit discovery resources

kgaikwad:
Audit resources - discovery rule, attribute-set & relation with hostgroup

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -1,4 +1,5 @@
 class DiscoveryRule < ApplicationRecord
+  audited :associated_with => :hostgroup
   include Authorizable
   extend FriendlyId
   friendly_id :name
--- a/test/unit/discovery_rule_test.rb
+++ b/test/unit/discovery_rule_test.rb
@@ -136,4 +136,44 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
     refute_valid rule
     assert_equal "Host group location #{loc.name} must also be associated to the discovery rule", rule.errors[:locations].first
   end
+
+  context 'auditing related to discovery rule' do
+    test "should be able to create audit on a rule creation" do
+      rule = nil
+      assert_difference 'Audit.count' do
+        rule = FactoryBot.create(:discovery_rule, :with_auditing)
+      end
+      recent_audit = rule.audits.last
+      assert_equal 'create', recent_audit.action
+    end
+
+    test "should create audit when updating a discovery rule" do
+      rule = FactoryBot.create(:discovery_rule, :with_auditing)
+      assert_difference 'rule.audits.count' do
+        rule.enabled = false
+        rule.save!
+      end
+      audit = rule.audits.last
+      assert_equal 'update', audit.action
+    end
+
+    test "should create audit record when destroying a discovery rule" do
+      rule = FactoryBot.create(:discovery_rule, :with_auditing)
+      assert_difference 'Audit.count' do
+        rule.destroy
+      end
+      audit = rule.audits.last
+      assert_equal 'destroy', audit.action
+    end
+
+    test 'should be able to create audit record on changing hostgroup of rule' do
+      rule = FactoryBot.create(:discovery_rule)
+      hostgroup_one = FactoryBot.create(:hostgroup, :name => "test-hg")
+
+      assert_difference 'hostgroup_one.associated_audits.count' do
+        hostgroup_one.discovery_rules << rule
+        hostgroup_one.save!
+      end
+    end
+  end
 end


app/models/discovery_attribute_set.rb
```diff
@@ -1,5 +1,6 @@
 class DiscoveryAttributeSet < ApplicationRecord
   belongs_to :host, :class_name => "Host::Discovered", :foreign_key => :host_id
+  audited :associated_with => :host
```
lzap: This can theoretically generate a lot of noise if there is some host (buggy for example) which reports change in memory or cpu count or disks amount on every request. Discovered hosts sends facts every 10 minutes or so by default. Just a warning, if you have some 2nd tier audits (things which users can turn off), this is a good candidate.

lzap (commented):
So far good, please add more tests.

ares:
Tested, looks good. Tests failures are unrelated so I think it could be merged.

kgaikwad:
@ares,
Yes, we can merge this PR as mentioned comment was related to attribute_set & host and I already removed attribute_set and host association.

ares:
Ok, perhaps remove WIP from the PR title so @lzap notices :-)

test/unit/discovery_rule_test.rb
```diff
@@ -136,4 +136,26 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
     refute_valid rule
     assert_equal "Host group location #{loc.name} must also be associated to the discovery rule", rule.errors[:locations].first
   end
+
+  test "should be able to create audit on a rule creation" do
+    rule = DiscoveryRule.new :name => "my_audit_rule",
```
rabajaj0509: why not just: 
```
test "create audit record when creating a discovery rule" do
      rule = FactoryBot.create(:discovery_rule, :with_auditing)
      audit = rule.audits.last
      assert_equal 'create', audit.action
    end
```

rabajaj0509:
@kgaikwad  can we add these tests to your pr also? was a little interested to do audit related work, so wrote these :)

Test for an update in a discovery rule:
```
test "create audit record when updating a discovery rule" do 
    rule = FactoryBot.create(:discovery_rule, :with_auditing)
    audit = rule.audits.last
    assert_equal 'create', audit.action
    assert_equal true,audit.audited_changes["enabled"]
    assert_equal "facts.fact1 == 'value1'", audit.audited_changes["search"]

    rule.enabled = false
    rule.search = "facts.fact1 == 'value2'"
    rule.save!
    audit = rule.audits.last
    assert_equal 'update', audit.action
end
```

Test for deletion of any discovery rule:
```
test "create audit record when destroying a discovery rule" do
    rule = FactoryBot.create(:discovery_rule, :with_auditing)
    rule.destroy
    audit = rule.audits.last
    assert_equal 'destroy', audit.action
end
```

rabajaj0509:
@lzap would it be nice to have a layout like it is in core (model tests go to test/models and not test/unit; controller tests to test/controllers/ and not test/integration, as @timogoebel has suggested). Can we start that with this pr?

kgaikwad:
@rahulbajaj0509,

Thank you for sharing more test cases. 
While adding test cases, I also searched for test/models folder first :-)
It would be nice to have similar layout like core but I don't want to block audit changes for test layout.

As you are more familiar with discovery stuff,  I would suggest you to have this layout changes in separate PR so that you can move existing tests from test/unit to test/models.  

@lzap, please give your thoughts on above discussion.

ares:
I agree that changing the test layout should be separate refactoring, if we later decide to cherry-pick this PR to stable branch, it would cause unnecessary pain.

rabajaj0509:
@kgaikwad @ares fair enough :) changes in the test layout should be a separate refactoring :+1:

kgaikwad:
@ares , moved `audited` method at the top & added more test cases.

app/models/discovery_rule.rb
```diff
@@ -5,6 +5,9 @@ class DiscoveryRule < ApplicationRecord
   include Parameterizable::ByIdName
   include Taxonomix
 
+  audited :associated_with => :hostgroup
```
ares: I think this will need to be first line of the model soon

ares (dismissed):
except for the definition moving, I think it's ready

app/models/discovery_rule.rb
```diff
@@ -1,4 +1,6 @@
 class DiscoveryRule < ApplicationRecord
+  audited :associated_with => :hostgroup
+  has_associated_audits
```
tbrisker: Does this have any other models associated with it? `has_associated_audits` is needed when some other resource defines its audits with, in this example, `:associated_with => :discovery_rule`

tbrisker (commented):
one question in line, other than that lgtm.

lzap:
@rahulbajaj0509 as @ares said refactoring needs to be a separate PR to prevent regressions. Let's get tests fixed first, then merge this PR and refactor on top.

rabajaj0509:
@lzap  yes, agree to that :+1:

lzap:
Develop is now green again, please rebase as long it appears green in CI:

http://ci.theforeman.org/job/test_plugin_foreman_discovery_develop/

app/models/discovery_rule.rb
kgaikwad: Yes, good catch!!
It is not needed here as I have removed audit part from `discovery_attribute_set`.

tbrisker (approved):
Ran tests locally and they pass, thanks @kgaikwad !

ares:
This seems to add more broken tests on Ruby 2.4

EDIT: I don't see it any longer, could have been some problem on jenkins

lzap:
Yeah some are related to:

https://community.theforeman.org/t/stubbing-global-objects-outside-of-tests-cause-random-failures/8647

Others - no idea (random failures)-

PR #416, opened 2018-03-29
Fixes #23056 - Host discovery list uses wrong icon for new, reported and non reported statuses

ik5:


ohadlevy:
please add a before/after image - thanks :)

ik5:
@ohadlevy Worked on doing just that :)

before:
gnome-shell-screenshot-b84cgz

After:
gnome-shell-screenshot-gmzngz

Rohoover (commented):
Is the info icon intentionally green as opposed to grey or blue?

ik5:
@Rohoover I kept the colors, because I do not know what colors to place there instead.

lzap:
If I look on the old screenshot, I can tell that the plus icon represents new hosts recently added. Now, on the proposal I can't immediately recognize difference between info and check. Do you have better ideas?

Rohoover (commented):
I would change the info icon to blue to align with PF and provide an additional visual differentiator from the check. I believe this is the correct color: 0088ce

Sorry for the delay, I was on PTO.

ik5:
@Rohoover how is that?
new host discovery icons color info

No green anymore, just two shades of blue and a a red for warning

lzap:
Let me explain the three icons here:

* Host is up to date - discovery host reported in last 24 hours and everything is ok
* Host was recently added - the above plus the report is 4 hours old (maximum)
* Host is down - discovery host did not respond in last 24 hours (something is going on - perhaps it was turned off or network issue)

lzap:
Now, I am trying to find best icons in the set: https://www.patternfly.org/styles/icons/

But I think what we have represents the above best, but maybe you have better idea. I don't like info icon, it's not telling anything in particular and it's too generic. Similarly to container icon, also you can imagine anything. How about:

* Up to date: pficon-on-running
* Recently added: pficon-on-running (in some different color or in bold if we have that)
* Unknown / down: pficon-pending

ohadlevy:
thats interesting, should we change the other host power status (e.g. in
host list) to be using pficon-on-running?

lzap:
Yeah I kinda like it. If folks don't like colors, we can go with pficon-on-running in gray for up-to-date hosts and pficon-on-running in black for recently added to draw attention on them.

ik5:
@lzap, @ohadlevy 
I did made some checking before hand of the icons, I didn't find any better icons to use.
I think the icons on patternfly are not that much, and because of it there are not much to choose from.

Also note that the [patternfly page](http://www.patternfly.org/styles/icons/) uses a table that explain the type of usage for most icons.

lzap:
We need to step out of PatternFly constraints here, we have special situation - there's OK host, FRESH host and FAILED host. PatternFly only provides OK icon, we need something that's actually "above OK", for this reason we need to ditch PF's OK because that's too much IMHO. Here is another idea:

* pficon-on-running - means "on and running" - for "recent hosts"
* pficon-on - means "on" - for "ok hosts"
* pficon-paused (or pending) - means "paused or pending" - for "out of sync" hosts

ohadlevy:
@lzap lets wait for @rohoover feedback regarding which icon to use....

Rohoover (commented):
So I saw a few statuses listed with the definitions colliding at time. It's important to clearly define these status and not conflate them because the iconography changes. Below is a list of what I'm seeing described by all the different posters. I'm also for brevity including what icon I would put with each status, even though not all those statuses are valid.

Host (on and running) - pficon-on-running
Host (on and okay) - pficon-ok
Host Pending - pficon-pending
Host Paused - pficon-paused
Host Error - pficon-error-circle-o
Host Warning - pficon-warning-triangle-o

Let's confirm which are valid and it's clear definition. For example, I think it's important to use the pending icon, when something is actually pending, not when a status is unknown.

Also there is a list of PF icons but we can also potentially use icons from Font Awesome, which is the base set that PF uses. 

@ohadlevy The PF icon for ok and running is relatively new, at least since we did the original host set, so I do recommend using it if it applies.

As far as colors, this is how we assign colors from UX:
Red = Error
Yellow = Warning
Blue = Interesting or primary action
Grey = Not as interesting, not primary

ik5:
@Rohoover , @lzap , @ohadlevy 
If I understand this discussion right, you are referring to this type of look, right?
tmp-colors

Rohoover (commented):
Yes, though I would comment that "on and running" may be green. Green is telling the user things are fine, nothing to worry about. I would ask oneself "Which status best represents that?"

Is the power icon meaning the same as just 'on and okay'?

ik5:
Green in this case was on a newly added server, and that's imo is not "good and running", but rather "discovered and known", so I'm not sure that green is the right color here.

I'm adding the code here, though, as seen on the screenshot

Rohoover (commented):
@ik5 

I think we are agreed. On and running = green, new = blue.

lzap:
@ik5 can you please add another screenshot with the final version so we can all approve here, I see there was another commit after last screenshot. Want to be sure we all comment the correct picture. Thanks!

ik5:
Here it is:
new host discovery icons color info2

app/helpers/discovered_hosts_helper.rb
```diff
@@ -56,17 +56,17 @@ def authorized_for_edit_destroy?
...
+      status_glyph = 'pficon-on-running'
       status_message = _('New in the last 24 hours')
-      status_color = '#89A54E'
+      status_color = '#0088ce'
     elsif host.last_report.present? && host.last_report < 7.days.ago
-      status_glyph = 'glyphicon-exclamation-sign'
+      status_glyph = 'pficon-paused'
       status_message = _('Not reported in more than 7 days')
-      status_color = '#AA4643'
+      status_color = ' #ec7a08'
```
lzap: Leading space, other than that I am good thanks.

ik5: fixed, thanks

lzap:
Thanks.

PR #364, opened 2017-07-17
Fixes #20126 - Enable selective reboot and autoprovision

rabajaj0509:

PR #417, opened 2018-04-01
Refs #23060 - Refactor Reboot All and Auto Provision All

ik5:
One important note: I was unable to figure out yet how to enable the drop down if no item was selected.

### DIFF
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -30,13 +30,27 @@ module DiscoveredHostsHelper
   end
 
   def multiple_discovered_hosts_actions_select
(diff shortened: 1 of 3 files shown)


ik5:
Before:
original discovered host reboot provision all buttons

After:
new discovered host reboot provision all buttons

rabajaj0509:
Thanks @ik5, I guess I have a better solution for this issue, only the tests are failing which I am not being able to fix. Would you mind having a look at https://github.com/theforeman/foreman_discovery/pull/364/files ? It will be great if we can fix that, the code works just fine but the tests are a little tricky to get away with :)

ik5:
@rahulbajaj0509 sure, I'll take a look and see how I can help :)

lzap:
I merged https://github.com/theforeman/foreman_discovery/commit/2c5cbe1724da5a76e6ab9c9d0cd943e91768c9df today which actually move this to the correct place. I am going to close this now. Thanks!

PR #418, opened 2018-04-02
Fixes #23073 - Redundant info on the Highlights

ik5:
Original:
original disocvery ipaddress

New:
new disocvery ipaddress

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -34,6 +34,8 @@ class DiscoveredHostsController < ::ApplicationController
   end
 
   def show
+    # translating the names of highlights keys
+    @highlights_translated_keys = { ipaddress: _('primary ipaddress') }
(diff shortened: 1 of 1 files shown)


lzap:
I think we should rather discuss this in the issue, but let's do it here: I don't think that Foreman should be trying to "explain" puppet facts, lots of puppet fact names are "weird" and I like consistency - the same name is used when puppet reports back in core or when you call `facter` command manually.

This fact is actually not important item for discovery, we have discovery_bootif which identifies the primary interface and ipaddress is therefore ignored. The fact also seems to be legacy now: https://puppet.com/docs/facter/3.9/core_facts.html#ipaddress

I'd rather kept it as is.

ik5:
Closing the PR over @lzap comment.

It is a hackish solution, because the data arrives from the DB.

lzap:
Thanks, other UX comments in the document are good.

PR #419, opened 2018-04-02
Refs #23078 - Changing UI locations of "Expand All" button, and accordion with interfaces

ik5:
Before:
original discoverws host ui

After:
new discoverws host ui

### DIFF
--- a/app/views/discovered_hosts/show.html.erb
+++ b/app/views/discovered_hosts/show.html.erb
@@ -18,14 +18,17 @@
       <div id="highlights-panel" class="panel-collapse collapse facts-panel in" >
         <table class="<%= table_css_classes('table-condensed') %>">
           <% @categories[0].sort.each do |key, val| %>
-          <tr id="fact-<%= key.try(:downcase) %>" class="">
-            <th class="ellipsis" width="40%"> <strong> <%= key %> </strong></th>
-            <td><%= val %></td>
-          </tr>
-        <% end -%>
+            <tr id="fact-<%= key.try(:downcase) %>" class="">
+              <th class="ellipsis" width="40%"> <strong> <%= key %> </strong></th>
+              <td><%= val %></td>
+            </tr>
+          <% end -%>
         </table>
       </div>
     </div>
+  </div>
+
+  <div class="col-md-6">
     <% unless @interfaces.empty? %>
       <div class="panel panel-default">
         <div class="panel-heading" role="tab">
@@ -56,12 +59,20 @@
       </div>
     <% end %>
   </div>
+</div>
 
-<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
-    <div class="col-md-6">
+<div class="row">
+  <div class="col-md-12">
+    <%= link_to(_("Collapse All"),"#", :id => "expand_all", :class => "btn btn-default pull-right", style: 'margin-bottom: 1rem') -%>
+  </div>
+</div>
+
+<div class="row">
+  <div class="col-md-12">
+    <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
       <% @categories.each_with_index do |val, index| %>
         <% next if index == 0 || val.empty? %>
-        <div id="category-<%= @categories_names[index].downcase %>" class="panel panel-default">
+        <div id="category-<%= @categories_names[index].downcase %>" class="panel panel-default" style="margin-bottom: 1rem">
           <div class="panel-heading" role="tab">
             <h4 class="panel-title">
               <a role="button" class="expendable-link" data-toggle="collapse" data-parent="#accordion" href="#<%= @categories_names[index].to_s + "panel" %>" aria-expanded="true">
@@ -69,14 +80,14 @@
               </a>
             </h4>
           </div>
-          <div id="<%= @categories_names[index].to_s + "panel" %>" class="panel-collapse collapse facts-panel" role="tabpanel">
+          <div id="<%= @categories_names[index].to_s + "panel" %>" class="panel-collapse collapse in facts-panel" role="tabpanel">
             <table class="<%= table_css_classes('table-condensed table-fixed') %>">
-            <% val.sort.each do |key, value|  %>
+              <% val.sort.each do |key, value|  %>
                 <tr class="">
                   <th width="40%" class="ellipsis"><strong> <%= key %> </strong> </th>
                   <td class="ellipsis"> <%= value %> </td>
                 </tr>
-            <% end %>
+              <% end %>
             </table>
           </div>
         </div>
@@ -100,9 +111,10 @@
     $('.primary-flag').tooltip();
     $('.provision-flag').tooltip();
 
+    expand_all_toggle_action = { 'Expand All': 'show', 'Collapse All': 'hide' };
     $('#expand_all').on('click', function () {
-      $('#accordion .panel-collapse').collapse('toggle');
       $(this).text(function(i, text){
+          $('#accordion .panel-collapse').collapse(expand_all_toggle_action[text]);
           return text === "Expand All" ? "Collapse All" : "Expand All";
       })
     });
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -20,10 +20,6 @@ module DiscoveredHostsHelper
         ),
         select_action_button( _("Select Action"), {}, provision_button(host, hash_for_edit_discovered_host_path(:id => host)), actions.map { |action| display_link_if_authorized(action[0] , action[1], action[2] || {}) }.flatten ),
       button_group(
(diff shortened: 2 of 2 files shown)


app/views/discovered_hosts/show.html.erb
```diff
@@ -56,9 +59,12 @@
...
   </div>
+</div>
 
-<div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
-    <div class="col-md-6">
+<div class="row">
+  <div class="col-md-12">
+    <div class="panel-group" id="accordion" role="tablist" aria-multiselectable="true">
+      <%= link_to(_("Expand All"),"#", :id => "expand_all", :class => "btn btn-default") -%>
       <% @categories.each_with_index do |val, index| %>
```
rabajaj0509: @ik5 +1 on changing the position of the `Expand All` button. One question, can we have the `Expand All` button on the right side as it will maintain consistency?

Also, if you have time, can you fix the accordion, we use toggle for the `Expand all` button. If you open `Hardware` tab and then press `Expand All` button you will see that the `Hardware` tab will close.

ik5: done for both :)

rabajaj0509 (approved):
Thanks @ik5  :+1:

lzap:
Thanks, please change the commit message to "Fixes #23078 - ..." and make the message shorter than 70 characters.

I like it, yeah. Can you add a little bit of spacing between the accordions? It's glued together, does not look good.

Also here is an idea - I'd love to have it expanded by default.

ik5:
@lzap done, what do you think of it now?

app/assets/stylesheets/discovery/show.scss
```diff
@@ -0,0 +1,3 @@
+.panel-space {
+  margin-bottom: 1rem !important;
+}
```
lzap: Can you pull this inline? Discovery is a small plugin, we won't have much UX and this actually involves some packaging work I would like to avoid.

app/helpers/layout_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    def stylesheet(*args)
+      args.map do |arg|
+        content_for(:stylesheets) { stylesheet_link_tag(arg) }
+      end
+      return ""
+    end
+
+  end
+end
```
lzap: Module/class vs dir/file names. But we are getting rid of it anyway.

lzap (changes requested):
First of all - it looks great! This is indeed an improvement.

I am hesitant to start precompiling assets just because of one line. I know this is good direction long term, but we have how many - three to five inline CSS styles? Please remove it from here.

If you want to work towards direction, I am good with this but in a separate (refactor) ticket, also review all our codebase and put everything into SCSS. We will need also some packaging changes I will guide you through.

app/views/discovered_hosts/show.html.erb
```diff
@@ -1,3 +1,6 @@
+<% content_for(:stylesheets) do %>
+  <%= stylesheet 'discovery/show' %>
```
ohadlevy: core has a mechanism for this already, not sure that you need to write your own..
example at https://github.com/theforeman/foreman_openscap/blob/94eea5947265ed270b0176f6f94ca80d5923dab4/app/views/arf_reports/show.html.erb#L3

config/initializers/assets.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.assets.precompile += %w( discovery/show.css )
```
ohadlevy: please use the plugin interface to handle this.. see https://github.com/theforeman/foreman_openscap/blob/94eea5947265ed270b0176f6f94ca80d5923dab4/lib/foreman_openscap/engine.rb#L31

app/assets/stylesheets/discovery/show.scss
ohadlevy: if you are adding more css, maybe its best to move straight to webpack based css vs asset pipeline. I would search if none of our existing patternfly predefined classes provide a similar experience

ohadlevy:
also, please make sure that any UI changes has @rohoover tagged.

app/assets/stylesheets/discovery/show.scss
lzap: Let's create a refactoring ticket please, I am a bit fuzzy about webpack, @ik5 are you able to create a ticket and describe what's needed? I would just copy and paste @ohadlevy 's text from here ;-)

lzap:
Thanks @ohadlevy for reviewing UX parts, I had no idea...

app/assets/stylesheets/discovery/show.scss
ik5: The thing is, that I dislike inline css, it's a code smell, I saw the inline css, and that's why I created the asset pipeline.

I will create a new ticket for it.
@ohadlevy one thing that I talked with @iNecas is that the whole move to webpack, and babel for moving into react instead of the current jquery, and such move, as you suggested, is the beginning for it, because it creates the infrastructure for it.

What do you think?

app/views/discovered_hosts/show.html.erb
ik5: I saw how it was done in other plugins (katello), I'm going to change all of this now.

ohadlevy: if its done incorrectly there, please send a PR to fix it too :)
also, not sure if you saw it, https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin has a lot of information about services that core provider for plugins

app/assets/stylesheets/discovery/show.scss
ohadlevy: we can rewrite parts of discovery - I actually think the whole index page should move from a table to a list view https://www.patternfly.org/pattern-library/content-views/list-view/, and thats probably best discussed with @rohoover first.

IMHO you can just use webpack/scss/es6 etc as core does, discovery has a fairly small UI footprint and we could easily convert it.

lzap: > What do you think?

I think let's not do it in this PR, I would like to have the small improvements finished first, because the whole refactoring can take some time. We can take the time for proper redesign.

ik5: @lzap i didn't meant to do it on this PR :)
But the whole idea that was talked about here.

app/views/discovered_hosts/show.html.erb
ik5: I have read and used that page, that's how I was able to make foreman compile it for me :)

config/initializers/assets.rb
ik5: removed it from at the moment, will do it properly on the next pr we talked about.

ik5:
@lzap, @ohadlevy, I'ved opened a new ticket regarding the moving to a better infrastructure regarding asset pipeline.
http://projects.theforeman.org/issues/23200

I will do in on few small PR rather then one big PR, not related to this one.

Rohoover (commented):
I would keep the Expand button on the right hand side above the accordion. That would be consistent with how we have other buttons placed on the page. Also, I would add a little padding so it's not touching the accordion.

ik5:
How is the current spacing (the default state is expand, but I collapsed it to show how it looks)?

new discovery host details

Rohoover (commented):
LGTM! Thank you!

lzap (approved):
Ok thanks, we are working on fixing discovery tests, then we can merge this one.

iNecas:
It seems the master tests are green, perhaps needs a rebase?

ik5:
@iNecas rebased, I noticed that it broke several open PR's in discovery, so I hope that will fix it.

app/views/discovered_hosts/show.html.erb
```diff
@@ -18,14 +18,17 @@
...
-        <% end -%>
+            <tr id="fact-<%= key.try(:downcase) %>" class="">
+              <th class="ellipsis" width="40%"> <strong> <%= key %> </strong></th>
+              <td><%= val %></td>
+            </tr>
+          <% end -%>
         </table>
       </div>
     </div>
+  </div>
```
lzap, line 29: This is an extra div, is this intentional?

app/views/discovered_hosts/show.html.erb
```diff
@@ -56,27 +59,35 @@
       </div>
     <% end %>
   </div>
+</div>
```
lzap, line 62: And another one, I guess I will be counting these.

lzap: Yup, all match correctly.

lzap (approved):
Looks nice, thanks!

lzap:
Minor comment - we tend to keep our initial line of the commit shorter than 70 characters, with "Fixes #XXXX" this gives you like 50 chars, it's tough but keep it short and explain below. Thanks!

PR #420, opened 2018-04-09
Fixes #23170 - Numeric input fields does not use the standard style

ik5:
Original:
original - discovery numeric fields

New:
new - discovery numeric fields

### DIFF
--- a/app/views/discovery_rules/_form.html.erb
+++ b/app/views/discovery_rules/_form.html.erb
@@ -23,11 +23,17 @@
             <%= text_f f, :hostname,
                        :label_help => render('discovery_rules/template_inline').html_safe,
                        :label_help_options => {:title => _("Hostname for provisioned hosts"), :'data-placement' => 'bottom' } %>
-            <%= number_f f, :max_count, :label => _('Hosts Limit'), :help_inline => _('Maximum hosts provisioned with this rule (0 = unlimited)'), :min => 0, :max => (2**31 - 1) %>
-            <%= number_f f, :priority, :help_inline => _('Rule priority (lower integer means higher priority)'), :min => 0, :max => (2**31 - 1) %>
+            <%= counter_f f, :max_count, :label => _('Hosts Limit'), :help_inline => _('Maximum hosts provisioned with this rule (0 = unlimited)'), :min => 0 %>
+            <%= counter_f f, :priority, :help_inline => _('Rule priority (lower integer means higher priority)'), step: 100, value: 0 %>
             <%= checkbox_f f, :enabled %>
           </div>
           <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @discovery_rule %>
         </div>
     <%= submit_or_cancel f %>
 <% end %>
+
+<script type="text/javascript" charset="utf-8">
+$(document).ready(function() {
+  $('input.counter_spinner').spinner();
+});
+</script>


ik5:
Note: Foreman does not use patternfly spinner for it.

ohadlevy:
I believe this is a per browser behavior ? you should still get the up/down arrow AFAIU?

and IMHO, if this fix is to be accepted, it should be in core, not discovery, and not in a script tag :) thanks

ik5:
@ohadlevy, yes the real spinners are still there based on browser.

The current "fix", is that it uses the same UI as foreman.

ohadlevy:
I don't follow, do you have an idea why they are different? which code in
foreman does not get executed? can you point me to a place where in foreman
it works correctly?

ik5:
@ohadlevy 
Let me provide some background.

foreman works well. This is the change for foreman_discovery (as you know).
There was an old discussion about issues of the discovery plugins UI.

One of the raised issues, was that the priority field (the only numeric field back then) does not use the patternfly touchspin component.

The thing is that the jquery version of touchspin is not supported by foreman, but jquery-ui-spinner does.

Do you think that I should not use any spinner at all, or am I missing something in my understanding?

ohadlevy:
Thanks for the background, I think I'm failing to understand why this only
effects discovery.... hence I'm looking for the difference

ik5:
@ohadlevy there is no difference, I was sure that it works different on foreman but went to check it now, and it seems that foreman is the same as the original, now I understand your point.

ares:
Please note that there is already a spinner helper in core - https://github.com/theforeman/foreman/blob/develop/app/helpers/form_helper.rb#L224-L229

see how it's used e.g. at https://github.com/theforeman/foreman/blob/9234d6c5e60c54129b01d45ee7d91af76e492c2a/app/views/compute_resources_vms/form/libvirt/_base.html.erb#L5

I'd vote for reusing of what we have in core. If we want to change how the spinner works, let's fix the core helper, which is orthogonal to this PR I think.

ik5:
Thank you Marek, I'ved changed to this version.

lzap:
The last version does not work for me - I still see text are. Is there some switch I need to check?

```html
<div class="form-group "><label class="col-md-2 control-label" for="max_count">Hosts Limit</label><div class="col-md-4"><input min="0" max="2147483647" class="form-control  counter_spinner" type="text" value="0" name="discovery_rule[max_count]" id="discovery_rule_max_count"><span class="help-block"><span class="maximum-limit hidden"><span class="pficon pficon-warning-triangle-o "></span> <span class="error-message"> Specified value is higher than recommended maximum</span></span></span></div><span class="help-block help-inline">Maximum hosts provisioned with this rule (0 = unlimited)</span></div>
```

ik5:
I'm not sure that I understand what you meant here. It's text, but should
be as foreman is currently works.

ares:
A screenshot might help. IIRC the `counter_f` adds arrows next to the text field if it works correctly (e.g. cpu number in host form when provisioning on libvirt). If no arrows are displayed here, it might be required to load some JS file on the page.

lzap:
@ik5 what I mean is that I see no arrows, just a plain and regular text box.

ik5:
@lzap the `counter_f` uses `text` type instead of `numeric` so the browser does not place there the buttons.

ares:
this is how the `counter_f` looks in core, if it looks differently in discovery, something is not working
counter_f

@ik5 did you check JS console?

lzap:
Yeah I kinda expected something like that. That is purpose of this patch, isn't it?

ik5:
I'm actually investigating it now, I thought it should work like this, but
made better investigation, and on some foreman pages it is like so, and
with others it is with up and down buttons, so looking at it to figure out
what I'm missing.

ik5:
While debugging, I still haven't found the reason, but i found out that only the VMs in foreman uses `counter_f`.

@ares are you sure that it is the right field for it?

ik5:
solved, @lzap do you see it like so?
screenshot-20180422143305

ares:
> @ares are you sure that it is the right field for it?

well, that's the only spinner field I'm aware of, it's still using jquery but my hope is that if you use this core helper, if later it's converted to react, it will still work

ik5:
There was a jquery event to capture and execute a callback for it, a hint
from avi helped me to start looking for it.

lzap (changes requested):
Thanks,

what's `max` is used for? I don't see that used in core.

There is a problem with priority, we never said it needs to be positive integer, it's just an integer and users might be using negative values, it makes much sense to use it. I think our limited spinner is hardcoded to have minimum value of "1". We want to allow negative values here.

I think we should use ui spinner directly. Also here is one tip - for priority we want steps to be 100 it is good practice to create wide spread values - remember BASIC programming? :-)

ik5:
@lzap `maz` is for the maximum numeric value it can have. 

I think we can ignore min regarding `priority` or place a negative range, for example a 16bit signed int range, or not to limit it at all.

> I think we should use ui spinner directly. Also here is one tip - for priority we want steps to be 100 it is good practice to create wide spread values - remember BASIC programming? :-)

So if I understand you currectly, you are looking for a series of `1, 100, 200, 300` etc?

Btw One does not talk about BASIC (`goto 100`) ;)

lzap:
Can you show me where exactly max value is used? Do we need to set it explicitly? I hope for sane default, if you don't set it.

Don't follow your comment on priority, users today use priority set to negative values, this patch makes this pattern impossible with the spinner. I think better approach would be to not to do any constraints on limits at all - full integer range should be possible.

Yeah I was thinking more consistent: 0, 100, 200...

ik5:
Okay, so I'll remove both `min` and `max`, and add support for `step`

lzap:
Thanks even the softlimit makes no sense, I hope the component works just fine if you don't provide those at all - it should if not let's fix this.

ik5:
No limits, and using step of 100 both signed and unsigned.

app/views/discovery_rules/_form.html.erb
```diff
@@ -23,11 +24,21 @@
...
+            <%= counter_f f, :priority, :help_inline => _('Rule priority (lower integer means higher priority)'), step: 100, value: 0 %>
             <%= checkbox_f f, :enabled %>
           </div>
           <%= render 'taxonomies/loc_org_tabs', :f => f, :obj => @discovery_rule %>
         </div>
     <%= submit_or_cancel f %>
 <% end %>
+
+<%#
+  calling it by hand, because the tfm.numFields.initAll() set a minimum of 1,
```
ares: Please don't use comments in erb, deface gem does not work with it and fails with very cryptic messages. If the change needs comment, put it to github PR or redmine issue.

ik5: sorry, removed them.

ares: no worries, I remember how disappointed I was when I spent whole day to figure out the cause, looking into libxml native extensions :-)

lzap: There's no deface in discovery anymore FYI.

app/views/discovery_rules/_form.html.erb
```diff
@@ -23,11 +24,17 @@
             <%= text_f f, :hostname,
                        :label_help => render('discovery_rules/template_inline').html_safe,
                        :label_help_options => {:title => _("Hostname for provisioned hosts"), :'data-placement' => 'bottom' } %>
-            <%= number_f f, :max_count, :label => _('Hosts Limit'), :help_inline => _('Maximum hosts provisioned with this rule (0 = unlimited)'), :min => 0, :max => (2**31 - 1) %>
-            <%= number_f f, :priority, :help_inline => _('Rule priority (lower integer means higher priority)'), :min => 0, :max => (2**31 - 1) %>
+            <%= counter_f f, :max_count, :label => _('Hosts Limit'), :help_inline => _('Maximum hosts provisioned with this rule (0 = unlimited)'), :min => 0, :max => max_count, 'data-soft-max': max_count %>
```
lzap: Assuming that priority will work without any limits (will test it later), can you also remove max limit completely from max_count as well? Just keep min set to zero, that is reasonable. This way we can get rid of INT_MAX weird thing, it's the component that should provide sane default, not us.

lzap (changes requested):
One more to go, then I am gonna test :-)

app/views/discovery_rules/_form.html.erb
ik5: done :)

lzap (approved):

lzap:
"Environment data not found in the schema. To resolve this issue, run: " failure irrelevant. Thanks!

PR #421, opened 2018-04-11
Refs #23200 - Moving inline code to external scss files

ik5:
This stage contains the "bootstrap" for all the node stack, nothing more

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,48 @@
+{
+  "root": true,
+  "extends": "airbnb-base",
+  "plugins": [
+    "react"
+  ],
+  "env": {
+    "browser": true,
+    "es6": true,
+    "node": true,
+    "jasmine": true,
+    "jest": true
+  },
+  "globals": {
+    "document": false,
+    "escape": false,
+    "navigator": false,
+    "unescape": false,
+    "window": false,
+    "$": true,
+    "_": true,
+    "__": true
+  },
+  "parser": "babel-eslint",
+  "rules": {
+    "react/jsx-uses-vars": "error",
+    "react/jsx-uses-react": "error",
+    "no-unused-vars": [
+      "error",
+      {
+        "vars": "all",
+        "args": "none"
+      }
+    ],
+    "no-underscore-dangle": "off",
+    "no-use-before-define": "off",
+    "import/prefer-default-export": "off",
+    "import/no-extraneous-dependencies": [
+      "error",
+      {
+        "devDependencies": true
+      }
+    ],
+    "import/no-unresolved": "off",
+    "import/extensions": "off"
+  }
+}
+
--- a/app/views/discovered_mailer/_discovered_host.html.erb
+++ b/app/views/discovered_mailer/_discovered_host.html.erb
@@ -1,10 +1,9 @@
-<% td_style = "text-align: left; border-collapse: collapse; background-color: #FFFFFF; padding: 4px; border: 1px solid #cccccc;" %>
-<td style="<%= td_style%>">
+<td class="discovery-mailer-td-style">
   <%= link_to host, discovered_host_path(:id => host, :host => @url.host, :port => @url.port, :only_path => false, :protocol => @url.scheme) %>
 </td>
-<td style="<%= td_style%>"><%= host.try(:hardware_model_name) || 'N/A' %></td>
-<td style="<%= td_style%>"><%= host.ip %></td>
-<td style="<%= td_style%>"><%= discovery_attribute(host, :cpu_count) %></td>
-<td style="<%= td_style%>"><%= number_to_human_size(discovery_attribute(host, :memory, 0) * 1024 * 1024) %></td>
-<td style="<%= td_style%>"><%= discovery_attribute(host, :disk_count) %></td>
-<td style="<%= td_style%>"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
+<td class="discovery-mailer-td-style"><%= host.try(:hardware_model_name) || 'N/A' %></td>
+<td class="discovery-mailer-td-style"><%= host.ip %></td>
+<td class="discovery-mailer-td-style"><%= discovery_attribute(host, :cpu_count) %></td>
+<td class="discovery-mailer-td-style"><%= number_to_human_size(discovery_attribute(host, :memory, 0) * 1024 * 1024) %></td>
+<td class="discovery-mailer-td-style"><%= discovery_attribute(host, :disk_count) %></td>
+<td class="discovery-mailer-td-style"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
--- a/app/views/discovered_mailer/discovered_summary.html.erb
+++ b/app/views/discovered_mailer/discovered_summary.html.erb
@@ -4,23 +4,22 @@
 <head>
   <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
   <title><%= _('Summary report for discovered hosts from Foreman') %></title>
+  <%= stylesheet_link_tag 'foreman_discovery/email_summary' %>
 </head>
-<body style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; color: #3f3f3f; background-color: #f1f1f1; padding: 10px 24px">
-<h2 style="font-weight: normal; text-transform: uppercase; font-size: 120%;"><%= _("<b>Foreman</b> Discovered hosts summary").html_safe %></h2>
-<h2 style="margin: 5px 0px;"><%= _("Summary from %{time} ago to %{now}") % { :time => @time.in_time_zone, :now => Time.zone.now } %></h2>
-<h3 style="margin: 0px;"><%= _("Discovered hosts from Foreman server at %{foreman_url}") % {:foreman_url => Setting[:foreman_url]} %></h3>
-<div style="background: #fff; padding: 10px 20px; border: 1px solid #e1e2e3">
+<body>
+<h1 class="title"><%= _("<b>Foreman</b> Discovered hosts summary").html_safe %></style>
+<h2 class="summary_title"><%= _("Summary from %{time} ago to %{now}") % { :time => @time.in_time_zone, :now => Time.zone.now } %></h2>
+<h3 class="discovered_host_title"><%= _("Discovered hosts from Foreman server at %{foreman_url}") % {:foreman_url => Setting[:foreman_url]} %></h3>
+<div class="table_pane">
   <% if @discovered_hosts.any? %>
-    <% table_style = "min-width: 400px; border: 1px solid #cccccc; padding: 5px; background-color: #ffffff; border-collapse: collapse; text-align: justify; width: 100%;" %>
-    <% th_style = "text-align: left; border-collapse: collapse;  background: #f5f5f5; background: linear-gradient(#f5f5f5, #e8e8e8); padding: 4px; border: 1px solid #cccccc;" %>
-    <table style="<%= table_style %>">
-      <th style="<%= th_style %>"><%= _('Name') %></th>
-      <th style="<%= th_style %>"><%= _('Model') %></th>
-      <th style="<%= th_style %>"><%= _('IP Address') %></th>
-      <th style="<%= th_style %>"><%= _('CPUs') %></th>
-      <th style="<%= th_style %>"><%= _('Memory') %></th>
-      <th style="<%= th_style %>"><%= _('Disk count') %></th>
-      <th style="<%= th_style %>"><%= _('Disks size') %></th>
+    <table>
+      <th><%= _('Name') %></th>
+      <th><%= _('Model') %></th>
+      <th><%= _('IP Address') %></th>
+      <th><%= _('CPUs') %></th>
+      <th><%= _('Memory') %></th>
+      <th><%= _('Disk count') %></th>
+      <th><%= _('Disks size') %></th>
       <% @discovered_hosts.each do |host| %>
         <tr>
           <%= render :partial => "discovered_host", :locals => {:host => host} %>
--- a/package.json
+++ b/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "foreman_discovery",
+  "version": "1.0.0",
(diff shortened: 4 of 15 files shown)


package.json
```diff
@@ -0,0 +1,42 @@
...
+    "eslint-plugin-jsx-a11y": "^6.0.3",
+    "eslint-plugin-react": "^7.7.0",
+    "jest": "^22.4.3",
+    "jest-cli": "^22.4.3",
+    "stylelint": "^9.2.0",
+    "stylelint-config-standard": "^18.2.0",
+    "uglifyjs-webpack-plugin": "^1.2.4"
+  },
+  "dependencies": {
+    "babel-polyfill": "^6.26.0",
```
ohadlevy: I would suggest not to load these unless you are adding code that uses it.

ik5: I wish to start moving stuff slowly to react, that's why I'ved added them, at the moment I want to start with supporting of scss loading using `.js` file.

ik5:
Okay it looks great (from my side), but one thing is that I'm not sure here, does this package requires the tranformation and polyfills of babel, or will it be implemented by the main foreman asset manager.

lzap:
Thanks, honestly this patch is terrifying for me. I don't understand much of JavaScript and I foresee packaging and production issues with that. Are you willing to do all the packaging changes and production testing?

But first let's get discovery tests fixed, then merge most of the other PRs and then let's talk more about this.

Also this change should pull all the javascript which we have, which is exactly one inline block. I am not really sold in this.

Why you use Refs instead Fixes? I see a separate issue for this. Should be Fixed imo.

ik5:
@lzap the changes are compatible with the way foreman is going. As I understand only new things or dependencies that obsolete are moving from jQuery to React, while I was starting to add features I have already found such cases.

My task here is first to create such migration, and then the second task is to separate the inline Javascript and the inline CSS code into an external pipeline rather then Rails asset pipeline the way Foreman is currently working (aka stage 2).

Note that at this point, the every Javascript and CSS changes are inline rather then using external files, so it will create a much cleaner code that easier to maintain in the long run, and create a bit more DRY code in a whole.

The actual dependencies that I'm using, are also installed by foreman, and at this time I'm trying to figure out if I need to add that dependencies also on my own or just use them, and load them at the code directly.

I am willing to get all the responsibility for this changes (otherwise I would not have taken this task on myself :) )

ik5:
@sharvit can you help, and review this PR to see that for "bootstrapping" at least the css for now, it is good?

Thank you

sharvit:
@ik5 It looks fine but I wouldn't merge it before I see it is actually doing the job.

Need to load the relevant compiled js and css into your views/layouts and see if it is actually work.
see: 
https://github.com/theforeman/foreman/blob/develop/app/views/layouts/base.html.erb#L13
https://github.com/theforeman/foreman/blob/develop/app/views/layouts/base.html.erb#L18-L19

ik5:
I'ved tested it with SCSS, and it was compiled into the bundle, at this
time, after discussions, the javascript will not be changed, unless needed
to.

ohadlevy:
@ik5 but you still didnt include it in your layout? nor has any content in it, should you move one of the inline css (or all of them) as part of this pr?

ik5:
I want to add them (all of the inline css) in the next stage, this stage
was intended for the setup of the infrastructure.

sharvit:
@ik5 I personally afraid to merge infrastructure and infrastructure-consumer separately. In my experience, many times, there are changes needed to be at the infrastructure.

I think a part of me can never trust a code that never ran.

ik5:
@sharvit, @ohadlevy I'll moved all of the inline css to this PR, and it will not be the first stage, but rather the whole process.

When there will be a need, I'll add also the Javascript support and even React.

ohadlevy:
@ik5 I can't see any of your changes yet? please just let us know when a re-review is needed.

ik5:
@sharvit, @ohadlevy here are my last changes.

webpack/stylesheets/main.scss
```diff
@@ -0,0 +1,15 @@
...
+  color: #0088ce;
+}
+
+.status-color-reported {
+  color: #72767b;
+}
+
+.status-color-not-reported {
+  color: #ec7a08;
+}
```
lzap, line 15: While I understand the whole direction where Foreman core is going, I am not sold on introducing the whole JS/CSS stack into this plugin because of four extracted inline styles. It does not make much sense, discovery has always been quite UX small, four screens actually in total, we are unlikely adding dozen of new screens in this plugin. It was doing fine with the four styles and two JS functions inline.

I am not familiar with these technologies myself, but I do most of reviews, all the release engineering, packaging and troubleshooting upstream and downstream and these four lines can make my life more tough. Therefore I'd like to get volunteers from @theforeman/discovery maintainers for active involvement in packaging and related issues. @ares

.eslintrc
```diff
@@ -0,0 +1,48 @@
+{
```
ohadlevy, line 1: technically - for css you dont need eslitnrc, but it maybe ok to leave here.

app/views/discovered_mailer/_discovered_host.html.erb
```diff
@@ -1,10 +1,18 @@
...
+<td class="discovery-mailer-td-style">
   <%= link_to host, discovered_host_path(:id => host, :host => @url.host, :port => @url.port, :only_path => false, :protocol => @url.scheme) %>
 </td>
-<td style="<%= td_style%>"><%= host.try(:hardware_model_name) || 'N/A' %></td>
-<td style="<%= td_style%>"><%= host.ip %></td>
-<td style="<%= td_style%>"><%= discovery_attribute(host, :cpu_count) %></td>
-<td style="<%= td_style%>"><%= number_to_human_size(discovery_attribute(host, :memory, 0) * 1024 * 1024) %></td>
-<td style="<%= td_style%>"><%= discovery_attribute(host, :disk_count) %></td>
-<td style="<%= td_style%>"><%= number_to_human_size(discovery_attribute(host, :disks_size, 0) * 1024 * 1024) %></td>
+<td class="discovery-mailer-td-style"><%= host.try(:hardware_model_name) || 'N/A' %></td>
```
ohadlevy, line 4: can't we find a better css matcher? you can probably take any td inside a discovery table or something like that?

app/views/discovered_mailer/discovered_summary.html.erb
```diff
@@ -4,23 +4,68 @@
 <head>
   <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
   <title><%= _('Summary report for discovered hosts from Foreman') %></title>
+  <style>
```
ohadlevy: core uses a different approach using the roadie-rails gem. if you don't want to do it as part of this pr, please revert these changes?

webpack/stylesheets/main.scss
ohadlevy: @lzap I understand your concern, these changes won't look any different if we would move from inline to the traditional rails asset pipeline.

TBH: using inline style is a hack, and we should try to allow great UX experience, maintainability of the code is one of the first steps towards that. I don't think that packaging webpack assets is any different at this stage from any other plugin...

ohadlevy (changes requested):

webpack/stylesheets/main.scss
ik5: @lzap the "inline" was a pain even to you.

For example the code had stuff like `<% th_style = "color: black ..." %> and reused it in many locations.

On email based HTML it can pass, because it's problematic to write good HTML styled for emails, but using an external styling actually does this in a more web concept rather then using tempting to do the same.

I don't think that helpers or templates should have to "know" what does a color, icon etc should be, but rather what class to use.
The reason is, that it's easier to change style without touching the Ruby/Template constantly, unless there is a change to the class name.


For SCSS, it's a macro language that transfer into CSS rules. It contains variables (that just started to appear in CSS itself), it contains mixin for not repeating the same rule, it contains the ability to create nested rules, and create a "story" in that language to explain the styling for all levels, and more.
And then the language is transpiled into CSS rules, and the macros translated to actual rules, the variables are like place holders etc...

app/views/discovered_mailer/_discovered_host.html.erb
ik5: I was thinking about [bem](http://getbem.com/introduction/), but didn't want to go all the way, because nothing works in that manner here, so even the name convention is not bem, but hints for it.

I do understand why to choose id instead of class for one element, but if I understand correctly the design, each cell might contain different styling, so I didn't want to break that design.

Is my understanding wrong?

app/views/discovered_mailer/discovered_summary.html.erb
ik5: missed that, sorry, I'm fixing it

ik5: Add the change, did not test it yet, I will notify when it is ready

app/assets/stylesheets/summary.css
```diff
@@ -0,0 +1,46 @@
+body {
```
sharvit: What is the reason to put this css here and not in webpack?

ik5:
Testing the current plugin, trying to figure out few things, actually
looking to see how to include it in webpack, but load it only for the email
and not as a link, but rather then a content. So it's for test purpose only.

ik5:
two issues:

1. following discussion with @amirfefer, the `roadie-rails` plugin, does not support the webpack helper.
2. I'm missing here something, but I can't make the css file `email_summary.css` to enter the pipeline.
@sharvit Can you please take a look and maybe see what I'm missing, I worked on it for a day and a half, and I think that it's something small and stupid that I'm missing.
Thanks :)

app/views/discovered_mailer/discovered_summary.html.erb
```diff
@@ -4,23 +4,22 @@
 <head>
   <meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
   <title><%= _('Summary report for discovered hosts from Foreman') %></title>
+  <%= stylesheet_link_tag 'email_summary.css', 'data-turbolinks-track': true %>
```
sharvit: Can you try it without the `.css`?
Do you actually need turbokinks? isn't it an email template?
```erb
<%= stylesheet_link_tag 'email_summary' %>
```

ik5: Removed them both, and still it is not part of the pipeline :(

sharvit:
Try moving it to `stylesheets/foreman_discovery/email_summary.css`

ik5:
now it finds the css, unable to test it yet by hand, working on it

So please DO NOT MERGE YET

ik5:
It works :)

lzap:
You say "stage 1", can you describe all stages, what is the plan? Please include detailed packaging changes required for this to work on RPM/DEB installs.

ik5:
@lzap the discussion here actually changed from "stage1" to the only stage.
I moved all possible CSS to webpack rather then inline CSS, and that's the whole work, that is adding support for webpack, and except for email sending, not using asset pipeline, but rather webpack.

lzap:
Ok, thanks. I need to know what the packaging necessary changes are. If I merge this, will packaging break? If so, let's do package changes first.

ik5:
I'm not sure that I know, that is I know what components I'm using, but not how packaging works for them regarding the way foreman package things.

@ekohl for asset pipeline, and for webpack, what is required for packaging purpose?
The Asset pipeline is used by the action-mailer, and webpack (css at this time) for anything else.

tbrisker:
FWIW, the last time I looked into formatting in HTML mails, some clients didn't support any non-inline styling. This may have changed by now, since it was about 4 years ago, but this should be tested to make sure the formatting works well on at least some common clients (such as Outlook for example)

ekohl:
> @ekohl for asset pipeline, and for webpack, what is required for packaging purpose?

In https://github.com/theforeman/foreman-packaging/commit/2655ffbc47fcd192f61bb900c04279abdd089537 we added it to the template. You can now regenerate a spec file with [add_gem_package.sh](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/add_gem_package.sh) and if the gem contains the webpack directory it should work. The only thing we haven't looked into is additional nodejs dependencies that foreman doesn't have.

ik5:
@tbrisker the email plugin that is used actually takes the asset pipe line and placed it literally as inline css for each class element that exists, so you'll find there also `style="..."`.

@lzap based on @ekohl's  replay, there is no need at this time for anything special and it will take it all. For JS at this time we do not have anything special at all.

lzap:
@ik5 that's not how I interpret @ekohl reply. He says that they made changes to Foreman Plugin SPEC template, foreman_discovery.spec still needs to be changed and tested. And I am not gonna do this, I have no clue how node.js works. I can review and test tho.

ik5:
@lzap the thing is, is that no change to the javascript was made, only moved styling to scss and loading them, and based on @ekohl answer, because it is under `webpack` directory it knows how to use it.

Or am I missing something?

ekohl:
@lzap is correct: we do need to modify the packaging but it's well known which changes need to be made and it should be easy to do.

lzap:
Ok, I am gonna test and merge this after release of discovery for 1.18 foreman, just to be sure we have enough time for fixes. I am not confident in this area. Ping me if I forget after the release.

ik5:
@lzap, please allow me first to finish understanding how to package it (it's not just the release but also a learning curve for me), so for next requirement it will be faster, but not necessarily simpler.

lzap:
No rush, I will be branching off Discovery next week.

ekohl:
Given the huge difficulties webpack introduces in plugins I'd advise any plugin author to hold it off until it stabilizes. This also mixes 2 asset compilation stacks. Is that intentional?

ik5:
@ekohl the two assets are intentional.
The webpack stack is for the web related parts, that is things that are in display on the browser.

The asset pipline, is actually for email, we are using email library that takes the classes names and place at the email body the actual style based on that.
But it does not know how to take it from webpack, only from Rails asset pipeline.

lzap:
Let's revisit this post 1.19 then.

lzap:

rabajaj0509:
@ik5 this seems really close, would you like to solve the conflicts and rebase?

PR #424, opened 2018-04-12
Fixes #9028 - Make priority unique

boaz0:
@lzap would you mind looking at this change. In addition, I didn't write any unit-tests, because I am not sure how to run the tests.

Thanks.

### DIFF
--- a/db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
+++ b/db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
@@ -1,0 +1,24 @@
+ class AddPriorityScoreToDiscoveryRules < ActiveRecord::Migration[4.2]
+  class DiscoveryRules < ActiveRecord::Base
+  end
+
+  def up
+    # to make enough room to future entries. set initial score and delta to 100
+    delta = score = 100
+    # This loop keeps the same priority behavior before migration
+    # and makes sure there are no two entities with the same priority.
+    DiscoveryRules.reset_column_information
+    DiscoveryRules.unscoped.reorder(:priority, :created_at).find_each do |rule|
+      rule.update_attributes!(:priority => score)
+      score += delta
+    end
+    change_column :discovery_rules, :priority, :integer, null: false
+    add_index :discovery_rules, :priority, unique: true
+  end
+
+  def down
+    remove_index :discovery_rules, :priority
+    change_column :discovery_rules, :priority, :integer, default: 0
+  end
+
+ end
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -6,7 +6,7 @@ module Foreman::Controller::DiscoveredExtensions
     raise(::Foreman::Exception.new(N_("Unable to find a discovery rule, no host provided (check permissions)"))) if host.nil?
     Rails.logger.debug "Finding auto discovery rule for host #{host.name} (#{host.id})"
     # rule with *lower* priority wins (older wins for same priority)
-    DiscoveryRule.where(:enabled => true).reorder(:priority, :created_at).each do |rule|
+    DiscoveryRule.where(:enabled => true).reorder(:priority).each do |rule|
       max = rule.max_count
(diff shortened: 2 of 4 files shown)


db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
```diff
@@ -0,0 +1,23 @@
+ class AddPriorityScoreToDiscoveryRules < ActiveRecord::Migration[4.2]
+  class DiscoveryRules < ActiveRecord::Base
+  end
+
+  def up
+    score = DiscoveryRules.distinct.order("priority asc").pluck(:priority).first
```
lzap: This is the query how we search discovery rules at the moment, can you do the very same query so we know for sure there are no regressions caused by incorrect oder after migration?

    DiscoveryRule.where(:enabled => true).reorder(:priority, :created_at)

Add `unscoped` as well to make sure we get them all. Also use batch processing, so I'd expect something like:

    DiscoveryRule.unscoped.where(:enabled => true).reorder(:priority, :created_at).find_each

db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
```diff
@@ -0,0 +1,23 @@
...
+    # To keep the same priority behavior before this change,
+    # set the priority score accordingly.
+    DiscoveryRules.reset_column_information
+    rules = DiscoveryRules.order("priority asc, name asc")
+    rules.each do |rule|
+      rule.update_attributes!(:priority => score)
+      score += 1
```
lzap: Can you simplify that and set initial score to `100` and then add `100` every loop so there is enough room for additional entries? I think we can ignore current priority numbers totally, they are not important. What's important is the order we must kept.

db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
```diff
@@ -0,0 +1,23 @@
...
+      say "#{rule.name} priority is #{rule.priority}"
+    end
+    change_column :discovery_rules, :priority, :integer, unique: true, default: 0
```
lzap: The `unique` option does not seem to be right. We actually want this column to be always present with no default: `null: false` - note the default is no longer here because we want it to fail when there is no priority. Then add index to guarantee consistency and faster search queries:

    add_index(:discovery_rules, :priority, unique: true)

lzap (changes requested):
Hey, very nice thanks. I have couple of comments in the migration, but most importantly we can now change this query in `find_discovery_rule` method to take advantage of this new database layout. Instead of:

DiscoveryRule.where(:enabled => true).reorder(:priority, :created_at)

we can simply do this:

DiscoveryRule.where(:enabled => true).reorder(:priority)

db/migrate/20180412124505_add_priority_score_to_discovery_rules.rb
boaz0: Thanks @lzap for the fast response. Sorry it took me too long.

I didn't add `where(:enabled => true)` to the migration script because we want to go over all the rules and verify that both enabled and disabled rules have different priority.

If you believe that's unnecessary, let me know.

Thanks again :grin:

lzap:
We are all good, can you rebase the commit message to Fixes #XXXX where XXXX is issue number from our tracker: http://projects.theforeman.org/projects/discovery/issues/new (if you haven' created one please create new one)

boaz0:
Thanks a lot :smile:  @lzap I have removed one of the tests as it's not relevant anymore due to this change and also modified two of the tests because they had the same priority.

I tried to execute the tests on my machine using `bundle exec rake test:discovery` but it fails on all the tests. I wish I knew how to execute them properly, though.

Anyway, please forgive me if the tests fail, it will take me time to figure out how to pass. :trollface: 

Thanks. :trophy:

lzap (approved):
Tests are green locally, looks good. Nice job there! I will squash and merge once it's green on CI.

lzap:
Thanks a bunch! Nice patch.

boaz0:
@lzap it was a pleasure working on this with you.
Thank you so much! :raised_hands:

PR #425, opened 2018-04-16
Fixes #23278 - Migrate two-pane to breadcrumb-bar

sharvit:


app/views/discovery_rules/edit.html.erb
```diff
@@ -1,3 +1,3 @@
-<% title _("Edit Discovery Rule") %>
+<% title _('Edit %s' % @discovery_rule.name) %>
```
lzap: Must be `_('Edit %s') % @discovery_rule.name`

app/views/discovery_rules/_form.html.erb
```diff
@@ -16,7 +16,7 @@
           <div class="tab-pane active" id="primary">
             <%= text_f f, :name %>
             <%= autocomplete_f f, :search,
-                               :path             => "discovered_hosts",
+                               :path             => "/discovered_hosts",
```
lzap: Is this related? Can you explain?

lzap (changes requested):
Thanks, can you give a bit of context? Is this happening in core? Mind appending a screenshot? @ik5 is also making a change atm, is this compatible?

I am working hard on getting tests fixed.

app/views/discovery_rules/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <% title _("Discovery Rules") %>
 <% title_actions new_link(_("Create Rule")), help_button %>
 
-<table class="<%= table_css_classes('table-two-pane') %>">
+<table class="<%= table_css_classes('table-fixed') %>">
```
ik5, line 4: Can you please explain why/what is that change and provide some screenshots of before and after?

sharvit:
@lzap  @ik5 I am writing a small tutorial about how to migrate from `two-pane` (deprecated) to the new `BreadcrumbBar` so stay tuned :smile: 

First, remove the `table-two-pane` class to remove the `two-pane` functionality, see [two-pane.js](https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/two-pane.js)

Before (two-pane):
 
discoveryrules-two-pane

After (BreadcrumbBar):
discoveryrules-breadrumbbar

app/views/discovery_rules/edit.html.erb
sharvit: Sorry @lzap i'm not sure i understand, can you explain?

app/views/discovery_rules/_form.html.erb
sharvit: Yes, this is what lead to 404 error.

This form used to get rendered inside the `two-pane` view, which keep the index page url a the browser) and now it get rendered inside it own page with it own url `discovery_rules/new` or `discovery_rules/:id/edit`.
It makes the url wrong:
http://localhost:3000/discovery_rules/discovered_hosts/auto_complete_search?search=

sharvit: I migrated it to use `discovered_hosts_path` :+1:

app/views/discovery_rules/edit.html.erb
ik5: @sharvit, you'ved made a format string, but called the format content inside the underscore function.

What @lzap wrote, is that you should do it on the result of the underscore function, and not as part of it's original string.

ik5:
@sharvit that's cool :)

app/views/discovery_rules/edit.html.erb
sharvit: updated

lzap:
Thanks, is there a core patch associated with this work? I don't see this in core yet.

sharvit:
@lzap can you re-review please?

This might be helpful:
https://gist.github.com/sharvit/f7c975b16a289ec381673a7cc5c880e0

ohadlevy:
@lzap bump?

lzap:
Sorry had huge backlog, thanks it works.

PR #428, opened 2018-04-19
Fixes #23331 - edit_discovered_hosts is edit_hosts

lzap:
An attempt to fix the problem with random failures in core.


lzap:
This is being replaced by nicer: https://github.com/theforeman/foreman/pull/5468

PR #430, opened 2018-05-02
Fixes #3862 - audit host discovery

lzap:
I think it makes sense because if a host is provisioned manually by user, then the audit trail does not make much sense - it starts with "user X edited host Y" while this was actually discovered host provisioning.


ares:
What user the audit is created by? Is that some internal api admin?

lzap:
Sorry for the delay, discovered hosts are being created via fact upload. That sounds like an API user to me.

ares:
Thanks @lzap, for the record, it creates audits under Anonymous Admin user. Merging.

PR #431, opened 2018-05-03
Fixes #23488 - Port robottelo tests for discovery_rule

ldjebran:
Port robottelo tier1 tests for discovery_rule

Related Robottelo issue: https://github.com/SatelliteQE/robottelo/issues/5948

part of robottelo minitest port project: https://github.com/SatelliteQE/robottelo/projects/1

### DIFF
--- a/test/functional/api/v2/discovery_rules_controller_test.rb
+++ b/test/functional/api/v2/discovery_rules_controller_test.rb
@@ -31,6 +31,23 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
     assert_equal "Location 1", discovery_rule['locations'].first['name']
   end
 
+  test_attributes :pid => 'b8ae7a80-b9a8-4924-808c-482a2b4102c4'
+  test "should create discovery rule with name" do
+    min_required_attr = { :name => 'new_rule', :search => 'CPU_Count = 1', :hostgroup_id => hostgroups(:unusual).id }
+    assert_difference('DiscoveryRule.unscoped.count') do
+      post :create, params: {:discovery_rule => min_required_attr}
+    end
+    assert_response :success
+    response = JSON.parse(@response.body)
+    assert response.key?('name')
+    assert_equal min_required_attr[:name], response['name']
+    assert response.key?('hostgroup_id')
+    assert_equal min_required_attr[:hostgroup_id], response['hostgroup_id']
+    assert response.key?('search')
+    assert_equal min_required_attr[:search], response['search']
+  end
+
+  test_attributes :pid => '121e0a30-8a24-47d7-974d-998886ed1ea7'
   test "should create discovery rule with taxonomy" do
     assert_difference('DiscoveryRule.unscoped.count') do
       hostgroup = FactoryBot.create(:hostgroup, :with_os, :with_rootpass, :organizations => [organization_one], :locations => [location_one])
@@ -45,6 +62,12 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
       refute_match /error/, response.body
     end
     assert_response :success
+    response = JSON.parse(@response.body)
+    assert response.key?('id')
+    discovery_rule = DiscoveryRule.unscoped.find(response['id'])
+    refute discovery_rule.nil?
+    assert_equal organization_one.id, discovery_rule.organizations.first.id
+    assert_equal location_one.id, discovery_rule.locations.first.id
   end
 
   test "should update discovery rule" do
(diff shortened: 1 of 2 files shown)


ldjebran:
@ares Hi, can you please launch tests , seems have no rights to do it here, many thanks in advance.
Please note also that this is the only tests we have to port here.

ares:
[test]

ares:
ok to test

lzap:
Weird [test]. Maybe [test discovery] but the former worked fine.

Edit: Does not seem to work, let's poke @ekohl or @ehelms maybe CI is stuck.

komidore64:
[test]

ares:
test failures seems to be related

ldjebran:
@ares Hi, The test that failed is not related to this PR, but to a foreman model  NIC test, that seems have a bug @odovzhenko has already created a fix https://github.com/theforeman/foreman/pull/5536

komidore64:
[test foreman_discovery]

mmoll:
[test foreman_discovery]

mmoll:
:green_heart:

ldjebran:
@ares @komidore64 is there some pending comments? 
or it needs an other testing round?

ares:
job seems to be 12 days old, let's see [test foreman_discovery] [test] [test default]

test/functional/api/v2/discovery_rules_controller_test.rb
```diff
@@ -31,6 +31,23 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
     assert_equal "Location 1", discovery_rule['locations'].first['name']
   end
 
+  test_attributes :pid => 'b8ae7a80-b9a8-4924-808c-482a2b4102c4'
```
ares: I'd prefer to combine this test with the second one, they only differ in specifying org which users normally use, no org (also known as any context) is edge case only used by administrators

test/functional/api/v2/discovery_rules_controller_test.rb
```diff
@@ -53,17 +76,120 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
...
     assert_difference('DiscoveryRule.unscoped.count', -1) do
       delete :destroy, params: { :id => rule.to_param }
     end
     assert_response :success
+    assert_raise(ActiveRecord::RecordNotFound) { DiscoveryRule.unscoped.find(rule.id) }
+  end
+
+  test_attributes :pid => '415379b7-0134-40b9-adb1-2fe0adb1ac36'
+  test "should not create with too long name" do
+    assert_difference('DiscoveryRule.unscoped.count', 0) do
```
ares, line 146: this feels more like a model/unit test, as this mostly tests validations

test/functional/api/v2/discovery_rules_controller_test.rb
```diff
@@ -53,17 +76,120 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
...
+        :hostgroup_id => hostgroups(:unusual).id
+      }}
+    end
+    assert_response :unprocessable_entity
+    assert_include @response.body, 'Name is too long (maximum is 255 characters)'
+  end
+
+  test_attributes :pid => '84503d8d-86f6-49bf-ab97-eff418d3e3d0'
+  test "should not create with invalid host limit" do
+    assert_difference('DiscoveryRule.unscoped.count', 0) do
```
ares: this feels more like a model/unit test, as this mostly tests validations

test/functional/api/v2/discovery_rules_controller_test.rb
```diff
@@ -53,17 +76,120 @@ class Api::V2::DiscoveryRulesControllerTest < ActionController::TestCase
...
+        :max_count => 'invalid max_count type'
+      }}
+    end
+    assert_response :unprocessable_entity
+    assert_include @response.body, 'Max count is not a number'
+  end
+
+  test_attributes :pid => '4ec7d76a-22ba-4c3e-952c-667a6f0a5728'
+  test "should not create with invalid priority" do
+    assert_difference('DiscoveryRule.unscoped.count', 0) do
```
ares: this feels more like a model/unit test, as this mostly tests validations

ares (changes requested):
left few comments, :+1: on the direction

test/functional/api/v2/discovery_rules_controller_test.rb
omaciel: I politely disagree. We clearly have 2 distinct use cases here that should be tested individually. For me to understand, what is your concern?

ares: This is a controller test, IMHO it should test the controller responsibilities, which is: taking parameters and passing it to underlying layer (model) and generate a adequate response. I don't think there's a big value of testing two parameters separately, it means going twice through the same controller code, while the only difference is the model behavior. This should and already is tested on model level, see [here](https://github.com/ldjebran/foreman_discovery/blob/e94e8c177dfb442afafa584644db14d41e7d3d55/test/unit/discovery_rule_test.rb#L87-L93) and [here](https://github.com/ldjebran/foreman_discovery/blob/e94e8c177dfb442afafa584644db14d41e7d3d55/test/unit/discovery_rule_test.rb#L14).

Therefore, I think it's fine to write single test which passes both inputs. That means only one execution but verifying both inputs works. If one of the input does not work, from the test output, it's should be easy to tell what failed, as it checks what attributes the newly created rule actually has.

TL;DR reducing execution time, deduplicating coverage to keep tests maintainable

ldjebran: @omaciel many thanks for your question
@ares many thanks for the explanation but we tend to create tests with the minimum required fields(in a predefined scenario) and the mentioned model tests also use organization and location, the second add priority also.
As this test is not covered any where will move it to model module.

test/functional/api/v2/discovery_rules_controller_test.rb
ldjebran: We have no negative test in this controller, think we should leave this one here.

test/functional/api/v2/discovery_rules_controller_test.rb
ldjebran: moving to model module

test/functional/api/v2/discovery_rules_controller_test.rb
ldjebran: moving to model module

test/functional/api/v2/discovery_rules_controller_test.rb
ldjebran: @omaciel @ares but of course also move the assertions to the second test, so we will a combined coverage of this test divided between the second test for the returned attributes and model test for the validation  functionality

test/functional/api/v2/discovery_rules_controller_test.rb
ares: :+1:

test/functional/api/v2/discovery_rules_controller_test.rb
ares: sounds good, thanks

ldjebran:
@ares many thanks, the changes here depend from this PR in Foerman project  https://github.com/theforeman/foreman/pull/5586

ares:
let's retest with the change in core now [test discovery]

ares:
[test foreman_discovery]

test/functional/api/v2/discovery_rules_controller_test.rb
omaciel: -- Therefore, I think it's fine to write single test which passes both inputs. That means only one execution but verifying both inputs works. If one of the input does not work, from the test output, it's should be easy to tell what failed, as it checks what attributes the newly created rule actually has.

@ares I see that you have reached an agreement with @ldjebran but here is why your proposal is not considered good practice for testing: assuming `parameter1` makes the test fail, `parameter2` would not be tested (unless this framework would assure that all parameters are executed?) and therefore you'd only see half the story. If the second test should fail, it would be best to see both failures at the same time.

ares: For me, the most important is that the test is red, I need to start investigating since my code can't get in, when I fix the original issue, I run tests locally and see there's another failure. I understand you might be interested in "how many different tests are failing", but I'd argue you'd need to have each assertion as a separate test. From my experience, that does not work well and generate a lot of failed tests for the same cause.

omaciel: @ares I agree that one assertion per test is a good rule of thumb.

ares (approved):

ares:
:+1:  good from my point of view, @lzap any concerns or can we merge?

test/functional/api/v2/discovery_rules_controller_test.rb
ares: So here we disagree :-) Seems like robottelo tests contain also multiple assertions. I think that would be worth of discussion on our forum, as just 2 of us can't decide. I'm happy to continue conversation there if you don't mind.

omaciel: Yup, robottelo will sometimes have multiple assertions, specially if we're talking about `end-to-end` or some of the higher level test types, but yes, good discussion for the the forum next week :)

ekohl: To jump in here: the model provides a method that validates. That abstraction is called in the controller. Therefor it's sufficient to test every single case fails the validation function. Ideally this is done in separate tests so you can easily see which input actually failed. Then in the controller it's sufficient to test that the validation method is called. Testing every single input at the controller level is testing at the wrong layer of abstraction. That's why it's sufficient to have 2 test cases for the controller: one with valid input and one with invalid input. What exactly triggers the invalidation is not that important.

At least, that's my view on it.

omaciel: @ekohl  No argument that testing at the controller level should suffice. I believe that @ldjebran however found an interesting case that required the extra step...

ldjebran:
@ares @lzap Hi, is there any additional comments?

ares:
spoke with @lzap offline, no concerns from his side, so merging, thanks @ldjebran

ldjebran:
@ares @lzap many thanks.

PR #432, opened 2018-05-10
Fixes #23554 - Handle both cases of options[:user] as arrive from foreman

ik5:


ik5:
[test]

ik5:
[test foreman_discovery]

ik5:
@lzap, @rahulbajaj0509 can you please take a look at this PR and review it?
Thank you :)

rabajaj0509:
@ik5 thanks for the pr :) 
the code LGTM, would it be feasible to add some test cases here though?

ik5:
@rahulbajaj0509 I'll try to figure something some small tests for it. 
Thanks :)

ik5:
@rahulbajaj0509 added three tests. 
They seems to work as expected (also broke them locally to make sure on that).

Thank you very much!

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,14 @@
 class DiscoveredMailer < ::ApplicationMailer
   helper :discovered_hosts
   def discovered_summary(options = {})
-    raise ::Foreman::Exception.new(N_("Must specify a user with email enabled")) unless (user = User.find(options[:user])) && user.mail_enabled?
+    user = if options[:user].kind_of? User
+             options[:user]
+           elsif options[:user].kind_of? Integer
+             User.find(options[:user])
+           else
+             raise ::Foreman::Exception.new(_('Invalid provided user type of %s ') % options[:user].class.to_s)
```
lzap: I'd probably include the fact the exception was raised during mail generation: "Unable to generate discovery summary: invalid..."

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,14 @@
...
+           end
+    raise ::Foreman::Exception.new(N_("Must specify a user with email enabled")) unless user.mail_enabled?
```
lzap: In this case, I'd rather log a warning to logs, this will break mailing rake process executed from job rendering mailer unable to continue with the task if the exception is not catched. Logging is safer.

lzap (changes requested):
Ok, was there an exception for the current behavior? Just curious how did you find this.

app/mailers/discovered_mailer.rb
ik5: But the email is for the facts, and it's before you are looking for them, so it does not make sense to write a list of them imo.

app/mailers/discovered_mailer.rb
ik5: done

ik5:
@lzap no exception, just when looked how to test my code (before writing tests), i found two ways in foreman that emails are called, but only one way that the discovery looks for user email, so I wanted to "fix" a possible bug that might never happen but might do exists.

ik5:
[test foreman_discovery]

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,18 @@
 class DiscoveredMailer < ::ApplicationMailer
   helper :discovered_hosts
   def discovered_summary(options = {})
-    raise ::Foreman::Exception.new(N_("Must specify a user with email enabled")) unless (user = User.find(options[:user])) && user.mail_enabled?
+    user = if options[:user].kind_of? User
+             options[:user]
+           elsif options[:user].kind_of? Integer
+             User.find(options[:user])
+           else
+             raise ::Foreman::Exception.new(_('Invalid provided user type of %s') % options[:user].class.to_s)
```
lzap: Incorrect use of exception, please read: https://projects.theforeman.org/projects/foreman/wiki/Translating#Exceptions

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,18 @@
...
+           end
+
+    unless user.mail_enabled?
+      log.warn(N_("Must specify a user with email enabled"))
+      return
+    end
```
lzap: I'd rather log this as `debug` because there can be many users in the database possibly all disabled and we run these summaries every day/hour and this can possibly flood logs. Also include user id in the message (not username that's security leak).

lzap (changes requested):
Frankly, the summary API is poorly designed and I'd rather see this improved in core. Key issues:

* accept only one form of passing user
* break up hash into named parameters
* define return values (how many emails was sent, list of errors etc)

I know that the patch had no ambitions to do that, but I don't think it's too difficult. The API change can be backward compatible

app/mailers/discovered_mailer.rb
ik5: Oh, I was wondering why some uses the `N_` method on some places but not always, thanks!

app/mailers/discovered_mailer.rb
ik5: done

ik5:
The problems with changing foreman core way of doing it, are:
1. People can still use the old way (even on deprecation notice)
2. It might break code we can't think of, even when there an attempt for backwards compatibility, because we still touch the logic a bit

So I think that in order to change it in core, we should address it also with a new API version, so it will be clear that it might break stuff.
WDYT?

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,18 @@
...
+    user = if options[:user].kind_of? User
+             options[:user]
+           elsif options[:user].kind_of? Integer
+             User.find(options[:user])
+           else
+             raise ::Foreman::Exception.new(N_('Invalid user type of %s was provided') % options[:user].class.to_s)
+           end
+
+    unless user.mail_enabled?
+      log.debug(N_("The user %d does not enable email") % user.id)
```
lzap: Wouldn't this raise MethodNotFound: log error? Can you add a test for this code branch?

Also do not translate logs, we rarely do that.

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,18 @@
 class DiscoveredMailer < ::ApplicationMailer
   helper :discovered_hosts
   def discovered_summary(options = {})
-    raise ::Foreman::Exception.new(N_("Must specify a user with email enabled")) unless (user = User.find(options[:user])) && user.mail_enabled?
+    user = if options[:user].kind_of? User
+             options[:user]
+           elsif options[:user].kind_of? Integer
+             User.find(options[:user])
+           else
+             raise ::Foreman::Exception.new(N_('Invalid user type of %s was provided') % options[:user].class.to_s)
```
lzap: I still see `%` instead `,` this is not correct. Exception constructor requires all arguments to be passed via arguments.

lzap (changes requested):
> So I think that in order to change it in core, we should address it also with a new API version, so it will be clear that it might break stuff.

If I understand this correctly, core mailer API can send either user ID or User object. Therefore all mailers must have the same code you propose here.

What I propose is let's change it in core so the API either sends one or another but not both. I assume that all mailers must have this branch to work anyway, so it won't break.

Anyway, I am fine merging this into discovery, but please consider fixing this in core.

There are related test failures tho.

app/mailers/discovered_mailer.rb
ik5: it should be `Rails.logger` you are correct :(

app/mailers/discovered_mailer.rb
ik5: the format string is for the translation, not for the exception. The exception should get just the link.
Am I missing something in your meaning?

lzap: From https://projects.theforeman.org/projects/foreman/wiki/Translating#Exceptions

We have defined two exceptions `Foreman::Exception` and `Foreman::WrappedException`. Both are generic exceptions which are i18n-friendly. Use the following style to properly format messages.

Let me show this to you:

```
[4] pry(main)> ::Foreman::Exception.new(N_('Invalid user type of %s was provided') % 'DummyType').code
=> "ERF42-7023"
[5] pry(main)> ::Foreman::Exception.new(N_('Invalid user type of %s was provided') % 'AnotherType').code
=> "ERF42-5230"
[6] pry(main)> ::Foreman::Exception.new(N_('Invalid user type of %s was provided'), 'DummyType').code
=> "ERF42-9840"
[7] pry(main)> ::Foreman::Exception.new(N_('Invalid user type of %s was provided'), 'AnotherType').code
=> "ERF42-9840"
```

ik5: oh, now i get it, didn't understand it without your explanation :/ 
thanks

lzap: Sorry my bad, it's just one character, very hidden indeed :)

app/mailers/discovered_mailer.rb
```diff
@@ -1,7 +1,18 @@
...
+    user = if options[:user].kind_of? User
+             options[:user]
+           elsif options[:user].kind_of? Integer
+             User.find(options[:user])
+           else
+             raise ::Foreman::Exception.new(N_('Invalid user type of %s was provided'), options[:user].class.to_s)
+           end
+
+    unless user.mail_enabled?
+      Rails.logger.debug("The user %d does not enable email" % user.id)
```
lzap: Nicpick: "The user %d does not have email enabled, skipping" perhaps?

lzap:
Other than that I am good, @rahulbajaj0509 any objections? Thanks

rabajaj0509:
@lzap LGTM.

app/mailers/discovered_mailer.rb
ik5: fixed, thanks :)

lzap (approved):

lzap:
Thank you very much for your contribution, this is now merged.

PR #436, opened 2018-06-01
Fixes #23224 - keep_subnet flag to prevent subnet override

lzap:
Setting "Update subnets from facts" makes discovery impossible if taxonomy does not match the host leading to "Subnet is not defined for host's organization". Discovery uses standard core Puppet parser to recognize all interfaces, therefore discovery users must have all subnet taxonomy aligned with discovered host taxonomy, even if they don't plan to use all interfaces.

This patch adds flag keep_subnet which is added by discovery, when present subnet is never updated. This prevents the misbehavior when the flag is set.

This patch needs core change:

https://github.com/theforeman/foreman/pull/5635

### DIFF
--- a/test/unit/host_discovered_test.rb
+++ b/test/unit/host_discovered_test.rb
@@ -32,6 +32,14 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert_equal subnet, host.primary_interface.subnet
   end
 
+  test "should setup subnet when update_subnets_from_facts is true" do
+    Setting[:update_subnets_from_facts] = true
+    subnet = FactoryBot.create(:subnet_ipv4, :name => 'Subnet99', :network => '10.35.27.0', :organizations => [organization_one], :locations => [location_one])
+    Subnet.expects(:subnet_for).with('10.35.27.3').returns(subnet)
+    host = discover_host_from_facts(@facts)
+    assert_equal subnet, host.primary_interface.subnet
+  end
+
   test "should setup subnet with org and loc set via settings" do
     org = FactoryBot.create(:organization, :name => "subnet_org")
     loc = FactoryBot.create(:location, :name => "subnet_loc")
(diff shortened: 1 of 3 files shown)


test/unit/host_discovered_test.rb
```diff
@@ -32,6 +32,16 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert_equal subnet, host.primary_interface.subnet
   end
 
+  test "should setup subnet when update_subnets_from_facts is true" do
+    Setting[:update_subnets_from_facts] = true
+    subnet = FactoryBot.create(:subnet_ipv4, :name => 'Subnet99', :network => '10.35.27.0', :organizations => [organization_one], :locations => [location_one])
+    Subnet.expects(:subnet_for).with('10.35.27.3').returns(subnet)
+    host = discover_host_from_facts(@facts)
+    assert_equal subnet, host.primary_interface.subnet
+  ensure
```
ik5: It fails on this ensure, because the block needs to have only `end`, and not that.
I recommend to place it under `begin` block

ik5 (changes requested):
the tests are failing

test/unit/host_discovered_test.rb
lzap: Haha DEJA VU! :-)

ares:
should this bump [foreman dependency](https://github.com/theforeman/foreman_discovery/blob/develop/lib/foreman_discovery/engine.rb#L46) and hence major version? it will silently fail on Foreman <= 1.18

lzap:
Good point, do not merge until 1.18 is branched (worst case: end of next week).

lzap:
Tested v12, tagged, rebased on top of this with the engine change. There are failures tho.

dLobatog (approved):
Thanks @lzap !

PR #437, opened 2018-06-01
Fixes #16330 - random and MAC name generators

lzap:
With deacon gem as foreman core dependency, it is possible to implement naming discovered hosts via deacon's MAC address generator.

https://github.com/lzap/deacon

This makes a lot of sense because the name stays with server until decommission. It's a feature of deacon gem - same MAC maps to same name. See the README.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -37,6 +37,12 @@ class Host::Discovered < ::Host::Base
     where(taxonomy_conditions).order("hosts.created_at DESC")
   }
 
+  NAMING_PATTERNS = {
+    'Fact' => _('Fact + prefix'),
+    'Random-name' => _('Random name'),
+    'MAC-name' => _('MAC-based name')
+  }.freeze
+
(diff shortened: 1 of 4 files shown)


lzap:
This is fun! It needs a small core PR: https://github.com/theforeman/foreman/pull/5637

ik5 (changes requested):
the tests breaks on me, the reason is that the `ensure` part is not part of a `begin` and not part of a method, but rather inside a block call.

It should be something like that:
```
...
  begin
   ...
  ensure
     Setting[:discovery_naming] = "Fact"
  end
end
```

lzap:
Yeah, weird it looks like newer version of Ruby supports it, good catch Jenkins! Rebasing, I actually removed those, it's not necessary since Settings are fixture records. They are reset on every test.

dLobatog (approved):
Thanks, works pretty well @lzap . Can we get the discovery docs updated to include this?

PR #439, opened 2018-06-07
Fixes #23851: first auto-bond implementation

Thomas-Gelf:
refs #23851


app/models/setting/discovered.rb
```diff
@@ -66,6 +67,10 @@ def self.discovery_hostname_fact_array
     from_array Setting['discovery_hostname']
   end
 
+  def self.discovery_auto_bond?
+    Foreman::Cast.to_bool(Setting['discovery_auto_bond'])
```
timogoebel: Is this really needed? I believe the value should already be a boolean.
In addition, this is not used anywhere.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -1,9 +1,12 @@
+require 'foreman_discovery/lldp_neighbors'
```
timogoebel: This should not be necessary. Rails does auto loading pretty well. Let's remove this.
You might want to move the library to app/services as well (instead of lib/foreman_discovery/).

lib/foreman_discovery/lldp_neighbors.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+        interfaces[iface] = {} unless interfaces.has_key? iface
+        interfaces[iface][property] = facts[key]
+      end
+
+      interfaces.each do |name, properties|
+        neighbors.set name, properties
+      end
+
+      return neighbors
```
timogoebel: Nit: superfluous return

app/services/foreman_discovery/host_converter.rb
```diff
@@ -19,6 +22,41 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
...
+        :attached_devices => neighbors,
+        :primary          => true,
+        :provision        => true,
+        :name             => name,
+        :ip               => ip,
+        :host             => host
+    )
+
+    bond.save!
+    host.interfaces.push bond
```
timogoebel: Nit: This needs brackets...

lib/foreman_discovery/lldp_neighbors.rb
```diff
@@ -0,0 +1,65 @@
...
+      interface = get(ifname)
+      return nil unless interface.has_key? 'PVID'
+
+      list = []
+      interfaces.each do |name, neighbor|
+        next unless neighbor.has_key? 'PVID'
+        list.push name if neighbor['PVID'] == interface['PVID']
+      end
+
+      list.size > 1 ? list : nil
```
timogoebel: Nit:
`list.empty? ? nil : list`

or

```ruby
return if list.empty?
list
```

timogoebel (changes requested):
@Thomas-Gelf: Thanks, looks good. Do you mind adding a couple of tests for this?

app/models/setting/discovered.rb
Thomas-Gelf: I don't think so, just added it as it's being done the same way the line below :laughing:

app/services/foreman_discovery/host_converter.rb
Thomas-Gelf: It drove me mad to figure out what to place where, as I faced quite some inconsistencies. Well, at least I haven't been able to really grasp the rules. No wonder Ruby projects fire an such an immense amount of stat-calls.

I placed this code in a separate class to make it easier to test in an isolated way. Just, I failed to run the tests, gave some rake test:foreman_discovery examples triggered from the foreman base directory a try, but nothing happened.

lib/foreman_discovery/lldp_neighbors.rb
Thomas-Gelf: Fixed.

app/services/foreman_discovery/host_converter.rb
Thomas-Gelf: The call to push? Do you have a brackets-everywhere policy?

Thomas-Gelf: (because then there would be more)

lib/foreman_discovery/lldp_neighbors.rb
Thomas-Gelf: Not sure... I had the second one first, but I hate it to read code where it's not immediately obvious that nil/null must be expected. And what's ugly about this is that it's not `list.empty?` - only results with more than one entry are returned, if there is only one `peer` it should return `nil`.

I changed it to read:
```ruby
      return if list.size < 2
      list
```

This way it's more obvious that at least two are required for a valid candidate.

Cheers,
Thomas

app/services/foreman_discovery/host_converter.rb
Thomas-Gelf: @timogoebel: any helpful hint on how to kick the test run would be greatly appreciated!

Thomas-Gelf:
LldpNeighbors is covered, HostConverter will follow

Thomas-Gelf:
NB: I tried to address all of the above, just the question related to brackets still stands

Thomas-Gelf:
```
# Running:

LldpNeighborsTest#test_0001_#get_neighbors_by_interface gives nothing with no LLDP facts = 0.77 s = .
LldpNeighborsTest#test_0003_#get_neighbors_by_interface gives a valid pair for LLDP facts with matching PVID = 0.00 s = .
LldpNeighborsTest#test_0006_#list_by_pvid pairs interfaces by PVID = 0.00 s = .
LldpNeighborsTest#test_0007_#list_by_pvid supports multiple PVIDs = 0.00 s = .
LldpNeighborsTest#test_0002_#get_neighbors_by_interface gives nothing with unmatching PVID in LLDP facts = 0.00 s = .
LldpNeighborsTest#test_0005_#list_by_pvid gives an empty Hash without LLDP facts = 0.00 s = .
LldpNeighborsTest#test_0004_#list_by_pvid gives an empty Hash without LLDP facts = 0.00 s = .

Finished in 0.783887s, 8.9299 runs/s, 12.7569 assertions/s.
7 runs, 10 assertions, 0 failures, 0 errors, 0 skips
```

timogoebel:
> just the question related to brackets still stands

We don't have a policy for that, but I'd prefer brackets everywhere except when the method is called without any arguments.

test/unit/lldp_neighbors_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  test "#get_neighbors_by_interface gives a valid pair for LLDP facts with matching PVID" do
+    assert_equal %w(eth0 eth1), bond_facts.get_neighbors_by_interface('eth0')
+    assert_equal %w(eth0 eth1), bond_facts.get_neighbors_by_interface('eth1')
+    assert_nil bond_facts.get_neighbors_by_interface('br182')
+  end
+
+  test "#list_by_pvid gives an empty Hash without LLDP facts" do
+    assert_equal Hash({}), simple_facts.list_by_pvid
```
timogoebel: this is the only case were i'd call `assert_equal` with brackets. Please omit the `Hash()` everywhere, it's not good ruby style.

test/unit/lldp_neighbors_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  test "#list_by_pvid supports multiple PVIDs" do
+    assert_equal Hash('182' => %w(eth2), '184' => %w(eth1)), lldp_facts.list_by_pvid
+  end
+
+  def simple_facts
+    neighbors 'facts'
+  end
+
+  def lldp_facts
+    neighbors'facts_with_lldp'
```
timogoebel: nit: missing whitespace

test/unit/lldp_neighbors_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  test "#list_by_pvid pairs interfaces by PVID" do
+    assert_equal Hash('182' => %w(eth0 eth1)), bond_facts.list_by_pvid
+  end
+
+  test "#list_by_pvid supports multiple PVIDs" do
+    assert_equal Hash('182' => %w(eth2), '184' => %w(eth1)), lldp_facts.list_by_pvid
+  end
+
```
timogoebel: the methods below can be private, a better (or modern) way to write this would be using `let`. But this is fine as well.

```ruby
context 'with facts' do
 let(:facts) { some code } # acts as a lazy accessor
  test 'abc' do
    assert facts
  end
end
```

Thomas-Gelf:
Also extended host_discovery_test:

```
HostDiscoveredTest#test_0033_provision_interface isn't touched with no LLDP facts = 0.87 s = .
HostDiscoveredTest#test_0034_provision_interface isn't touched with no peer on the same VLAN = 0.77 s = .
HostDiscoveredTest#test_0036_provision_interface is not switched to bond0 if disabled = 0.77 s = .
HostDiscoveredTest#test_0035_provision_interface is switched to bond0 with more than one interface on the same VLAN = 1.00 s = .
```

Thomas-Gelf:
There is one more:

    HostDiscoveredTest#test_0037_former provision_interface is not managed after switching to bond0 = 0.95 s = .

Thomas-Gelf:
@timogoebel: tests look good to me, please let me know in case you need more

timogoebel:
ok to test

timogoebel (approved):
LGTM.

Thomas-Gelf:
Don't blame `ReferenceError: Can't find variable: jQuery` (`Capybara::Poltergeist::JavascriptError`) on me ;-)

Thomas-Gelf:
@timogoebel: tests are failing because of a deprecation exception introduced in Foreman, I pushed a related pull request in #440

Thomas-Gelf:
Hint: I rebased this on fix/replace-notice-with-success (#440) to get the tests green

rabajaj0509:
@Thomas-Gelf  can you remove the `success` as it got merged in #440 and rebase this PR?

Thomas-Gelf:
@rahulbajaj0509: sure :-)

app/services/foreman_discovery/host_converter.rb
```diff
@@ -19,6 +20,41 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
...
+        :identifier       => "bond0",
+        :attached_devices => neighbors,
+        :primary          => true,
+        :provision        => true,
+        :name             => name,
+        :ip               => ip,
+        :host             => host
+    )
+
+    bond.save!
```
timogoebel: I'm wondering if it's a good idea to save the host here.
`ForemanDiscovery::HostConverter.to_managed` just converts the host object but does not save it so a user can still make changes to the host in the UI or even abort the operation. Same for the call to `update` above.

Maybe this should just update the models but not persist the changes to the database?
What do you think?

Thomas-Gelf: I would prefer not to do so, but haven't been able to figure out a clean way. Problem seemed to be that at this time the primary_interface has already been stored. Removing the primary flag and creating another primary interface always led to a conflict. It behaved like a check that verified the database first without taking the in-memory changes into account.

Thomas-Gelf: You could give it a try, but as far as I remember when starting with an empty database you'll see a single (managed, primary) interface in the DB as soon as a host with (multiple) interfaces got "discovered".

lzap: Hello, thanks for your contribution! Very nicely done.

Now, what @timogoebel mentioned is a real issue - this method was designed specifically in a way that it does not store anything in database, it is called when Edit Host form is being rendered to convert host to Host::Managed so the form from Foreman Core can be reused. We wanted to avoid creating New Host form, it's the most complicated one in the whole project.

Thomas-Gelf: I'd love if it could work that way, that's what I initially tried to do. However as far as I remember even without this patch it didn't behave that way. The primary interface showed up in the database immediately after discovery, and my attempt to apply an in-memory modification (remove primary flag from the primary interface, add the bond interface to the hosts `interfaces`) had always thrown some kind of a duplicate primary interface error.

In case you think that's not what the current behavior should be I can try to reproduce it tomorrow with an empty database and the latest development branch.

timogoebel: You could try to actually update the host object with nested attributes. Just a thought.

lzap:
I don't think we need to necessary do this NIC "transformation" during discovery, we can leave discovered host "as is" and do the transformation during provisioning. The LDAP and other facts will not disappear, they will be carried over unless "discovery_clean_facts" setting is set. You can refactor your patch to do the action via `host_extensions.rb` (see `after_validation :queue_reboot` which does the reboot). You need to perform the transformation orchestration before DHCP/TFTP steps are done (we do have an "ordering" in our orchestration framework). That should do it and this way you can avoid doing the save in the `to_managed` which is I think a blocker (will change host when it enters Edit Host form).

If you want to go further, one of our top five upvoted features across *all* projects is *Auto-provisioning custom scripts for NIC configurations*: https://projects.theforeman.org/issues/13847 with simple idea - the moment a discovered host is being (auto) provisioned a custom ERB would be executed allowing to do anything that's needed with NICs. Convert them into bonds, vlans or delete unnecessary ones - anything that is possible with ActiveRecord in safemode (more methods must be added). Instead of having a global setting for bond creation, you would have an ERB template associated with discovery rule that would do the changes. This can easily be beginning of something big that lots of our users want. Initially the idea is to link the ERB with discovery rule but this could also work with regular provisioning, we'd need to have a template association in core tho.

timogoebel:
> I don't think we need to necessary do this NIC "transformation" during discovery, we can leave discovered host "as is" and do the transformation during provisioning.

I think we do as I believe showing the current interfaces in the host form is a critical feature. Showing the user what is going to happen and allow further changes is the only way this makes sense to me. Changing some models in memory and show the modified values in the form should be doable imho. Let's give this a shot before we change this too much because of technical depts.

Thomas-Gelf:
@timogoebel: also in my believes `HostConverter` makes a good fit for this if you want this to work with manual and auto provisiong - while still allowing the user to tweak the bond when opting for manual provisioning.

Problem is that I found no way to change the primary interface in-memory together with a pushed (unstored) interface. I'm missing knowledge about all the ActiveRecord-based models and the magic that goes on behind the scenes. With some guidance I should be able to fix this. I tried hard with `Nic::Bond.new`, with `.instantiate`, with `.attributes=`, by setting every single attribute - I haven't been able to get my interface accepted.

I tried with `interfaces.push` and I also tried to serialize all interfaces to an Array, setting `host.interfaces = interfaces`. Either way, it fails. My best attempt breaks at `became.interfaces = self.interfaces` in `host/base.rb#becomes` with `Failed to replace interfaces because one or more of the new records could not be saved.`

To me it looks like it is somehow not possible to set a `has_many` relation with unstored objects. If that's possible please teach me how.

lzap:
Timo, let me explain my thinking. I think from the day one that discovery should not build on top of Foreman host model like we currently do using clunky ActiveRecord "inheritance". I believe the correct design is keeping its own inventory (a db table) of discovered hosts in plain form - just an unnamed list of facts. No parsing, no NIC detection or anything like that.

In such a design, new managed or unmanaged hosts should be created **during** provisioning. This includes NIC parsing or transformation, as I named it. This has nothing to do with ability to provide "preview" of parsed NICs - we could still deliver this feature by parsing NICs on demand, our API would need to be enriched for this.

I would like us going towards this goal rather than making tighter integration discovery and core (host and nics) models. This patch does the opposite, but that should not block us from merging it. We are not here to redesign discovery of course. During transformation during provisioning was more aligned with my long-term vision. Back to review.

@Thomas-Gelf method `to_managed` is not the right place to do the transformation. I think you need to do it earlier, during NIC parsing. As you can see, what we do is leveraging core fact parser to build NICs, we override one or two methods to correctly detect primary interface. You could try to hook into the core parser similar way, but rather than monkey patching or other overriding (we already have too many) I think the best approach would be to move this patch into core. This will give Foreman core fact parser ability to recognize bonds from newly registered regular or discovered hosts. @ares opinion?

lzap (changes requested):
Looks like @ares won't be available for a while.

I am giving this a no-go for now as this version calls save in `to_managed` method. Please prevent doing that, I've thrown some ideas already.

timogoebel:
@lzap: Thanks, I have talked with @Thomas-Gelf about this. We agreed, that I'll take a look at this when I find the time. Probably later this week. Thanks for your suggestions to move forward on this.

timogoebel:
I checked this out. Making the changes in memory is actually pretty easy.

My naive approach looks like this:

```diff
 diff --git a/app/services/foreman_discovery/host_converter.rb b/app/services/foreman_discovery/host_converter.rb
index bcb1c05..25fa91e 100644
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -33,26 +33,19 @@ class ForemanDiscovery::HostConverter

     ip = primary.ip
     name = primary.name
-    primary.update(
-       :primary   => false,
-       :provision => false,
-       :managed   => false,
-       :name      => nil,
-       :ip        => nil
-    )
+    primary.name = nil
+    primary.ip = nil
+    primary.provision = false
+    primary.primary = false

-    bond = Nic::Bond.create(
-        :identifier       => "bond0",
-        :attached_devices => neighbors,
-        :primary          => true,
-        :provision        => true,
-        :name             => name,
-        :ip               => ip,
-        :host             => host
+    bond = host.interfaces.build(
+      type: 'Nic::Bond',
+      identifier: 'bond0',
+      primary: true,
+      provision: true,
+      name: name,
+      ip: ip
     )
-
-    bond.save!
-    host.interfaces.push(bond)
   end

   def self.set_build_clean_facts(host)
```

This issue now is, that the primary interface cannot be found as rails thinks the relation is `nil`, see https://github.com/theforeman/foreman/blob/fcc2b55d07cbb3708f2fab212bf57a3366561d06/app/models/host/base.rb#L22
I have no idea yet, how we can workaround this.

timogoebel:
@Thomas-Gelf: I like adding the idea of adding this to the fact parser so this code runs when the host is saved to the database. But I don't think it makes a lot of sense in core as the fact parser in core should just provide the data from a system so it can be displayed in Foreman's UI. In discovery we need this before provisioning happens. The fact parser prepares the host for proper provisioning. 
We already have [our own fact parser](https://github.com/theforeman/foreman_discovery/blob/bc2380bd6f3c42a18f9d69f39d668579c64e2c3e/app/services/foreman_discovery/fact_parser.rb) where we should be able to override the `interfaces` method to get this working.

@lzap, what do you think?

lzap:
Overriding our parser is the best idea IMHO so far. :)

Thomas-Gelf:
Update: I gave the `fact_parser` a try, shifted the whole logic - but failed. Have been unable to enforce a `Nic::Bond` and to pass `attached_devices`. Eventually I did it wrong, but anyways: I moved the call to `eventually_make_bond()` to `Host::Discovered.populate_discovery_fields_from_facts`. This way the transformation takes place immediately once the interfaces are being stored for the first time.

Thomas-Gelf:
Hint: my attempt with the `fact_parser` was solely in `foreman_discover`, I didn't tweak the base class in `foreman`. So, by enhancing the possibilities there this should also be possible. However, the new approach looks good to me. @lzap: thoughts?

Thomas-Gelf:
Rebased and fixed mac address setting

app/models/host/discovered.rb
```diff
@@ -143,6 +143,7 @@ def populate_discovery_fields_from_facts(facts)
     self.discovery_attribute_set = DiscoveryAttributeSet.where(:host_id => id).first_or_create
     self.discovery_attribute_set.update_attributes(import_from_facts)
     # set additional discovery attributes
+    ::ForemanDiscovery::HostConverter.eventually_make_bond(self) if Setting[:discovery_auto_bond]
```
lzap: This place is better but the code does not belong into HostConverter.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -19,6 +19,43 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
     host
   end
 
+  def self.eventually_make_bond(host)
```
lzap: Can you move this method out of HostConverter into LldpNeighbors or somewhere else? I suggest creating a "post processing framework":

* ForemanDiscovery::ImportHookService - a simple registry of hooks
* ForemanDiscovery::ImportHook - base class for processor - keep it simple `after_populate(host, facts)` as the only method for now
* ForemanDiscovery::ImportHooks::LldpNeighbor - implementation - move this method there

Optionally break method `populate_discovery_fields_from_facts` into three other separate hooks: DiscoveryAttribute, SubnetAndTaxonomy and LockTemplates. I can do this later if you don't want to touch this.

The reason I want this is in the future we might break the processing of NICs and Facts in Core in similar way, so we could easily merge this. The idea is not to save anything into database during the whole import phase so plugins can hook into the process easily and at the very end of the whole processing, save all NICs and fields to database in a single place.

lzap (changes requested):
Ok looks better but I want you to move the eventually_make_bond method somewhere else please.

Thomas-Gelf:
@lzap: I moved the code as requested and adjusted tests accordingly. Regarding the "post processing framework": could you please point me to an example Hook in the Foreman code, just to get an idea of how that should look like? In case it is not tricky I could give it a try.

timogoebel:
I'd suggest to implement the "post processing framework" in a separate PR. It's not related to this issue. I agree it's a nice refactoring, but we don't need it now. @lzap, agreed?

lzap (approved):
Ok, although I think better direction would be creating post-processing framework. Please document the changes in https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/foreman_discovery/nightly/index.md and we are good to merge.

timogoebel:
> Ok, although I think better direction would be creating post-processing framework.

I agree, luckily we can still do that later.

> Please document the changes

Here you go: https://github.com/theforeman/theforeman.org/pull/1129

lzap:
Sure, I'd rather go that direction sooner because discovery plugin is basically a collection of hacks :-)

timogoebel:
Thanks, @Thomas-Gelf.

PR #441, opened 2018-06-25
API does not always set build flag

lzap:
And fixed test.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -80,7 +80,7 @@ module Api
       end
 
       def update
-        @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host, true, true, managed_host_params)
+        @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host, true, managed_host_params[:build], managed_host_params)
         forward_request_url
         update_response = @host.save
         process_response update_response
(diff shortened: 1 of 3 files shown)


test/functional/discovered_hosts_controller_test.rb
```diff
@@ -294,6 +294,9 @@ def test_no_dns_rebuild_if_dns_pending
     hostgroup = prepare_hostgroup_for_dns_rebuild(host)
     Nic::Managed.any_instance.expects(:rebuild_dns).never
     Host::Managed.any_instance.stubs(:skip_orchestration?).returns(false)
+    Host::Managed.any_instance.stubs(:queue_puppetca_create).returns(true)
+    Host::Managed.any_instance.stubs(:queue_puppetca_certname_reset).returns(true)
+    Host::Managed.any_instance.stubs(:queue_puppetca_autosign_destroy).returns(true)
```
lzap, line 299: There are two commits, the first one just reflect change in core: https://github.com/theforeman/foreman/pull/5580

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -80,7 +80,7 @@ def create
       end
 
       def update
-        @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host, true, true, managed_host_params)
+        @host = ::ForemanDiscovery::HostConverter.to_managed(@discovered_host, true, managed_host_params[:build], managed_host_params)
```
lzap, line 83: Introduced in 044cd42ffcb96819af0bd3d88bb348f2f9dc8ea9 and regression - users were unable to override build flag to false, it was always true.

lzap:
Quick merge to unblock unit tests and blocker issue for v12.0.1 @timogoebel @tbrisker or @rahulbajaj0509 please?

timogoebel (approved):
LGTM, though I haven't tested this.

timogoebel:
Thanks, @lzap.

PR #445, opened 2018-07-12
Fixes #24235 - autosuggest rule priority

ares:
@rahulbajaj0509 mind to take a look? @rplevka mind to test if that's what you want?

priority

PR #446, opened 2018-07-12
Fixes #24235 - autosuggest rule priority

ares:
reposted https://github.com/theforeman/foreman_discovery/pull/445 github failed to update the branch

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -1,4 +1,5 @@
 class DiscoveryRule < ApplicationRecord
+  STEP = 100
   audited :associated_with => :hostgroup
   include Authorizable
   extend FriendlyId
@@ -38,6 +39,10 @@ class DiscoveryRule < ApplicationRecord
   def default_int_attributes
    self.max_count ||= 0
    self.priority  ||= 0
+  end
+
+  def self.suggest_priority
+    self.unscoped.maximum(:priority).to_i + STEP
   end
 
   def enforce_taxonomy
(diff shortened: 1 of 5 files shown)


lzap (approved):
LGTM, let's wait for tests.

rplevka:
@ares - ran all our T1 tests on this with no issues. Thanks!

lzap:
Thank you so much! Merged for 13, if you need this in 12 let me know.

lzap:
I am aligning this bug into 12.0.2 as I think it is important one.

app/models/discovery_rule.rb
```diff
@@ -40,6 +41,10 @@ def default_int_attributes
    self.priority  ||= 0
   end
 
+  def self.suggest_priority
+    self.unscoped.maximum(:priority).to_i + STEP
```
lzap, line 45: Sorry to jump in so late, but I am wondering won't this make the priority increasing across all organizations? Not a big deal, I wonder if this was meant to be.

rabajaj0509: Yeah, this makes the priority increasing across all orgs. I dont think this is meant to be.
Will fix it in my upcoming pr :)

ares: It was meant to behave this way. When this was written, the prority uniqueness validation ignored taxonomy context. If it has changed, the `unscoped` can be removed. But I think the validation still exist or not?

lzap: Oh yes you are right, we are not there yet!

rabajaj0509: let me try to address rule priority in my pr(https://github.com/theforeman/foreman_discovery/pull/450) itself, works? @ares @lzap

ares: Sure

lzap: Sure, since we are actually relaxing the constraint there is no need of migration now. Just make sure it's suggested correctly.

PR #449, opened 2018-07-25
Adding to table header cell word Actions instead of empty cell

tstrych:
The word Actions which is used also in another table headers in foreman, after this correction it will be possible to use the word Actions instead of sixth table header cell.


lzap:
Allright thanks!

PR #450, opened 2018-07-25
Fixes #24236 - Distinguish between organizations while assigning prio…

rabajaj0509:
This patch makes sure there is no conflict between priorities while in different organizations.

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -17,9 +17,9 @@ class DiscoveryRule < ApplicationRecord
   validates :max_count, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates :priority, :presence => true, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates_lengths_from_database
-  validates_uniqueness_of :priority
   before_validation :default_int_attributes
   before_validation :enforce_taxonomy
+  validate :unique_priority, if: :priority_changed?
 
   belongs_to :hostgroup
   has_many :hosts, :dependent => :nullify
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority
+      %w(locations organizations).each do |taxonomy|
+        discovery_rule_ids = TaxableTaxonomy.where(
+          taxable_type: 'DiscoveryRule',
+          taxonomy_id: send("#{taxonomy.singularize}_ids")).pluck(:taxable_id)
+        discovery_rules = DiscoveryRule.unscoped.where(id: discovery_rule_ids)
+        if discovery_rules.pluck(:priority).include?(priority)
+          errors.add(:priority, _("is already taken"))
+        end
+      end
+  end
+
   def default_int_attributes
    self.max_count ||= 0
    self.priority  ||= 0
   end
 
-  def self.suggest_priority
-    self.unscoped.maximum(:priority).to_i + STEP
+  def self.suggest_priority(organization = Organization.current)
+    discovery_rules = DiscoveryRule.unscoped
+    return (discovery_rules.maximum(:priority).to_i + STEP) if organization.nil?
+    discovery_rule_ids = TaxableTaxonomy.where(
+      taxable_type: 'DiscoveryRule',
+      taxonomy_id: organization.id).pluck(:taxable_id)
+    discovery_rules = discovery_rules.where(id: discovery_rule_ids)
+    discovery_rules.maximum(:priority).to_i + STEP
   end
 
   def enforce_taxonomy
--- a/db/migrate/20180724144925_remove_unique_index_on_priority.rb
+++ b/db/migrate/20180724144925_remove_unique_index_on_priority.rb
@@ -1,0 +1,5 @@
+class RemoveUniqueIndexOnPriority < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :discovery_rules, :priority
+  end
+end
--- a/test/unit/discovery_rule_test.rb
+++ b/test/unit/discovery_rule_test.rb
@@ -217,4 +217,26 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
       assert_equal DiscoveryRule::STEP, second_new.priority - first_new.priority
     end
   end
+
+  context 'suggest next priority' do
+    setup do
+      @organization = FactoryBot.create(:organization)
+      @location = FactoryBot.create(:location)
+    end
+
+    test 'when there exists a discovery_rule of the given organization' do
+      hostgroup = FactoryBot.create(:hostgroup, organizations: [@organization], locations: [@location] )
+      discovery_rule = FactoryBot.create(:discovery_rule,
+        priority: rand(100),
+        hostgroup: hostgroup,
+        organizations: [@organization],
+        locations: [@location])
+
+      assert_equal DiscoveryRule.suggest_priority(@organization), (discovery_rule.priority + DiscoveryRule::STEP)
+    end
+
+    test 'when there is no discovery_rule of the given organization' do
+      assert_equal DiscoveryRule.suggest_priority(@organization), DiscoveryRule::STEP
+    end
+  end
 end
--- a/app/controllers/discovery_rules_controller.rb
+++ b/app/controllers/discovery_rules_controller.rb
@@ -12,7 +12,7 @@ class DiscoveryRulesController < ApplicationController
   end
 
   def new
-    @discovery_rule = DiscoveryRule.new(:priority => DiscoveryRule.suggest_priority)
+    @discovery_rule = DiscoveryRule.new(priority: DiscoveryRule.suggest_priority)
   end
 
   def create


app/models/discovery_rule.rb
```diff
@@ -36,6 +36,15 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority_per_organization
+    organizations.each do |org|
```
ares: Could this perhaps be written more effectively by loading all discovery rules of all self.organizations? as far as I understand, this iterate organization over organization. Also I think it ignores nesting of organizations.

app/models/discovery_rule.rb
```diff
@@ -17,9 +17,9 @@ class DiscoveryRule < ApplicationRecord
   validates :max_count, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates :priority, :presence => true, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates_lengths_from_database
-  validates_uniqueness_of :priority
   before_validation :default_int_attributes
   before_validation :enforce_taxonomy
+  validate :unique_priority_per_organization
```
ares: don't we have the same issue with locations?

ares (changes requested):

rabajaj0509:
@ares query looks a better?

app/models/discovery_rule.rb
```diff
@@ -36,6 +36,13 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority_per_organization
+    discovery_rule_ids  =  TaxableTaxonomy.where(taxable_type: 'DiscoveryRule', taxonomy_id: organization_ids).pluck(:taxable_id)
```
ares: it's better but still ignores nested orgs since it's using organization_ids only

lzap: There are also two extra spaces, just a nitpick.

lzap: Rahuj just use business layer methods to return all organizations associated and then get all their ids.

db/migrate/20180724144925_remove_unique_index_on_priority.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveUniqueIndexOnPriority < ActiveRecord::Migration[5.1]
+  def up
+    remove_index :discovery_rules, :priority
+  end
+
+  def down
+    add_index :discovery_rules, :priority, unique: true
+  end
+end
```
lzap, line 5: I am pretty sure Rails devs will deprecate this soon, there is some new syntax I think...

lzap:
I want to do v13 this week, would be great to have it.

app/models/discovery_rule.rb
lzap: Scratch that comment, query is the proper way of finding IDs.

rabajaj0509: @ares @lzap i tried to write a more generic method, using the `Taxonomy.enabled_taxonomies` but when iterating with the location_ids i am facing some issues, which i am not being able to understand. Here is the code that i tried:
```
  def unique_priority
    Taxonomy.enabled_taxonomies.each do |enabled_taxonomy|
      discovery_rule_ids  =  TaxableTaxonomy.where(taxable_type: 'DiscoveryRule', taxonomy_id: send("#{enabled_taxonomy.singularize}_ids")).pluck(:taxable_id)
      if DiscoveryRule.find(discovery_rule_ids).pluck(:priority).include?(priority)
        errors.add(:priority, _("is already taken"))
      end
    end
  end
```

lzap:

app/models/discovery_rule.rb
```diff
@@ -36,6 +36,15 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority
+    Taxonomy.enabled_taxonomies.each do |enabled_taxonomy|
+      discovery_rule_ids  =  TaxableTaxonomy.where(taxable_type: 'DiscoveryRule', taxonomy_id: send("#{enabled_taxonomy.singularize}_ids")).pluck(:taxable_id)
+      if DiscoveryRule.unscoped.find(discovery_rule_ids).pluck(:priority).include?(priority)
```
rabajaj0509: @lzap  @ares i have used `unscoped` here because when i try to find the `discovery_rules_id` in `DiscoveryRule`, i dont get the value. But, as soon i upscope the default_scope i am able to find the value.
Can you explain to me the purpose of using the default_scope?

lzap: Default scope has `with_taxonomy_scope` and some ordering. The taxonomy limits what you can access according to your organization and location. This is actually what we want to use - user in org A should be able to create rule with the same priority as user B. Can you add a test for that actually? Because with `unscoped` this will not be possible.

Looking in the code, I also see that `suggest_priority` does `self.unscoped.maximum(:priority).to_i + STEP` (with unscoped too) which seems to incorrect. You want to have highest priority from your taxonomy, not globally. A fix with a test would be great too :-)

lzap: Wait, taxonomy is somehow being incorporated the line above, then you want to use `unscoped`.

app/models/discovery_rule.rb
```diff
@@ -36,6 +36,15 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority
+    Taxonomy.enabled_taxonomies.each do |enabled_taxonomy|
+      discovery_rule_ids  =  TaxableTaxonomy.where(taxable_type: 'DiscoveryRule', taxonomy_id: send("#{enabled_taxonomy.singularize}_ids")).pluck(:taxable_id)
```
lzap: There are some doublespaces.

Also, put `distinct` call in front of `pluck` to remove extra entries for rules assigned to multiple orgs.

lzap (changes requested):
Looks good but this is missing a test, it's a must in this case.

rabajaj0509:
Now we need to fix the `suggested priority` numbers, I am not sure what would be the better approach for resolving that issue. Can you provide any suggestions on the same?

My thoughts on the above issue are that we use ajax call to find out the highest priority and then add 100 to it as suggested priority.

lzap:
As part of this PR just make sure the `suggest_priority` works across organizations, so use taxonomy when searching for the maximum number.

db/migrate/20180724144925_remove_unique_index_on_priority.rb
lzap: What I mean, use Rails `change` method instead.

rabajaj0509:
The `suggest_priority` method handles three cases:
1) When the org is set to `Any Organization`
2) When the  org has no existing discovery rules
3) When the org has multiple existing discovery rules.

Thanks for the suggestion @lzap, i have made changes to 
the suggest_priority method as suggested by you. Looks good?

db/migrate/20180724144925_remove_unique_index_on_priority.rb
lzap: Going back in time a bit, on IRC you said "one question related to the same, so if i use the `change` method for migration, then while rollback it wont use the `unique: true`". What exactly was the problem? This should work:

    add_index :discovery_rules, column: :priority, unique: true

rabajaj0509: @lzap i guess I had misunderstood the change method, only if there is specific `query` that needs to done we should probably go for the up and down methods, else change should do the work. thanks!

rabajaj0509:
@lzap i am trying to write tests for this but it is seems a little tricky for me! 

So my plan was to create two different discovery rules with different hostgroups, organization and location. 
For doing the same, i have written something like this(please ignore the variable names.):
```
test 'should be able to suggest next priority with orgs' do
      FactoryBot.create(:organization, :name => 'Organization three')
      FactoryBot.create(:location, :name => 'Location three')
      
      hg3 = FactoryBot.create(:hostgroup, organizations: Organization.where(name: 'Organization three'), locations: Location.where(name: 'Location three'))
      three = FactoryBot.create(:new_discovery_rule, name: "NewRule1", :priority => DiscoveryRule.suggest_priority, hostgroup: hg3)
        
      org2 = FactoryBot.create(:organization, :name => 'Organization two')
      loc2 = FactoryBot.create(:location, :name => 'Location two')
      
      hg = FactoryBot.create(:hostgroup, organizations: Organization.where(name: 'Organization two'), locations: Location.where(name: 'Location two'))
      
      two_new = FactoryBot.create(:another_discovery_rule, name: "NewRule", :priority => DiscoveryRule.suggest_priority, hostgroup: hg)
      # binding.pry
      assert_equal two_new.priority, three.priority
end

```

I also tried  to create a new discovery rule in the factory-bot.
Here is what I tried to do:
```
FactoryBot.define do
  factory :discovery_rule do
    sequence(:name) { |n| "rule#{n}" }
    sequence(:priority) { |n| n }
    search { "facts.fact1 == 'value1'" }
    hostname { "" }
    hostgroup
    max_count { 0 }
    enabled { true }
    organizations { [organization_one] } if SETTINGS[:organizations_enabled]
    locations { [location_one] } if SETTINGS[:locations_enabled]
  end

  factory :another_discovery_rule, class: DiscoveryRule do
    sequence(:name) { |n| "rule#{n}" }
    sequence(:priority) { |n| n }
    search { "facts.fact1 == 'value1'" }
    hostname { "" }
    hostgroup
    max_count { 0 }
    enabled { true }
    organizations { [organization_two] } if SETTINGS[:organizations_enabled]
    locations { [location_two] } if SETTINGS[:locations_enabled]
  end
end
```

I was unsuccessful in doing so because of two reasons:
1) I can not use the same `hostgroup` as used earlier otherwise it would have the same org and loc
2) To create a new one i need to create it in the Hostgroup Factory that is present in foreman core.

Can you please help me with the test cases? I might be completely wrong and the fix would be easy but I guess I am not looking in the right direction.

rabajaj0509:
@ares @kgaikwad audit doesnt seem to accept the segregation between the organizations and locations. This might be because in discovery we have `audited :associated_with => :hostgroup`. When I see the hostgroups code in core, hostgroups are not properly isolated between the organizations and locations. Do you think it is time to fix hostgroup isolation, because imo till then audit for discovery rule would fail.

Errors related to audit are:
```
Error:
DiscoveryRuleTest::auditing related to discovery rule#test_0002_should create audit when updating a discovery rule:
ActiveRecord::RecordInvalid: Validation failed: Priority is already taken
    /home/vagrant/foreman_discovery/test/unit/discovery_rule_test.rb:189:in `block (3 levels) in <class:DiscoveryRuleTest>'
    /home/vagrant/foreman_discovery/test/unit/discovery_rule_test.rb:187:in `block (2 levels) in <class:DiscoveryRuleTest>'

Failure:
DiscoveryRuleTest::auditing related to discovery rule#test_0004_should be able to create audit record on changing hostgroup of rule [/home/vagrant/foreman_discovery/test/unit/discovery_rule_test.rb:208]:
"hostgroup_one.associated_audits.count" didn't change by 1.
Expected: 1
  Actual: 0
```

Both are related errors:
```
[3] pry(#<DiscoveryRuleTest::auditing related to discovery rule>)> rule.errors
=> #<ActiveModel::Errors:0x000000000dcb5db0
 @base=
  #<DiscoveryRule:0x000000000dcb7ca0
   id: 42,
   name: "rule5",
   search: "facts.fact1 == 'value1'",
   hostgroup_id: 928644311,
   hostname: "",
   max_count: 0,
   priority: 3,
   enabled: false,
   created_at: Sat, 27 Oct 2018 07:38:37 UTC +00:00,
   updated_at: Sat, 27 Oct 2018 07:38:37 UTC +00:00>,
 @details={:priority=>[{:error=>"is already taken"}, {:error=>"is already taken"}]},
 @messages={:priority=>["is already taken"]}>

```

kgaikwad:
@rahulbajaj0509,

I didn't get what you meant by  `audit doesn't seem to accept the segregation between the organizations and locations`.  In debug output that you shared, it says `:priority=>[{:error=>"is already taken"}, {:error=>"is already taken"}]},  @messages={:priority=>["is already taken"]}>` so there is nothing to do with audit part. Due to this validation error, it is not creating audit record for discovery_rule.

I have few concerns about your changes. Adding those separately for better understanding.

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority
```
kgaikwad, line 39: Does this validation pass for a discovery_rule in case of update?

With the current code, if you try to update a discovery_rule record without change in priority column, it will definitely fail every time.
This is one of reason why you are getting errors for few tests.

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
...
+    Taxonomy.enabled_taxonomies.each do |enabled_taxonomy|
```
kgaikwad: @rahulbajaj0509,
Here, it iterates on `Taxonomy.enabled_taxonomies`. Let's say  `organizations, locations` both are enabled.
But then why the method [suggest_priority](https://github.com/theforeman/foreman_discovery/pull/450/files#diff-dce2846f7e5c2e776201ba13b1445290R61) considered only `organization`.

@lzap, please confirm that this is an expected behavior?

kgaikwad (changes requested):
Added few inline comments.

app/models/discovery_rule.rb
rabajaj0509: Good catch @kgaikwad :) thanks. Will make the corrections.

app/models/discovery_rule.rb
lzap: You are right, only organizations are relevant in this case.

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
...
+      discovery_rule_ids  =  TaxableTaxonomy.where(
```
lzap: Double spaces around `=`.

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
...
+        taxable_type: 'DiscoveryRule',
+        taxonomy_id: send("#{enabled_taxonomy.singularize}_ids")).pluck(:taxable_id)
+      discovery_rules = DiscoveryRule.unscoped.where(id: discovery_rule_ids)
+      if discovery_rules.pluck(:priority).include?(priority) && priority_changed?
+        errors.add(:priority, _("is already taken"))
+      end
```
lzap, line 48: Can this be written without plucking and fetching all priorities? Like DiscoveryRule.unscoped.where(id: discovery_rule_ids, priority: priority)

Anyway, please move the `priority_changed?` to the very top and exit the method early if priority field is unchanged, this saves several SQL queries on every update.

db/migrate/20180724144925_remove_unique_index_on_priority.rb
lzap: Why this is removing priority index? Don't we use it during provisioning when performing auto provisioning?

test/factories/discovery_rule_related.rb
```diff
@@ -10,4 +10,15 @@
...
+  factory :another_discovery_rule, class: DiscoveryRule do
+    sequence(:name) { |n| "rule#{n}" }
+    sequence(:priority) { |n| n }
+    search { "facts.fact1 == 'value1'" }
+    hostname { "" }
+    max_count { 0 }
+    enabled { true }
+    organizations { [organization_two] } if SETTINGS[:organizations_enabled]
+    locations { [location_two] } if SETTINGS[:locations_enabled]
+  end
```
lzap: Please use some name like `discover_rule_in_organization_two`. But `trait` could be used for that I guess.

test/test_helper_discovery.rb
```diff
@@ -108,10 +110,18 @@ def organization_one
...
   Location.find_by_name('Location 1')
 end
 
+def location_two
+  Location.find_by_name('Location 2')
+end
+
 def current_path_info
   current_url.sub(%r{.*?://}, '')[%r{[/\?\#].*}] || '/'
 end
```
lzap: Hmm this will slow down all tests, I'd consider putting this only into the test where you need it. Or make it lazy-loaded.

test/unit/discovery_rule_test.rb
```diff
@@ -172,6 +172,27 @@ class DiscoveryRuleTest < ActiveSupport::TestCase
...
+      hostgroup2 = FactoryBot.create(:hostgroup, organizations: Organization.where(name: 'Organization two'), locations: Location.where(name: 'Location two'))
+
+      # create discovery_rules using difffent taxonomies
+      FactoryBot.create(:discovery_rule, name: "NewRule1", :priority => 100, hostgroup: hostgroup1, organizations: Organization.where(name: 'Organization one'), locations: Location.where(name: 'Location one'))
+
+      FactoryBot.create(:another_discovery_rule, name: "NewRule", :priority => 100, hostgroup: hostgroup2, organizations: Organization.where(name: 'Organization two'), locations: Location.where(name: 'Location two'))
+
+      # check if both have equal suggested priorities.
+      assert_equal DiscoveryRule.suggest_priority(organization2), DiscoveryRule.suggest_priority(organization1)
+  end
```
lzap: Also check the actual value to verify it's not like `-1` or something like that. Also adding one more test where you try to increase priority for org A but suggested priority is unaffected for org B would be useful. Please use test contexts to wrap setup code.

lzap (changes requested):
Looks good, couple of comments.

ares (commented):
makes more sense now, I haven't tested but the code look alright

rabajaj0509:
@lzap based on your previous comments, I have cleaned the tests and all tests are green locally. Irrelevant tests are failing in core, can you please have a look. Thanks!

lzap:
It's green locally, mind rebasing again? All Ruby versions fail this is no good, discovery develop should be green.

rabajaj0509:
@lzap looks green locally, can you have a look at this one?

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
     end
   }
 
+  def unique_priority
+      Taxonomy.enabled_taxonomies.each do |enabled_taxonomy|
```
lzap: All `enabled_taxonomies` must be deleted from the patch, that's the reason why tests fail.

lzap (changes requested):
Rahul, I've realized one thing moments before I wanted to smash that merge button.

Discovery rules are assigned to one or many taxonomies and the current patch correctly validates uniquess on these, but suggestion code only grabs current organization which is actually irrelevant. User can associate a rule with any organization/location he/she has access to. We need to suggest across all taxonomy and that is I believe not possible.

I am starting to think that priority actually must be unique across all organizations.

lzap:
Look, here is the code which finds the rule:

```ruby
DiscoveryRule.where(:enabled => true).reorder(:priority)
```

That is scoped by taxonomix, however this is processed always via API, therefore API admin account which has access to everything anyway:

```ruby
    default_scope lambda {
      with_taxonomy_scope do
        order('discovery_rules.priority, discovery_rules.created_at')
      end
    } 
```

I believe we need to rethink this. Either we must keep priority globally unique or rescope the rule search using detected taxonomy (from subnet). And for this suggestion will be impossible I guess as rules can be in multiple taxonomies (taking highest one will not work good I guess).

lzap:
I think the correct approach is to also fix rule searching code in `find_discovery_rule` method. Let's remove the `with_taxonomy_scope` from the default scope and move it into the method `find_discovery_rule(host.organization, host.location)` because discovered host should already have an organization and location assigned by subnet detection. That should make sure the rules are searched correctly, let's add some tests around this.

Uniqueness test looks sane to me that will work.

Unsure how to approach suggestion code tho, what we can suggest if a rule is being added to organizations A, B and C with maximum priorities 101, 102 and 303? Is that 403? This is already quite confusing.

I think we should start showing taxonomy on discovery rules index/list page, so it's obvious when a rule does not work why.

It's all getting complicated, maybe BJ tomorrow with @ares ?

app/models/discovery_rule.rb
```diff
@@ -36,13 +36,31 @@ class DiscoveryRule < ApplicationRecord
...
 
+  def unique_priority
+      %w(locations organizations).each do |taxonomy|
+        discovery_rule_ids = TaxableTaxonomy.where(
+          taxable_type: 'DiscoveryRule',
+          taxonomy_id: send("#{taxonomy.singularize}_ids")).pluck(:taxable_id)
+        discovery_rules = DiscoveryRule.unscoped.where(id: discovery_rule_ids)
+        if discovery_rules.pluck(:priority).include?(priority)
+          errors.add(:priority, _("is already taken"))
+        end
```
lzap, line 47: I think we should be able to do the check in a single query:

    DiscoveryRule.unscoped.where(id: discovery_rule_ids, priority: priority)

Can you also improve the error message and indicate which taxonomy (org or loc) is conflicting and present the name as well?

lzap:
We had a call and the summary is:

* let's drop priority uniqueness test (introduced in https://projects.theforeman.org/issues/9028)
* let's fix taxonomy scope for discovery rule find process
* let's fix the order of discovery rule processing (and on index page as well): priority first, name second
* let's add organizations/locations to the index page
* let's keep the suggestion in the current form (suggest from the current organization)

rabajaj0509:
Let's start re-work on this. It is a really good to have PR. Thank you for your patience with this one.

PR #454, opened 2018-08-29
Fixes #19737 - kernel/initrd kexec shown in preview

lzap:
This cleans up code a bit. Needs core PR: https://github.com/theforeman/foreman/pull/6010

Edit: The core PR was merged into Foreman 1.20, so this patch can be backported into Discovery 14.0 series as well.

### DIFF
--- a/app/models/host/managed_extensions.rb
+++ b/app/models/host/managed_extensions.rb
@@ -45,7 +45,6 @@ module Host::ManagedExtensions
     template = provisioning_template(:kind => 'kexec')
     raise ::Foreman::Exception.new(N_("Kexec template not associated with operating system")) unless template
     @host = self
-    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
     # try to parse JSON and error out early
     json = JSON.parse(unattended_render(template))
     ::Foreman::Exception.new(N_("Kernel kexec URL is invalid: '%s'"), json['kernel']) unless json['kernel'] =~ /\Ahttp.+\Z/
--- a/app/views/foreman_discovery/debian_kexec.erb
+++ b/app/views/foreman_discovery/debian_kexec.erb
@@ -1,6 +1,6 @@
 <%#
 kind: kexec
-name: Debian kexec
+name: Discovery Debian kexec
 oses:
 - Debian
 - Ubuntu
@@ -32,9 +32,8 @@ Please read kexec(8) man page for more information about semantics.
   options << "inst.stage2=#{@host.operatingsystem.medium_uri(@host)}" if @host.operatingsystem.name.match(/Atomic/i)
 -%>
 {
-  "comment": "WARNING: Both kernel and initram are not set in preview mode due to http://projects.theforeman.org/issues/19737",
-  "kernel": "<%= @kexec_kernel %>",
-  "initram": "<%= @kexec_initrd %>",
+  "kernel": "<%= @kernel_uri %>",
+  "initram": "<%= @initrd_uri %>",
   "append": "url=<%= foreman_url('provision') + "&static=yes" %> interface=<%= mac %> netcfg/get_ipaddress=<%= ip %> netcfg/get_netmask=<%= mask %> netcfg/get_gateway=<%= gw %> netcfg/get_nameservers=<%= dns %> netcfg/disable_dhcp=true netcfg/get_hostname=<%= @host.name %> BOOTIF=<%= bootif %> <%= options.compact.join(' ') %>",
   "extra": []
 }
--- a/app/views/foreman_discovery/redhat_kexec.erb
+++ b/app/views/foreman_discovery/redhat_kexec.erb
@@ -1,6 +1,6 @@
 <%#
 kind: kexec
-name: Red Hat kexec
+name: Discovery Red Hat kexec
 oses:
 - CentOS 4
 - CentOS 5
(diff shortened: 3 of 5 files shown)


app/models/host/managed_extensions.rb
```diff
@@ -45,7 +45,6 @@ def setKexec
     template = provisioning_template(:kind => 'kexec')
     raise ::Foreman::Exception.new(N_("Kexec template not associated with operating system")) unless template
     @host = self
-    @kexec_kernel, @kexec_initrd = operatingsystem.boot_files_uri(@host.medium, @host.architecture, @host)
```
lzap, line 48: By the way this line was broken, tests stub this call.

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -1,6 +1,6 @@
 <%#
 kind: kexec
-name: Debian kexec
+name: Discovery Debian kexec
```
lzap, line 3: This correct the real name as it's being added via seeds.

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -31,9 +31,8 @@ Please read kexec(8) man page for more information about semantics.
   options << "locale=#{@host.params['lang'] || 'en_US'}"
 -%>
 {
-  "comment": "WARNING: Both kernel and initram are not set in preview mode due to http://projects.theforeman.org/issues/19737",
```
lzap, line 35: This patch actually fixes this bug, so the warning can go away.

lzap:

lzap:
Rebased this should pass now @rahulbajaj0509 - make sure to test the migration as well please.

lzap:
Rebased, mind to look? Thanks.

rabajaj0509:
@lzap I am not very familiar with telemetry, can you please give me some reference/steps for testing this pr?

lzap:
This is not the telemetry patch. [test foreman_discovery]

rabajaj0509:
oops, sorry. Was looking into both the prs at the same time.

lzap:
Error during migration: https://github.com/theforeman/foreman_discovery/pull/454/files#diff-e1a732c4a21e925c68fa12a1254f0e2b (ERF42-7495 [Foreman::Exception]: Cannot find user foreman_admin when switching context) looking into ie.

db/migrate/20180829130513_update_kexec_uri.rb
```diff
@@ -0,0 +1,18 @@
+class UpdateKexecUri < ActiveRecord::Migration[5.0]
+  def up
+		ProvisioningTemplate.without_auditing do
+      ["Discovery Red Hat kexec", "Discovery Debian kexec"].each do |template_name|
+        if t = ProvisioningTemplate.unscoped.find_by_name(template_name)
```
iNecas: I would probably go the way https://github.com/inecas/foreman/blob/5963f341d505d73d96b616db02af3d93e06b74c5/db/migrate/20170629080604_change_preseed_default_pxe_grub2_kind.rb#L9 does it, so that we still get audit logs for existing installations: might be useful for the seeing this changes.

lzap: Ok done thanks.

iNecas:
Test failures related

lzap:
Yeah rebased. The failing test had to go since it was testing the code which I deleted actually.

lzap:
I'd appreciate quick merge since I need to rebase https://github.com/theforeman/foreman_discovery/pull/468 on top of this.

db/migrate/20180829130513_update_kexec_uri.rb
```diff
@@ -0,0 +1,30 @@
+class UpdateKexecUri < ActiveRecord::Migration[5.0]
+  def update_templates
+    ProvisioningTemplate.without_auditing do
```
iNecas: you probably want to remove the `without_auditing` from here

iNecas:
One last comment

lzap:
Rebased, tests failed, couple of (random?) failures around auditing in core. Let's see.

lzap:
@iNecas I'd appreciate merge or at least giving it a go ASAP.

db/migrate/20180829130513_update_kexec_uri.rb
```diff
@@ -0,0 +1,28 @@
+class UpdateKexecUri < ActiveRecord::Migration[5.0]
+  def update_templates
+    ["Discovery Red Hat kexec", "Discovery Debian kexec"].each do |template_name|
+      if t = ProvisioningTemplate.unscoped.find_by_name(template_name)
+        contents = t.template.gsub(/@kexec_kernel/, '@kernel_uri').gsub(/@kexec_initrd/, '@initrd_uri')
```
iNecas: The migration is not removing the ` "comment": "WARNING: Both kernel and initram are not set in preview mode due to http://projects.theforeman.org/issues/19737",` line.

db/migrate/20180829130513_update_kexec_uri.rb
```diff
@@ -0,0 +1,28 @@
...
+        t.update_attributes(:template => contents)
```
iNecas: Btw. `update_attributes` is skipping the audit creation. Given we would need to fix it by `t.save!`, which has a risk of us hitting some issues with invalid user data (Using a fake model is not an option either, as the audited would use the fake's class name to store the audit.

I guess it's not worth the effort and we could just call `without_auditing` and not caring about the audits at all here.

iNecas (changes requested):

db/migrate/20180829130513_update_kexec_uri.rb
lzap: Since discovery kexec templates are now have the locking mechanism from core, I wonder if it is worth simply ditching this migration for seed script update. Currently it does not overwrite, I think it should overwrite them.

iNecas: Yes, I think that's option as well. It seemed like you wanted to have the migration around, so I reviewed that. Feel free to remove the migration completely

lzap:
@iNecas all finally green.

iNecas (approved):
Thanks @lzap. No other comments from me. Ready to go

lzap:
@ares or @rahulbajaj0509 can you please merge? I need to rebase the other PR on top of this. Thanks.

ares (approved):

ares:
Thanks @lzap, merging

PR #457, opened 2018-09-10
[WIP] Fixes #20549 - Use Multiexception for reboot all method

rabajaj0509:
This pr uses https://github.com/theforeman/foreman/pull/5785 for multi-exception. While the pr is still in the [WIP] mode, you can test it by having the  original code in the [hammer](https://github.com/theforeman/hammer-cli-foreman-discovery/pull/38/files#diff-5e45be4d27eedc48fe561b6cf4775195L147)

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -210,16 +210,9 @@ module Api
       api :PUT, "/discovered_hosts/reboot_all", N_("Rebooting all discovered hosts")
 
       def reboot_all
-        error_message = perform_reboot_all
-        if error_message
-          render_error :custom_error,
-                       :status => :unprocessable_entity,
-                       :locals => {
-                           :message => error_message
-                       }
-        else
-          process_success _("Discovered hosts are rebooting now")
-        end
+        exceptions = perform_reboot_all
+        return process_success _("Discovered hosts are rebooting now") unless exceptions.present?
+        render :json => (::Foreman::MultiException.new(exceptions,"Errors During Rebooting")).as_json
       end
 
       private
--- a/app/controllers/concerns/foreman/controller/discovered_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/discovered_extensions.rb
@@ -69,32 +69,27 @@ module Foreman::Controller::DiscoveredExtensions
   end
 
   def perform_reboot_all hosts = Host::Discovered.all
-    result = true
-    error_message = _("Errors during reboot: %s")
-    overall_errors = ""
+    overall_errors = []
 
     if hosts.count > 0
       hosts.each do |discovered_host|
         begin
           unless discovered_host.reboot
-            error =  "#{discovered_host.name}: failed to reboot "
-            overall_errors << error
-            logger.error error
+            msg = "#{discovered_host.name}: Failed to reboot "
+            e = Excpetion.New()
+            overall_errors << WrappedException.new(e, msg)
+            logger.error e.message
           end
         rescue Exception => e
-          error = "#{discovered_host.name}: #{e.to_s} "
-          overall_errors << error
-          logger.error error
+          overall_errors << e
+          logger.error e.message
         end
       end
     else
-      error_message = _("No discovered hosts to reboot")
-      result = false
+      msg = _("No discovered hosts to reboot")
+      overall_errors << WrappedException.new(Exception.new(), msg)
     end
-
-    if overall_errors.present? || !result
-      error_message % overall_errors.strip
-    end
+    overall_errors
   end
 
 end
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -125,7 +125,7 @@ class DiscoveredHostsController < ::ApplicationController
     error_message = perform_reboot_all(@hosts)
 
     if error_message
-      process_error :error_msg => error_message, :redirect => :back
+      process_error :error_msg => error_message.first.message, :redirect => :back
     else
       process_success :success_msg => _("Discovered hosts are rebooting now"), :success_redirect => :back
     end


app/controllers/discovered_hosts_controller.rb
```diff
@@ -123,7 +123,7 @@ def submit_multiple_reboot
     error_message = perform_reboot_all(@hosts)
 
     if error_message
-      process_error :error_msg => error_message, :redirect => :back
+      process_error :error_msg => error_message.first.message, :redirect => :back
```
rabajaj0509, line 128: this was just for the testing purpose, will remove it.

rabajaj0509:
Current output looks like: 

screenshot from 2018-09-10 15-29-53

@lzap one thing i was unable to understand here is that, as shown in the picture above, in the array the error is getting printed twice, not sure how.

lzap:
If you add `[skip ci]` to the second line of the commit message, it won't do CI. However, it's red because of Rails 5.2 migration.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -69,32 +69,27 @@ def perform_auto_provision original_host, rule
...
 
     if hosts.count > 0
       hosts.each do |discovered_host|
         begin
           unless discovered_host.reboot
-            error =  "#{discovered_host.name}: failed to reboot "
-            overall_errors << error
-            logger.error error
+            msg = "#{discovered_host.name}: Failed to reboot "
+            e = Excpetion.New()
```
timogoebel, line 79: typo here? I guess this should be `StandardError.new`.

app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -196,16 +196,9 @@ def reboot
...
-                       :status => :unprocessable_entity,
-                       :locals => {
-                           :message => error_message
-                       }
-        else
-          process_success _("Discovered hosts are rebooting now")
-        end
+        exceptions = perform_reboot_all
+        return process_success _("Discovered hosts are rebooting now") unless exceptions.present?
+        render :json => ::Foreman::MultiException.new(exceptions,"Errors During Rebooting").json_form
```
lzap: Missing space, also i18n: `::Foreman::MultiException.new(exceptions, N_("Errors during rebooting"))`

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -69,32 +69,27 @@ def perform_auto_provision original_host, rule
...
+            overall_errors << WrappedException.new(e, msg)
```
lzap, line 80: Do not use WrappedException if you don't have any exception to wrap. Just use Foreman::Exception and give it some text, then you won't see this twice as well.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -69,32 +69,27 @@ def perform_auto_provision original_host, rule
...
       end
     else
-      error_message = _("No discovered hosts to reboot")
-      result = false
-    end
-
-    if overall_errors.present? || !result
-      error_message % overall_errors.strip
+      msg = _("No discovered hosts to reboot")
+      overall_errors << WrappedException.new(Exception.new(), msg)
```
lzap, line 90: Ditto, simply use Foreman::Exception and that's it.

app/controllers/concerns/foreman/controller/discovered_extensions.rb
```diff
@@ -69,32 +69,27 @@ def perform_auto_provision original_host, rule
...
         end
       end
     else
-      error_message = _("No discovered hosts to reboot")
-      result = false
-    end
-
-    if overall_errors.present? || !result
-      error_message % overall_errors.strip
+      msg = _("No discovered hosts to reboot")
```
lzap, line 89: It is important to send messages *untranslated* into Foreman::Exception, see https://projects.theforeman.org/projects/foreman/wiki/Translating. Hint: `_()` vs `N_()`.

lzap (changes requested):
Overall looks great, this can leave WIP mode now. Couple of things here and there around i18n but good work!

rabajaj0509:
@lzap your suggestion about not wrapping an exception if there is no need to made life so much easy, thanks!

 So i figured out that wrapping the exception [here](https://github.com/theforeman/foreman_discovery/blob/develop/app/models/host/discovered.rb#L163) causes  the exception to be printed twice. 

The `perform_reboot_all` method sends an exception to the `reboot` method mentioned in the above link,
we tend to Wrap that exception again, that is the reason that can see the error printed twice.

For better exception handling i would suggest we  use 
```
diff --git a/app/models/host/discovered.rb b/app/models/host/discovered.rb
index a3fd59f..73a1ca9 100644
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -224,7 +224,7 @@ class Host::Discovered < ::Host::Base
     resource.reboot
   rescue => e
     ::Foreman::Logging.exception("Unable to reboot #{name}", e)
-    raise ::Foreman::WrappedException.new(e, N_("Unable to reboot %{name} via %{url}: %{msg}"), :name => name, :url => proxy_url, :msg => e.to_s)
+    raise ::Foreman::Exception.new(N_("Unable to reboot %{name} via %{url}: %{msg}"), :name => name, :url => proxy_url, :msg => e.to_s)
   end
```

Also, the output with the above suggestion looks like:
screenshot from 2018-10-05 21-22-51

What are your thoughts on this? do you think we should go ahead with the above mentioned changes?

lzap:
The result image looks great, exactly how I like it, it's simple approach to the problem.

Now, I'd rather keep using wrapped exception in that reboot method, it's good use of this class actually. We send a copy of the message in the wrapped message for one purpose - it's way too long to render WrappedException and we render only the message in the UI because space is limited there (toast error). Just keep it, it won't hurt.

If you want to remove be sure to catch the exception and render it in a better way for example using some hover or expansion button. But I'd just ignore it for now, no biggie.

rabajaj0509:
Thanks for the go ahead @lzap, will push the changes :)

lzap:
Pls rebase tests are green now.

lzap:
Red because of a typo: uninitialized constant Foreman::Controller::DiscoveredExtensions::Excpetion.

rabajaj0509:
Let's start re-work on this. It is a really good to have PR. Thank you for your patience with this one.

rabajaj0509:
@lzap can you please assign this PR to someone else?

lzap:
We will pick it from here, thank you!

PR #460, opened 2018-10-02
Fixes #25101 - append stage2 for Atomic OS

lzap:
Also cleans option handling a bit.


lzap:
Failure parsing Discovery Red Hat kexec: bare backslash only allowed before newline. near line 24: "\",\\n\".freeze".

Weird.

lzap:
Rebased this should pass now @rahulbajaj0509

lzap:
I was thinking using snapshotting service, but long term we would like to move rendering tests into community-templates and then we would move discovery templates there as well.

rabajaj0509:
@lzap smooth! thanks, merging this now :)

PR #465, opened 2018-12-03
Fixes #25606 - fixed URL in discovery mailer

lzap:
Tests were passing because we were only testing for the presence of the base URL, which is at the end of the mail anyway. This now creates new discovered host and perform the test correctly. FYI you can't use `_path` helpers in mails, links must be all absolute (via `_url` helpers).


lzap:
@rahulbajaj0509 hey after this is merged, we need this to be in Foreman 1.20. Do you want to create stable branch in git and cherry pick this patch for 14.0.1? Basically all you need to do is to checkout label `14.0.0`, create new branch called `14-stable` and cherry pick this patch into it, then push. Let's wait with releasing 14.0.1 for now we will likely have more bugs.

rabajaj0509:
@lzap change looks good :) merging now! Thanks Lukas.

rabajaj0509:
@lzap hey i am not able to see the 14.0.0 label created, if I am correct I need to create that label and then create a new branch `14-stable`. How to relate the two? how do we decide when to release?

lzap:
Sorry my bad, I forgot to push the latest tag. Now it's there and you can proceed creating new branch and cherry picking this one. Thanks!

PR #466, opened 2018-12-27
Fixes #25757 - remove undefined method disable_taxonomies

rabajaj0509:


kgaikwad (approved):
Code wise looks legitimate!

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -84,27 +84,27 @@ def test_edit_form_build_mode_enabled
...
-
-      managed_host = Host.find(host.id)
-      assert managed_host.build
-      assert_redirected_to host_url(managed_host)
-      assert_equal hostgroup.id, managed_host.hostgroup_id
-      assert_match(/Successfully/, flash[:success])
-    end
+    organization = FactoryBot.create(:organization)
+    location = FactoryBot.create(:location)
+    domain = FactoryBot.create(:domain, :organizations => [organization], :locations => [location] )
```
tbrisker: You don't need to manually create the taxomonies, the domain factory assigns them by default - https://github.com/theforeman/foreman/blob/develop/test/factories/domain.rb#L9 and so does the hostgroup factory, so if you assign different ones here you may get wrong result

rabajaj0509: good catch, thanks! rebased with the corrections.

test/functional/discovered_hosts_controller_test.rb
```diff
@@ -84,27 +84,25 @@ def test_edit_form_build_mode_enabled
...
-        } }, session: set_session_user_default_manager
-
-      managed_host = Host.find(host.id)
-      assert managed_host.build
-      assert_redirected_to host_url(managed_host)
-      assert_equal hostgroup.id, managed_host.hostgroup_id
-      assert_match(/Successfully/, flash[:success])
-    end
+    domain = FactoryBot.create(:domain)
+    hostgroup = FactoryBot.create(:hostgroup, :with_subnet, :with_environment, :with_rootpass, :with_os, :domain => domain, organizations: domain.organizations, locations: domain.locations)
```
tbrisker: I'm not sure setting org/loc here is needed as well, the factory default should be the same as the domain one.

tbrisker (commented):
Looks like there are 10 other tests failing

test/functional/discovered_hosts_controller_test.rb
rabajaj0509: Yeah, earlier I had not set the loc/org here but still I was facing same failures. So I just gave this a try.

tbrisker (approved):
While there are failing tests, this at least fixes this test. merging for now and looking forward to another pr to fix the rest of them.

rabajaj0509:
@tbrisker thanks! I am almost done with the other pr that fixes the rest of the tests :)

PR #471, opened 2019-02-21
Fixes #26100 - drop foreman_organization and tests

lzap:
Tests to make sure this regression does not happen again:

https://github.com/theforeman/foreman/pull/6478

In order to make the tests effective, this patch also removes
"foreman_organization" and "foreman_location" taxonomy detection, this can be
easily changed by users to the same fact but with different name:
"discovery_organization" and "discovery_location".


rabajaj0509:
Looks pretty straight forward, and the thanks for the great test @lzap :) merging this!

PR #472, opened 2019-02-28
Fixes #26184 - skip callbacks on host save

lzap:
I found the following error in test logs when running core tests:

    Failed to handle notifications - this is most likely a bug: undefined method `notifications' for nil:NilClass

It is caused by the host extension:

    after_save :delete_discovery_attribute_set, :update_notifications

Both methods are called when there is a save on Host::Managed, we need to do this because of the STI. However we should not attempt to delete notifications if `type` was not previously changed. Only when type changed that's the discovery provisioning and that's when we want to delete it.

The same for deleting attribute set - an associated 1:1 record.

### DIFF
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -34,7 +34,7 @@ module DiscoveredHostsHelper
 
     select_action_button( _("Select Action"), {:id => 'submit_multiple'},
       actions.map do |action|
-        link_to_function(action[0], "build_modal(this, '#{action[1]}')", :'data-dialog-title' => _("%s - The following hosts are about to be changed") % action[0]) if authorized_for(action[2])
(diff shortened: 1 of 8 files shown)


rabajaj0509:
Locally does not seem to fail, can you re-run the tests?

timogoebel:
[test foreman_discovery]

timogoebel:
[test foreman_discovery]

lzap:
It's broken again, started https://community.theforeman.org/t/the-asset-host-checkbox-js-is-not-present-in-the-asset-pipeline/13744

lzap:
PR that aims to fix tests: https://github.com/theforeman/foreman_discovery/pull/475

Then I will rebase this one.

lzap:
Ok pulled in from #475 and fixed the remaining problem too. It's a separate commit, please keep it like that just in case we wanted to backport the original issue. I don't want to create another PR and rebase this if you don't insist, thanks!

mmoll:
the bundler problem is something different, but two runs were green, so I'd merge this :+1:

timogoebel:
Thanks, @lzap and @mmoll.

PR #473, opened 2019-03-01
Fixes #26204 - expose fips_enabled param

iNecas:


lzap (approved):
This is good, I am unable to find any info how to enable this on Ubuntu (they announced they "support" it but there is huge lack of info on how). Debian is off the table, so only RH template then.

PR #474, opened 2019-03-04
Fixes #26127 - adds nokaslr kernel cmd option

lzap:
According to our testing, this helps with kexecing 7.6+ kernels with have kernel space randomization feature enabled.


lzap:
No risk, no need to wait with this one.

PR #476, opened 2019-05-07
Fixes #26759 - snippet can be used in kexec

lzap:
This makes use of the unattender renderer. Finally we can use the same renderer as in unattended controller to make sure preview is always exactly the same.


timogoebel (approved):
Looks good, but tests are red. Looks unrelated, though.

timogoebel:
Thanks, @lzap.

PR #480, opened 2019-07-11
Fixes #27278 - fix the form location

rabajaj0509:
The wrong location of the form does not allow to load the ajax call properly. This sometimes leads to non-functioning of the buttons placed on the modal.

This pr makes sure that the buttons are working and the form is placed correctly.

### DIFF
--- a/app/views/discovered_hosts/_discovered_host_modal.html.erb
+++ b/app/views/discovered_hosts/_discovered_host_modal.html.erb
@@ -4,25 +4,25 @@
       <div class="modal-header">
         <h4 class="modal-title" id="fixedPropertiesSelectorLabel">Select initial host properties</h4>
       </div>
-      <div class="modal-body">
-        <%= form_for host, :url => edit_discovered_host_path(host), :method => :get do |f| %>
-            <%= select_f f, :hostgroup_id, accessible_resource_records(:hostgroup, :title).to_a, :id, :to_label,
-                         { :include_blank => true },
-                         { :help_inline => :indicator, :size => 'col-md-10' } %>
+      <%= form_for host, :url => edit_discovered_host_path(host), :method => :get do |f| %>
+        <div class="modal-body">
+          <%= select_f f, :hostgroup_id, accessible_resource_records(:hostgroup, :title).to_a, :id, :to_label,
+                      { :include_blank => true },
+                      { :help_inline => :indicator, :size => 'col-md-10' } %>
 
-            <% if show_organization_tab?  %>
-                <%= select_f f, :organization_id, Organization.my_organizations, :id, :to_label, {}, {:size => 'col-md-10'} %>
-            <% end %>
+          <% if show_organization_tab?  %>
+            <%= select_f f, :organization_id, Organization.my_organizations, :id, :to_label, {}, {:size => 'col-md-10'} %>
+          <% end %>
 
-            <% if show_location_tab? %>
-                <%= select_f f, :location_id, Location.my_locations, :id, :to_label, {}, {:size => 'col-md-10'} %>
-            <% end %>
-      </div>
-      <div class="modal-footer">
-        <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
-        <input type="submit" class="btn btn-default" value="<%= _('Customize Host') %>" />
-        <input type="submit" class="btn btn-primary" value="<%= _('Create Host') %>" name="quick_submit" />
-      </div>
+          <% if show_location_tab? %>
+            <%= select_f f, :location_id, Location.my_locations, :id, :to_label, {}, {:size => 'col-md-10'} %>
+          <% end %>
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+          <input type="submit" class="btn btn-default" value="<%= _('Customize Host') %>" />
+          <input type="submit" class="btn btn-primary" value="<%= _('Create Host') %>" name="quick_submit" />
+        </div>
       <% end %>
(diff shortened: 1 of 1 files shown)


ares:
fixes the issue, waiting for tests to finish, APJ

ares:
This needs to be better investigated, the fact it breaks tests and it's not clear why it helps indicates, we need to find the root cause.

rabajaj0509:
Yes, was looking into this today, was not able to figure out the root cause yet, but i am working on it.

lzap:
Guys, these tests seem to be broken due to changes in core. It is very unlikely this creates any regression. Welcome to the world of discovery plugin!

However we should fix those tests first, since this is high priority can you take a look @rahulbajaj0509 ? We need to get tests fixed in a separate commit so we can easily cherry pick if needed.

lzap:
Guys, a blocker bug was reported by Satellite QA and it appears to be the same error messages. I investigated the issue and my patch which I am going to file soon fixes most of the failures. Looks like some regression in core or rails (upgade) in regard to 1:1 relations.

lzap:
This PR should resolve most failures: https://github.com/theforeman/foreman_discovery/pull/482

amirfefer:
@ares After a deeper investigation, I think the root cause is the location of the buttons which is outside of the `form` scope, and there is no script that connects between the form and the buttons AFAIK. Looks like with turbolinks the browser doesn't refer these buttons to their form. moving the buttons to be inside of the `form` tag is the solution and not just a workaround IMHO.

app/views/discovered_hosts/_discovered_host_modal.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="modal fade" id="<%= "fixedPropertiesSelector-#{host.id}" %>" tabindex="-1" role="dialog" aria-labelledby="fixedPropertiesSelectorLabel">
   <div class="modal-dialog" role="document">
     <div class="modal-content">
+      <%= form_for host, :url => edit_discovered_host_path(host), :method => :get do |f| %>
```
amirfefer: how about moving this form_for to be above `modal-body` div? I'm not sure if the header should be inside the form, plus please indent the correlated `end`

rabajaj0509: I think this should do the trick!

ares:
Thanks @amirfefer for looking into this! As Tomer also explained me this morning, it's probably changed due to how new browsers deal with invalid html and that this probably doesn't work with turbolinks.

Let's see how tests turn out, if they are green now, I can retest the new patch and check for regressions.

lzap:
Tests are green, except one DB post cleanup rake task failure. We occasionally get these.

I wonder if there is a JS helper tool that can check/validate HTML/XHTML page and show a warning if it is not good. We could add such a tool into our Dev menu toolbox.

lzap:
So I tested this using HTML validator extension for Firefox and it looks like before and after the patch we still have problems with HTML tags:

```
Info | 0 | 0 | 0 errors / 29 warnings (81 hidden) |  
-- | -- | -- | -- | --
Warning | 225 | 84 | Warning: missing </span> before <div> |  
Warning | 225 | 146 | Warning: inserting implicit <span> |  
Warning | 225 | 152 | Warning: discarding unexpected </span> |  
Warning | 286 | 84 | Warning: missing </span> before <div> |  
Warning | 286 | 146 | Warning: inserting implicit <span> |  
Warning | 286 | 152 | Warning: discarding unexpected </span>
```

lzap:
Looks like these span vs div errors are generated by `select_f` and maybe not relevant. I think the validator wants us not to have div inside span which we have. I fail to understand the explanation by @amirfefer however I will try to reproduce the issue now.

ares:
I spoke with @amirfefer and the introducing change was the fact we now use turbolinks for pages withoit react router. In combination with our invalid html, it causes the issue. So it basically uncovered problem we had. Therefore I'm +1 for the patch, given @lzap confirms it works.

lzap:
I was not able to see the bug, but I also don't have specific steps to reproduce it. Once this gets approval from @amirfefer I think we can give it a try - at least it does not break it and I can only hope ;-)

If you were able to reproduce this share steps how to do this.

ares:
I put repro steps in BZ but can retest myself if that helps. It was pretty straightforward. I hope to get to it by end of the week.

lzap:
That's what I don't understand, is it the reproducer or a workaround?

There's a workaround for the same functionality in UI that works

Steps to Reproduce:
1. Discover the host in the satellite.
2. Create an hostgroup for provisioning the above-discovered host.
3. Go to Hosts -> Discovered Hosts.
4. Don't go to discovered host detail page, instead on the table row, navigate to right and click Provision
5. From Provision dialog, choose HostGroup, Org, and Location.
6. Click 'Create Host' or 'Customize Host' button.

This flow (change is in step 4) works fine.

lzap:
Anyway, I tried this about 30 times and it works fine. I don't know how to trigger turbolinks tho, if there are no objections I am going to merge this today.

lzap (approved):
Works fine, code looks good, no errors via HTML validator (except few warnings like div inside span which is not caused by discovery anyway).

ares:
the repro steps were trivial:
* go to discovered host detail, then use breadcrumbs to get back to index and click provision, then click submit host button
or
* go to discovered host index, click host name to navigate to it's detail, then use select action drop down and chose provision, try to click hosts

in both cases, no action happens without the fix

ares (approved):
The patch fixes the issue and I didn't find any regression, thanks @rahulbajaj0509, @amirfefer and @lzap, merging!

PR #481, opened 2019-07-18
Fixes #27335 - getting rid of Rails 6 incomp

lzap:
I have never understood how this "help" pages work. In discovery plugin I have a welcome.html.erb view that renders "No discovered hosts found in this context" page when there are no records and this has been working fine for years.

This suggested patch adds ability to use /discovered_hosts/help URL which was previously not working, but I don't believe there is any link in the discovery plugin to this URL. So I am not sure how useful the route is.

### DIFF
--- a/extra/discover-host
+++ b/extra/discover-host
@@ -16,6 +16,7 @@ base = find_base
 version = "3.4.0"
 interfaces = []
 preserve_interfaces = false
+use_facts_endpoint = false
 primary = nil
 bootif = nil
 url = "http://admin:changeme@localhost:3000"
@@ -38,6 +39,10 @@ OptionParser.new do |opts|
 
   opts.on("-x", "--[no-]preserve-interfaces", "Preserve interfaces from JSON") do |v|
     preserve_interfaces = true
+  end
+
+  opts.on("-a", "--facts", "Use host facts endpoint") do |v|
+    use_facts_endpoint = true
   end
 
   opts.on("-pNAME", "--primary=NAME", "Interface to use as primary (defaults to the first one)") do |v|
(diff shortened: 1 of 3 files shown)


rabajaj0509 (commented):
@lzap  I think apparently the welcome page is rendered by the `model_of_controller` method.

We have a method `model_of_controller` in the core, that is overridden in discovery [here](https://github.com/theforeman/foreman_discovery/blob/develop/app/controllers/discovered_hosts_controller.rb#L19).

In core, in the [application_controller.rb](https://github.com/theforeman/foreman/blob/d1e18585baf05ab9530b6548b2f8873f067393b6/app/controllers/application_controller.rb#L30) file, we check if the `model_of_controller` is nil or not, if not, then it renders the welcome page.

Does that make sense? Let me know what are your thoughts on the way we are rendering it now.

lzap:
Yes, that I get, however there is no link pointing to `/help` URL. Thus the route is not needed. Anyway, I am rebasing this now let's merge this.

lzap:
Ping-

timogoebel (approved):
Makes sense to me.

lzap:
Rebased to see if tests can pass then.

timogoebel:
🔴 , sorry.

"permission for discovered_hosts/welcome not found, check access_permissions.rb...."

I guess it should be enough to define a permission for the route in `engine.rb`.

lzap:
All good now! Thanks, RTM.

extra/discover-host
```diff
@@ -16,6 +16,7 @@ base = find_base
 version = "3.4.0"
 interfaces = []
 preserve_interfaces = false
+use_facts_endpoint = false
```
timogoebel, line 19: How is this related?

lzap: It is not, do you insist on splitting this into two PRs?

timogoebel (approved):

timogoebel:
Thanks, @lzap.

PR #482, opened 2019-07-18
Fixes #27340 - fixes tests and attribute set validators removed

lzap:
The initial discovery upload fails with 422 ActiveRecord::RecordNotSaved: Failed to save the new associated discovery_attribute_set. The problem is that Rails presence validator errors out when host_id is not set when discovered host is not yet saved. This happens because update call calls save when host_id is not yet set.

As a consequence, discovery_attribute_set is not set however discovered host is saved with null reference, therefore subsequent discovery fact upload already has host.id set therefore this should save. Unfortunately due to the other validator (:discovery_attribute_set, :presence => true) it won't because then it fails with a different error 422 Validation failed: Discovery attribute set can't be blank.

This patch removes the unnecessary validators because the code is ready today that attributes association can be blank. This was added because when discovery fails, it is set to blank anyway. This could be prevented by utilizing SQL transactions, however we've been there and it caused huge pains in our orchestration and it was reverted. I left a note about this in the codebase.

### DIFF
--- a/app/models/discovery_attribute_set.rb
+++ b/app/models/discovery_attribute_set.rb
@@ -3,5 +3,4 @@ class DiscoveryAttributeSet < ApplicationRecord
 
   validates :cpu_count, :presence => true, :numericality => {:greater_than_or_equal_to => 0}
   validates :memory, :presence => true,    :numericality => {:greater_than_or_equal_to => 0}
-  validates :host, :presence => true
 end
--- a/app/services/foreman_discovery/import_hooks/discovery_attribute.rb
+++ b/app/services/foreman_discovery/import_hooks/discovery_attribute.rb
@@ -3,8 +3,15 @@ module ForemanDiscovery
   module ImportHooks
     class DiscoveryAttribute < ImportHook
       def after_populate
-        host.discovery_attribute_set = DiscoveryAttributeSet.where(host_id: host.id).first_or_create
-        host.discovery_attribute_set.update(import_from_facts)
+        # This should be done in a SQL transaction but we cannot use transactions for
+        # host updates because it utterly breaks our orchestration. We must work with
+        # the fact that discovery_attribute_set can end up being nil and controllers
+        # and views must be aware and render zero/blank values.
+        if host.discovery_attribute_set
+          host.discovery_attribute_set.update(import_from_facts)
+        else
+          host.build_discovery_attribute_set(import_from_facts)
+        end
       end
 
       def import_from_facts
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -9,8 +9,7 @@ class Host::Discovered < ::Host::Base
   include ::Hostext::OperatingSystem
 
   has_one :discovery_attribute_set, :foreign_key => :host_id, :dependent => :destroy
-
-  validates :discovery_attribute_set, :presence => true
+  validates_associated :discovery_attribute_set
 
   delegate :memory, :cpu_count, :disk_count, :disks_size, :to => :discovery_attribute_set, :allow_nil => true
   after_destroy :delete_notification
--- a/test/functional/discovered_hosts_controller_test.rb
+++ b/test/functional/discovered_hosts_controller_test.rb
@@ -86,7 +86,7 @@ class DiscoveredHostsControllerTest < ActionController::TestCase
   def test_edit_form_quick_submit
     domain = FactoryBot.create(:domain)
     hostgroup = FactoryBot.create(:hostgroup, :with_subnet, :with_environment, :with_rootpass, :with_os, :domain => domain)
-    new_ip = hostgroup.subnet.ipaddr
+    new_ip = hostgroup.subnet.ipaddr.to_s
     host = discover_host_from_facts(@facts.merge({
                                                     'ipaddress' => new_ip,
                                                     'ipaddress_eth0' => new_ip
--- a/test/integration/discovered_hosts_test.rb
+++ b/test/integration/discovered_hosts_test.rb
@@ -9,6 +9,7 @@ class DiscoveredHostsTest < IntegrationTestWithJavascript
     assert discovered_notification_blueprint
     discovered_host.save!
     visit discovered_hosts_path
+    page.evaluate_script("document.getElementById('fixedPropertiesSelector-#{discovered_host.id}').classList.remove('fade')")
   end
 
   teardown do
@@ -64,15 +65,14 @@ class DiscoveredHostsTest < IntegrationTestWithJavascript
 
     test 'and forwards to editing it' do
       create_host
-      assert_equal edit_discovered_host_path(id: discovered_host),
-                   current_path
+      assert_current_path edit_discovered_host_path(id: discovered_host), ignore_query: true
     end
 
     context 'with a Hostgroup selected' do
       let(:discovery_hostgroup) { Hostgroup.first }
 
       test 'it passes it on' do
-        select_from('host_hostgroup_id', discovery_hostgroup.id)
+        select2(discovery_hostgroup.name, from: 'host_hostgroup_id')
         create_host
         assert_param discovery_hostgroup.id.to_s,
                      'host.hostgroup_id'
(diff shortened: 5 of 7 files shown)


lzap:
One more test will fail and I am able to figure out: "can search discovered hosts by disks_size" fails because foreman core now filters out blockdevice facts by default via setting. Can do in a different commit

However I believe there are still some tests failing. Locally at least, many of them are integration test failures: Selenium::WebDriver::Error::UnknownError: unknown error: Devtools port number file contents <41755> were in an unexpected format

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
```diff
@@ -3,8 +3,13 @@ module ForemanDiscovery
   module ImportHooks
     class DiscoveryAttribute < ImportHook
       def after_populate
-        host.discovery_attribute_set = DiscoveryAttributeSet.where(host_id: host.id).first_or_create
```
timogoebel, line 6: Can't you just do `host.discovery_attribute_set || host.build_discovery_attribute_set`.

`create` actually tried to save the record, `build_discovery_attribute_set` will just create the record in memory. The `update` call should then save this.

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
```diff
@@ -27,6 +32,14 @@ def import_from_facts
           disk_count: disk_count,
           disks_size: disks_size
         }
+      rescue Exception => e
+        ::Foreman::Logging.exception("Unable to parse attributes from facts", e)
```
timogoebel: Do you think it's a good idea to swallow any errors here? Users will never report bugs and we'll never fix them if we just log them.

app/models/discovery_attribute_set.rb
```diff
@@ -3,5 +3,4 @@ class DiscoveryAttributeSet < ApplicationRecord
 
   validates :cpu_count, :presence => true, :numericality => {:greater_than_or_equal_to => 0}
   validates :memory, :presence => true,    :numericality => {:greater_than_or_equal_to => 0}
-  validates :host, :presence => true
```
timogoebel, line 6: I think we should try to keep this, see my earlier comment for a possible solution.

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
lzap: Yes this is intentional in this case, if there is a parsing bug then we at least end up with a valid discovered host and the error is logged properly.

app/models/discovery_attribute_set.rb
lzap: There are many other reasons why discovery process can fail and if there is no attribute associated the discover host gets stuck. Any attempt to save it lead to validation error. And since I am removing the other side of the validation, I see no particular reason why to have it here.

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
lzap: Let me try that then.

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
lzap: However if you don't like this, I can remove. It was added as an "extra idea"...

lzap:
I've split the commit into two - one is only dedicated to fix the attribute set issue. The other one fixes several unrelated test failures. But integration tests are still failing, Ondrej is taking a look:

https://community.theforeman.org/t/discovery-integration-tests-broken-due-to-selenium/14675

lzap:
I've split the commit into two - one is only dedicated to fix the attribute set issue. The other one fixes several unrelated test failures. But integration tests are still failing, Ondrej is taking a look:

https://community.theforeman.org/t/discovery-integration-tests-broken-due-to-selenium/14675

lzap:
Tests are passing except the integration test. Strange thing is locally I do see the selenium error while on jenkins there is a weird:

```
NoMethodError: undefined method `[]' for nil:NilClass
    /home/jenkins/workspace/foreman_discovery-pull-request@2/database/postgresql/label/fast/ruby/2.6/plugin/test/test_helper_discovery.rb:146:in `block in assert_param'
    /home/jenkins/workspace/foreman_discovery-pull-request@2/database/postgresql/label/fast/ruby/2.6/plugin/test/test_helper_discovery.rb:145:in `each'
    /home/jenkins/workspace/foreman_discovery-pull-request@2/database/postgresql/label/fast/ruby/2.6/plugin/test/test_helper_discovery.rb:145:in `assert_param'
    /home/jenkins/workspace/foreman_discovery-pull-request@2/database/postgresql/label/fast/ruby/2.6/plugin/test/integration/discovered_hosts_test.rb:88:in `block (3 levels) in <class:DiscoveredHostsTest>' (NoMethodError)
/home/jenkins/workspace/foreman_discovery-pull-request@2/database/postgresql/label/fast/ruby/2.6/plugin/test/integration/discovered_hosts_test.rb:88    
```

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
timogoebel: I guess we should think about having a public sentry instance to collect all exceptions so we can do something about them. I'll leave it to you to decide.

lzap:
This is now ready for re-review @timogoebel tests are FIXED! Also @rahulbajaj0509 FYI.
Please keep the three separate commits when merging, I might be backporting this change. Let me know if you need some changes and I will rebase the first one.

Incorporated https://github.com/theforeman/foreman_discovery/pull/483 from @ezr-ondrej thanks a bunch!

timogoebel:
> This is now ready for re-review @timogoebel tests are FIXED! Also @rahulbajaj0509 FYI.

I won't be able to do a full review this. Week. @rahulbajaj0509: Can you do the honor?

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
lzap: I did that in the new rebase.

app/models/discovery_attribute_set.rb
lzap: I can probably keep this one, it's not causing any harm it's the other one which breaks re-discovery when attribute_set related table was not associated due to bug/problem.

app/services/foreman_discovery/import_hooks/discovery_attribute.rb
lzap: I am going to drop this from the patch, it is really not relevant and I don't want to introduce another regression.

lzap: By the way exceptions can be configured to be passed into ElasticSearch and this works today!

lzap:
I was able to figure it out. For association, `validates` cannot be used, there is a special validator called `validates_associated` which does not fail if association was not yet created. It must be used on only one side of the association tho according to docs:

* https://guides.rubyonrails.org/active_record_validations.html#validates-associated
* https://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_associated

app/models/discovery_attribute_set.rb
lzap: Replaced with `validates_associated` on the other side of the association.

lzap:
Database drop failed on postgres, but it seems to be irrelevant failure. Tests passed above.

```
+ bundle exec rake db:drop RAILS_ENV=production DISABLE_DATABASE_ENVIRONMENT_CHECK=true
rake aborted!
ActiveRecord::AdapterNotSpecified: 'production' database is not configured. Available: ["development", "test"]
```

lzap:
[test discovery]

timogoebel:
[test foreman_discovery]

ares:
now it fails on only ruby 2.3, wihle saying `94 runs, 1098 assertions, 0 failures, 0 errors, 0 skips` so I think tests can be considered green

lzap:
Guys since develop is all red and we need to unblock RH QA as they are hitting this attribute issue in automated tests, I am gonna merge this now. Thanks for help!

rabajaj0509:
@lzap sorry I was unable to test it until now because I was travelling, but doing so, I think the change is valid and on my instance, it works just fine.

lzap:
NP good luck with your talks! See you soon. Please rebase all your discovery PRs in the meantime so I can re-review them.

PR #484, opened 2019-08-01
Refs #27340 - fixed begin-rescue in a unit test

lzap:
When I was fixing tests, I have missed this. It is passing on more modern Ruby versions however not on Jenkins.


lzap:
I am merging, tests are failing now because a different core change: https://github.com/theforeman/foreman/pull/6933

PR #487, opened 2019-08-09
14.0-stable compatibility with 1.21

mmoll:


mmoll:
14.0 is still used with Foreman core 1.21, so it should get tested and behave well.

https://ci.theforeman.org/job/test_plugin_matrix/707/ against 1.21-stable looks good.

lzap:
Thanks merged.

PR #488, opened 2019-08-10
Refs #27340 - fixed begin-rescue in a unit test

mmoll:


lzap:
Yeah thanks, I think it is somewhere in some pending PR.

PR #490, opened 2019-09-25
Fixes #27935 - allow deletion of existing hosts during discovery

lzap:
When there is an existing managed host, discovery process either fail or pass depending on if the existing host has the same MAC provisioning address or just hostname. Hostnames must be unique (due to DNS) however MAC addresses do not. There was a setting to change the behavior to error out explicitly or just ignore and allow discovery of hosts with different hostname but conflicting MAC address.

There is a third solution and that is often asked "delete existing managed host when it exists". This is dangerous because it allows to delete hosts via discovery API, so this feature comes only with explicit setting. Documentation will be updated to reflect this new (dangerous) setting.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -72,12 +72,17 @@ class Host::Discovered < ::Host::Base
     Rails.logger.warn "Hostname does not start with an alphabetical character" unless hostname.downcase.match(/^[a-z]/)
 
     # check for existing managed hosts and fail or warn
-    existing_managed = Nic::Managed.joins(:host).where(:mac => bootif_mac, :provision => true, :hosts => {:type => "Host::Managed"}).limit(1)
+    existing_managed = Nic::Managed.joins(:host).where(:mac => bootif_mac, :provision => true, :hosts => {:type => "Host::Managed"}).limit(1) ||
+      Host::Managed.find_by_name(hostname).limit(1)
     if existing_managed.count > 0
-      if Setting[:discovery_error_on_existing]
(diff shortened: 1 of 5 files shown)


lzap:
@lero in my hotfix I had a typo this is a working version. Tell me if it works for you. Btw your company would be a nice reference for Foreman open-source project, can you arrange getting the link onto https://theforeman.org/who_uses.html page? We just need an approval via Google Form (link is on the page).

lero:
Cool. I'll test this one. About the company reference, I'll talk internally about it and let you know. Thank you!

timogoebel (approved):
Code looks good to me, though I haven't tested it.

lzap:
@lero does it work? Can we merge?

lero:
It does. But remember that the settings option didn't show up here and I just removed all the ifs from host/discovered.rb to force the deletion? So I'm not sure if the full patch works :)

lzap:
@timogoebel I am pretty confident about the deletion/error code, however not so sure about the setting and migration. Could you just migrate and boot up if you see the setting correctly? The migration there is to delete the old setting and to copy original value to the new one.

timogoebel:
> Could you just migrate and boot up if you see the setting correctly?

I don't see it.

```
irb(main):001:0> Setting.find_by_name('discovery_action_on_existing')
=> nil
```

timogoebel (changes requested):
We need to investigate why the new setting is not being created.

lzap:
My dev database is screwed up it seems, I need to get back to this after I start over.

stejskalleos:
I'm closing the PR since the last update was on Nov 29, 2019

PR #492, opened 2019-11-22
Fixes #28330 - reboot/kexec use old IP address

lzap:
With the RM-16143 change discovery now performs unused_ip call during provisioning. This changes IP address and creates DHCP reservation, however the discovered node still have the old lease. The fix is to use old lease IP to perform reboot or kexec API commands. There could be a race condition when a node already acquired new IP and reconfigured network, for this reason if the first request fails, there is additional second query against the reserved IP just to make sure.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -137,12 +137,12 @@ class Host::Discovered < ::Host::Base
     subnet.present? && subnet.discovery.present?
   end
 
-  def proxy_url
-    proxied? ? subnet.discovery.url + "/discovery/#{self.ip}" : "https://#{self.ip}:8443"
+  def proxy_url(node_ip)
+    proxied? ? subnet.discovery.url + "/discovery/#{node_ip}" : "https://#{node_ip}:8443"
   end
 
   def refresh_facts
-    facts = ::ForemanDiscovery::NodeAPI::Inventory.new(:url => proxy_url).facter
+    facts = ::ForemanDiscovery::NodeAPI::Inventory.new(:url => proxy_url(self.ip)).facter
     self.class.import_host facts
     import_facts facts
   rescue => e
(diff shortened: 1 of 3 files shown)


lzap:
@timogoebel can I have you to look on this one? This is 1.24 blocker.

app/models/host/discovered.rb
```diff
@@ -137,33 +137,49 @@ def proxied?
     subnet.present? && subnet.discovery.present?
   end
 
-  def proxy_url
-    proxied? ? subnet.discovery.url + "/discovery/#{self.ip}" : "https://#{self.ip}:8443"
+  def proxy_url(ip)
```
timogoebel: `ip` shadows the ip attribute of the model. Can we be more specific here?

app/models/host/discovered.rb
```diff
@@ -137,33 +137,49 @@ def proxied?
...
     self.class.import_host facts
     import_facts facts
   rescue => e
     ::Foreman::Logging.exception("Unable to get facts from proxy", e)
     raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url, :error => e)
   end
 
-  def reboot
-    resource = ::ForemanDiscovery::NodeAPI::Power.service(:url => proxy_url)
+  def reboot(ip = facts["discovery_bootip"] || facts["ipaddress"] || self.ip)
```
timogoebel: I think this would be more readable as a kwarg, something like this:
```suggestion
  def reboot(proxy_ip: facts["discovery_bootip"] || facts["ipaddress"] || self.ip)
```

lzap: Well not so much for the `kexec` method down below. I slightly lean towards keeping regular argument.

app/models/host/discovered.rb
lzap: Yeah good catch, I actually had `node_ip` originally there.

lzap:
Rebased.

timogoebel (approved):
Looks ok. I‘m wondering if iterating over an array of IPs makes this easier to read. We can then do a break if the action was successful.

Test failure is unrelated.

lzap:
It definitely looks better, rebased.

lzap:
I am giving this a shot on my prod-test instance.

lzap:
It works fine! @timogoebel thanks.

ares:
Perhaps the green button needs clicking? :-)

timogoebel (approved):

timogoebel:
Thanks!

lzap:
Yeah thanks this was tested on my instance. However it's a quite bug change for discovery users, there might be some dragons after the release. Let's see.

PR #495, opened 2020-02-28
Fixes #29221 - Subnet is blank after provisioning a discovered host

lzap:
Subnet is blindly taken from hostgroup even if its set to nil then. The expected behavior is that subnet where discovered host was detected in should carry over to provisioning (if its nil in hostgroup).

Edit: This is no longer a doco PR, @melcorr nevermind.

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -199,15 +199,17 @@ class DiscoveredHostsController < ::ApplicationController
 
   def setup_host_class_variables
     if @host.hostgroup
+      subnet = @host.hostgroup.subnet || @host.subnet
+      subnet6 = @host.hostgroup.subnet6 || @host.subnet6
       @architecture    = @host.hostgroup.architecture
       @operatingsystem = @host.hostgroup.operatingsystem
       @environment     = @host.hostgroup.environment
       @domain          = @host.hostgroup.domain
-      @subnet          = @host.hostgroup.subnet
+      @subnet          = subnet
+      @subnet6         = subnet6
       @compute_profile = @host.hostgroup.compute_profile
       @realm           = @host.hostgroup.realm
-      @host.interfaces.first.assign_attributes(subnet: @subnet,
-                                               domain: @domain)
+      @host.interfaces.first.assign_attributes(subnet: subnet, subnet6: subnet6, domain: @domain)
       @host.environment = @environment
     end
   end
--- a/extra/discover-host
+++ b/extra/discover-host
@@ -33,7 +33,7 @@ OptionParser.new do |opts|
     version = v
   end
 
-  opts.on("-iARRAY", "--interface=ARRAY", Array, "Comma separated array: name,subnet,ip,mac (can be used multiple times, default: eth0,122,7,AA:BB:CC:DD:EE:FF") do |v|
+  opts.on("-iARRAY", "--interface=ARRAY", Array, "Comma separated array: name,network/prefix,mac,ip (can be used multiple times, default: eth0,192.168.122.0/24") do |v|
     interfaces << v
   end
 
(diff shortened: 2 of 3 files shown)


melcorr:
You pushed while I was reviewing it @lzap - no worries!

lzap:
> You pushed while I was reviewing it @lzap - no worries!

Yeah because I've changed my mind, it was surprisingly easy to fix this.

lzap:
@aruzicka I consider this an important one to get through :-)

lzap:
Was wondering if we can squeeze this into the next release for 2.1 @adamruzicka.

extra/discover-host
```diff
@@ -76,15 +76,21 @@ json["discovery_version"] = version
...
+    else
+      random_ip = IPAddr.new(subnet, Socket::AF_INET6) | IPAddr.new(rand(2**prefix.to_i),Socket::AF_INET6)
+      fact_name = "ipaddress6"
+    end
+    json["#{fact_name}_#{name}"] = explicit_ip ? explicit_ip : random_ip
     if name == primary
-      json["macaddress"] = mac
-      json["ipaddress"] = ip
+      json["macaddress"] = json["macaddress_#{name}"]
+      json["#{fact_name}"] = json["#{fact_name}_#{name}"]
```
adamruzicka:
```suggestion
      json[fact_name] = json["#{fact_name}_#{name}"]
```

extra/discover-host
```diff
@@ -76,15 +76,21 @@ json["discovery_version"] = version
...
-    ipo ||= rand(1..253)
-    ip = "192.168.#{subnet}.#{ipo}"
     json["macaddress_#{name}"] = mac
-    json["ipaddress_#{name}"] = ip
+    subnet_woprefix, prefix = subnet.split('/')
+    if subnet.include?('.')
+      random_ip = IPAddr.new(subnet, Socket::AF_INET) | IPAddr.new(rand(2**prefix.to_i),Socket::AF_INET)
+      fact_name = "ipaddress"
+    else
+      random_ip = IPAddr.new(subnet, Socket::AF_INET6) | IPAddr.new(rand(2**prefix.to_i),Socket::AF_INET6)
```
adamruzicka: Nitpick `s/,S/, S/`

extra/discover-host
```diff
@@ -76,15 +76,21 @@ json["discovery_version"] = version
...
+      fact_name = "ipaddress6"
+    end
+    json["#{fact_name}_#{name}"] = explicit_ip ? explicit_ip : random_ip
```
adamruzicka, line 90: How about generating a random ip on demand when it is actually needed?

adamruzicka (commented):
Left a couple of comments inline. This seems like a reasonable fix to have in 2.1

extra/discover-host
lzap: What's that?

lzap: Oh I get that.

extra/discover-host
lzap: Good idea, but this is a helper script we only use when generating fake hosts during development. Do you insist? :-)

lzap:
I have rebased and incorporated your comments. I have already branched off 16.1 unfortunately, nvm keep reviewing.

extra/discover-host
adamruzicka: Oh, I missed it is a development only thing, guess I can let it slide then :)

app/controllers/discovered_hosts_controller.rb
```diff
@@ -199,15 +199,17 @@ def submit_multiple_auto_provision
 
   def setup_host_class_variables
     if @host.hostgroup
+      subnet = @host.hostgroup.subnet ? @host.hostgroup.subnet : @host.subnet
```
adamruzicka: Couldn't we just do `@host.hostgroup.subnet || @host.subnet`?

timogoebel: Yep, and we can use @subnet below and do not need the helper vars. That makes this a little cleaner.

lzap: Rebasing, I am leaving the helper vars as they are used down below.

lzap:
Rebased.

timogoebel:
Tests are failing (though the failures look unrelated).

lzap:
Pushed, develop is green now.

adamruzicka (approved):
LGTM

timogoebel:
Thanks, @lzap.

PR #496, opened 2020-03-06
Refs #16143 - fix unit test for subnet

lzap:
A copy and paste error from the test above. Minitest 5.x prints a warning in case `nil` is passed into `assert_equal` which I blindly ignored for some time now and now it bites for some reason. Possibly a bug in Minitest itself: https://github.com/seattlerb/minitest/issues/837

Anyway, this fixes the root cause - `assert_nil` is now properly used. These were not expected to be `nil` at all.

### DIFF
--- a/test/unit/discovered_extensions_test.rb
+++ b/test/unit/discovered_extensions_test.rb
@@ -204,6 +204,8 @@ class DiscoveredExtensionsTest < ActiveSupport::TestCase
     assert_empty host.errors
     assert managed_host
     assert_empty managed_host.errors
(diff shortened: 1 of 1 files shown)


mmoll:
I guess this is still valid?

mmoll:
Fixes the following at least locally for me:
```
DiscoveredExtensionsTest#test_0018_subnet6 is changed and unused_ip called:
NoMethodError: undefined method `split' for nil:NilClass
```

mmoll (approved):

PR #499, opened 2020-04-29
Fixes #29658 - discovery_rule hosts association on Rails 6

ezr-ondrej:
@lzap as of https://community.theforeman.org/t/activerecord-woe-implicit-order-and-association/18492 this solves it, but there might be other places. I would love if you could think of getting rid of `Host` factory in favor of `Host` model from `Host::Discovered` perspective, as it could be only blocker.


lzap (approved):
This fixed the issue thank you :-)

PR #505, opened 2020-08-27
Fixes #30726 - hostname prefix setting should not be nil

rabajaj0509:


rabajaj0509:
This pr depends upon: https://github.com/theforeman/foreman/pull/7934

app/models/setting/discovered.rb
```diff
@@ -8,7 +8,9 @@ class Setting::Discovered < ::Setting
   BLANK_ATTRS << 'discovery_facts_hardware'
   BLANK_ATTRS << 'discovery_facts_network'
   BLANK_ATTRS << 'discovery_facts_ipmi'
-  BLANK_ATTRS << 'discovery_prefix'
+
+  HOSTNAME_ATTRS = ['discovery_hostname', 'discovery_prefix']
```
ekohl: Are these really attrs that contain a hostname? prefix suggests otherwise. Perhaps `STRING_PRESENCE_ATTRS` is closer to what it does?

ekohl:
> This pr depends upon: [theforeman/foreman#7934](https://github.com/theforeman/foreman/pull/7934)

This is no longer true. Code was rewritten to not depend on a core change.

app/models/setting/discovered.rb
rabajaj0509: Done!

rabajaj0509:
I think there should be a migration for this issue because things might break for users that have set either of these settings to nil. Will update the PR with a migration and fix the test cases.

test/unit/host_discovered_test.rb
```diff
@@ -260,12 +260,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
     assert_equal 'teste41f13cc3658', host.name
   end
 
-  test "should create discovered host without prefix" do
```
rabajaj0509, line 263: I have removed this test cases all together because there is no option of having the `Setting[:discovery_prefix] = ''` in the settings, didn't make any sense of having them. Let me know your thoughts.

lzap: Yeah it is dangerous not to have it because hostname must NOT start with a number. So actually this test was not good at all.

lzap:
Thanks this works for me.

test/unit/host_discovered_test.rb
ekohl: > Yeah it is dangerous not to have it because hostname must NOT start with a number. So actually this test was not good at all.

Since [RFC 1123](https://tools.ietf.org/html/rfc1123#section-2) hostnames can start with a digit. https://tools.ietf.org/html/rfc3696#section-2 does state that the TLD shouldn't be numberic so the FQDN can't be all numeric.

test/functional/api/v2/settings_controller_test.rb
```diff
@@ -6,10 +6,10 @@ class Api::V2::SettingsControllerTest < ActionController::TestCase
   end
 
   test_attributes :pid => '2c5ecb7e-87bc-4980-9620-7ae00e3f360e'
-  test "should update hostname prefix without value" do
+  test "should not update hostname prefix without value" do
     setting = Setting.find_by_name("discovery_prefix")
     put :update, params: { :id => setting.id, :setting => { :value => '' } }
-    assert_equal JSON.parse(@response.body)['value'], '', "Can't update discovery_prefix setting with empty value"
+    assert_nil JSON.parse(@response.body)['value']
```
ekohl, line 12: Should this test also verify the status code is some 4xx code?

PR #506, opened 2020-09-11
Fixes #30478 - try again when IP is invalid

lzap:
The fix to use old lease when rebooting was not working correctly. It was not implemented for kexec completely, also it was not passing IP correctly. Also it could timeout on client. This fixes it all.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -146,9 +146,11 @@ class Host::Discovered < ::Host::Base
     raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url, :error => e)
   end
 
-  def reboot(node_ip = nil)
+  def reboot(old_ip = nil, new_ip = nil)
     # perform the action against the original lease as well as the new reservation
-    [node_ip || facts["discovery_bootip"] || facts["ipaddress"], self.ip].compact.each do |next_ip|
+    ips = [old_ip, new_ip, self.ip].compact.uniq
+    logger.debug "Performing reboot calls against #{ips.to_sentence}, facts left #{facts.count}"
+    ips.each do |next_ip|
       begin
         node_url = proxy_url(next_ip)
         logger.debug "Performing reboot call against #{node_url}"
@@ -157,17 +159,17 @@ class Host::Discovered < ::Host::Base
       rescue => e
         msg = N_("Unable to perform reboot on %{name} (%{url}): %{msg}")
         ::Foreman::Logging.exception(msg % { :name => name, :url => node_url, :msg => e.to_s }, e)
-        if next_ip == self.ip
-          raise ::Foreman::WrappedException.new(e, msg, :name => name, :url => node_url, :msg => e.to_s)
-        end
       end
     end
-    false
+    msg = N_("Unable to perform %{action} on %{ips}")
+    raise ::Foreman::Exception.new(msg, action: "reboot", ips: ips.to_sentence)
   end
 
-  def kexec(json, node_ip = nil)
+  def kexec(json, old_ip = nil, new_ip = nil)
     # perform the action against the original lease as well as the new reservation
-    [node_ip || facts["discovery_bootip"] || facts["ipaddress"], self.ip].compact.each do |next_ip|
+    ips = [old_ip, new_ip, self.ip].compact.uniq
+    logger.debug "Performing kexec calls against #{ips.to_sentence}, #{facts.count} facts left"
+    ips.each do |next_ip|
       begin
         node_url = proxy_url(next_ip)
         logger.debug "Performing kexec call against #{node_url}"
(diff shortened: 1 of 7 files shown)


lzap:
Hey @adamruzicka or @timogoebel can you take a look? This is tested on real instance, I also tested timeouts by reversing the IP array order.

lzap:
Need to test few unit tests tho.

app/models/host/managed_extensions.rb
```diff
@@ -27,7 +27,7 @@ def queue_reboot
   end
 
   def setReboot
-    old.becomes(Host::Discovered).reboot(facts["discovery_bootip"] || facts["ipaddress"])
+    old.becomes(Host::Discovered).reboot(old.ip, ip)
```
lzap, line 30: Root cause was that `old` and `self` are only available for host (this class is included into host). Not later.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -51,7 +60,6 @@ def self.unused_ip_for_host(host, new_subnet = nil, new_subnet6 = nil)
       interface.subnet6 = new_subnet6 if new_subnet6
       interface.ip = unused_ip_for_subnet(interface.subnet, interface.mac, interface.ip) if interface.subnet
       interface.ip6 = unused_ip_for_subnet(interface.subnet6, interface.mac, interface.ip6) if interface.subnet6
-      interface.save!
```
lzap, line 54: This calls save twice, interfaces are actually saved when host is saved. At least according to my testing, this works both for quick provision and normal edit host form.

app/models/host/discovered.rb
```diff
@@ -146,9 +146,11 @@ def refresh_facts
     raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url, :error => e)
   end
 
-  def reboot(node_ip = nil)
+  def reboot(old_ip = nil, new_ip = nil)
```
timogoebel, line 149:
```suggestion
  def reboot(old_ip: nil, new_ip: nil)
```

Might make the API a little bit better to read, WDYT?

app/models/host/discovered.rb
```diff
@@ -146,9 +146,11 @@ def refresh_facts
...
     # perform the action against the original lease as well as the new reservation
-    [node_ip || facts["discovery_bootip"] || facts["ipaddress"], self.ip].compact.each do |next_ip|
+    ips = [old_ip, new_ip, self.ip].compact.uniq
+    logger.debug "Performing reboot calls against #{ips.inspect}, facts left #{facts.count}"
```
timogoebel:
```suggestion
    logger.debug "Performing reboot calls against #{ips.to_sentence},  #{facts.count} facts left"
```

app/models/host/discovered.rb
```diff
@@ -157,17 +159,17 @@ def reboot(node_ip = nil)
       rescue => e
         msg = N_("Unable to perform reboot on %{name} (%{url}): %{msg}")
         ::Foreman::Logging.exception(msg % { :name => name, :url => node_url, :msg => e.to_s }, e)
-        if next_ip == self.ip
-          raise ::Foreman::WrappedException.new(e, msg, :name => name, :url => node_url, :msg => e.to_s)
-        end
       end
     end
+    logger.warn "Could not contact discovered host on any of the IPs: #{ips.inspect}"
```
timogoebel:
```suggestion
    logger.warn "Could not contact discovered host on any of the IPs: #{ips.to_sentence}"
```

app/services/foreman_discovery/host_converter.rb
timogoebel: Did we by any chance add this for the bond creation to work? I don't remember, need to check.

rplevka:
I tested the commit on downstream satellite and the kexec workflow seems to work correctly now.
i tested both scenarios, where the discovered host still uses the reported IP as well as a special case, where its lease times out and it picks up the new ip assigned by DHCP during the orchestration.

Both work fine. Perhaps I'd add a simple unit test to test the fallback mechanism.

app/services/foreman_discovery/host_converter.rb
lzap: Bond creation?

timogoebel: I meant this: https://github.com/theforeman/foreman_discovery/blob/54f1724d8965dfd45eea3c0531df8ed95c509eea/app/services/foreman_discovery/import_hooks/lldp_neighbor.rb

But scratch that, it's unrelated.

lzap:
Amended mentioned changes.

lzap:
Oh I had to put the method arguments back, it broke actually other parts and tests and I don't want to tackle this in this PR. I also prefer C-like style to Ruby-like style.

timogoebel (approved):

timogoebel:
@rplevka: Do you want to give this another shot?

rplevka:
@timogoebel @lzap 
I just successfully retested it.

- kexec works
- fallback (new ip) kexec works

LGTM

lzap:
Thanks merged.

PR #508, opened 2020-09-30
Fixes #30945: Add create.json.rabl template for discovery rules

ehelms:
The lack of a template for create results in a different output
format being used from the show response. Further, without this
template the rendering engine will perform a to_json and skip
the Rabl layer all together.


lzap:
Can you rebase on top of develop? I have fixed the broken UI tests.

lzap:
Thanks a bunch! Great analysis on RoR. :-)

PR #510, opened 2020-10-02
Fixes #30958 - fix argument error on refresh facts

lzap:
An overlook in RM-30478 causes incorrect error message due to missing parameter.


timogoebel (approved):

timogoebel (approved):

timogoebel:
Thanks!

PR #513, opened 2020-10-27
Fixes #31006 - set taxonomy from subnet6 too

lzap:
Some users report that discovered hosts don't get token generated during provisioning. This comes back regularly, discovery relies on validation callback in core to create new token, let's simply call this explicitly when converting the host to ensure it exists.


lzap:
Sorry messed up my git and forcepushed. Fixed.

PR #514, opened 2020-10-27
Fixes #31183 - Enforce new token during discovery

lzap:
Some users report that discovered hosts don't get token generated during provisioning. This comes back regularly, discovery relies on validation callback in core to create new token, let's simply call this explicitly when converting the host to ensure it exists.


stejskalleos:
I'm closing the PR since the last update was on Oct 27, 2020

PR #515, opened 2020-11-03
Fixes #31242 - update the engine.rb to use latest foreman

rabajaj0509:
I think the empty state component does not display in the nightly because we are using an older version of Foreman where the EmptyState component is not updated. 

Although, I am  not sure if a version upgrade to gem will be required or not, please let me know your thoughts on this @lzap

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -43,7 +43,7 @@ module ForemanDiscovery
 
     initializer 'foreman_discovery.register_plugin', :before => :finisher_hook do |app|
       Foreman::Plugin.register :foreman_discovery do
-        requires_foreman '>= 2.2'
+        requires_foreman '>= 2.3'
 
         # discovered hosts permissions
         security_block :discovery do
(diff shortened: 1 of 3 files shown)


lzap:
Sorry I forgot this during my release engineering process. Due to 2.2 being so late I checked our frontpage and thought - ok the next version is 2.2 then :-)

lzap (approved):
Indeed. Merging.

lzap:
Now, this was an overlook and I can release new version of the gem, but I doubt this will fix the problem with javascript you showed me. If you install nightly you always install latest and greatest Foreman + Discovery. So although these version requirements were incorrect, you were actually using the correct version. Let me know what to do next to fix the problem.

rabajaj0509:
While testing I realised that the Foreman instance I was using was the latest. You are correct, we have not yet resolved our main problem. Let me take a look at it and get back to you. @lzap 

@laviro can you think of any dependency that we are missing?

Ron-Lavi:
@rabajaj0509 Where can I see that error?
do I need to install nightly?

PR #516, opened 2020-11-09
Fixes #31278 - include `lib` directory in production

Ron-Lavi:


Ron-Lavi:
@lzap @rabajaj0509 can you take a look please?

in nightly I got the following error: `uninitialized constant ForemanDiscovery::VERSION`
which we are trying to pass into the EmptyState component in https://github.com/theforeman/foreman_discovery/blob/a70abce45cf85fc93bc46668fb286d83275da2c5/app/views/discovered_hosts/welcome.html.erb#L15

I added the `lib` dir to be loaded with `eager_load`, and now the component is visible and works as expected

rabajaj0509 (approved):

rabajaj0509:
@laviro waiting for tests to pass, once they pass, I shall merge this PR :) good job here buddy!

lzap:
I don't understand the failure `ActiveRecord::AdapterNotSpecified: The `production` database is not configured for the `production` environment.` That is probably not related?

Can you maybe give little bit more context about the change? Doesn't this break auto-loading in that directory? I can't comment much as Rails non-expert about this one. Maybe @ezr-ondrej or @tbrisker can help to verify if `eager_load` like that is correct?

lzap:
[test discovery]

Ron-Lavi:
@lzap maybe this article will put some light on the issue: https://pathbox.github.io/2016/03/22/autoload_eager_autoload_path/#if-you-add-code-in-your-lib-directory

Ron-Lavi:
sorry for the last ugly push, didn't realize discovery has both `master` and `develop` branches :D

lzap:
Whoops, okay. Still running...

lzap:
Again the same error, is this something expected, maybe 2.7 Ruby is too new on Jenkins? @ekohl or @tbrisker ? I scroll all the output and I could not find any other error just the database production one from above.

lib/foreman_discovery/engine.rb
```diff
@@ -15,6 +15,8 @@ class Engine < ::Rails::Engine
     config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
     config.autoload_paths += Dir["#{config.root}/app/services"]
 
+    config.eager_load_paths += Dir["#{config.root}/lib"]
```
ekohl, line 18: This feels very weird. This file itself is in `lib` so isn't it already loaded at this point?

ekohl (changes requested):
> in nightly I got the following error: `uninitialized constant ForemanDiscovery::VERSION`

I think the Foreman plugin already has a version via Gems after you call `register`.

https://github.com/theforeman/foreman/blob/a600617c1ce75582fdb61a7dca5d549fc6a01cca/app/registries/foreman/plugin.rb#L102

That means it'd be better to avoid the constant and get the plugin instance. Then you can call version on it:

```ruby
Foreman::Plugin.find(:foreman_discovery).version
```

Untested, but I believe that to a cleaner fix.

Also, it may break branding if you do it this way. A better way would be to use the [LinksController](https://github.com/theforeman/foreman/blob/develop/app/controllers/links_controller.rb). However, it looks like it's currently not possible to pass the version (https://github.com/theforeman/foreman/blob/a600617c1ce75582fdb61a7dca5d549fc6a01cca/app/controllers/links_controller.rb#L13-L14) even though its implementation supports this (https://github.com/theforeman/foreman/blob/a600617c1ce75582fdb61a7dca5d549fc6a01cca/app/controllers/links_controller.rb#L37-L41). You should submit a PR to support that and then use it.

> Again the same error, is this something expected, maybe 2.7 Ruby is too new on Jenkins? @ekohl or @tbrisker ? I scroll all the output and I could not find any other error just the database production one from above.

I wouldn't expect the Ruby 2.7 version to time out after 2 hours. Perhaps it the loading is entering some sort of loop?

lib/foreman_discovery/engine.rb
Ron-Lavi: I am not very familiar with how foreman loads the plugins, maybe it is looking for a specific engine file? And the rest of the lib is ignored?

I will try tomorrow the solution you suggested in nightly, meanwhile adding @ShimShtein to the party :)

Ron-Lavi:
Doing `Foreman::Plugin.find(:foreman_discovery).version` made that page to work, thanks @ekohl !!
I will open another PR

lzap:
Thanks for help with this one @ekohl ! I am gonna close this one, the other one looks much cleaner.

PR #519, opened 2020-12-01
Fixes #31203 - welcome page for discovery rules

rabajaj0509:
Tasks involved in this effort are as follows:

* [x]  Adding button for the `Create Discovery Rule`
* [x]  Suitable icon for Discovery Rules
* [x]  Suitable description for Discovery Rules
* [x]  Add test cases for welcome page

RulesPage

Apart from this, what do you think @laviro?

### DIFF
--- a/app/controllers/discovery_rules_controller.rb
+++ b/app/controllers/discovery_rules_controller.rb
@@ -6,6 +6,10 @@ class DiscoveryRulesController < ApplicationController
 
   before_action :find_resource, :only => [:edit, :update, :destroy, :enable, :disable, :auto_provision]
 
+  def model_of_controller
+    Host::Discovered
+  end
+
   def index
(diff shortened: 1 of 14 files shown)


rabajaj0509:
We are still using the server to display the welcome page, soon we will use the react router to perform the routes and then we will not use the `welcome` method and the rails magic :stuck_out_tongue: It will make the loading of the pages very quick.

@laviro can you take a look at this PR? @lzap on your local we can compile the webpack in the test environemnt, something like `bundle exec rake webpack:compile  RAILS_ENV=test` and it should work fine :smile:

lzap (approved):
It works on my instance. @laviro to review the code thanks.

Ron-Lavi (approved):
Code wise looks good to me, didn't test it out yet

lzap:
Thanks all! Merged.

PR #521, opened 2021-01-13
Fixes #31622 - add react router for quick page response

rabajaj0509:


rabajaj0509:
As discussed with @laviro, we would like to use the react-table from the foreman core. I don't think I will be able to work on this any more due to a shift of duties. Please take over this PR @laviro :)

PR #524, opened 2021-01-28
fixes #18932 - Introducing Discovery kind of facts

domitea:
Fixing bug in Discovery where all facts from dicovered hosts are recognised as puppet facts. So this PR solved this problem by creating Discovery kind of facts for non-Puppet environments.

The code problem was in `FactImporterRegistry` where is defined only one importer `PuppetFactImporter` and this importer as marker as default. That means that this importer is used every time when there no importer that fits better for given fact. Because Discovery doesn't have any own importer, every fact from Discovery is taken as puppet fact. Thanks to that the `FactName` is type of 'PuppetFactName' and Puppet fact is created and recognized even there no puppet installed at all.

So to avoid this, I created a `DiscoveryFactImporter` with own `DiscoveryFactName`. This will keep facts of discovered hosts as discovery type. However, this PR is targeted to non-puppet environments so I'm not sure if it's work properly with puppet. 

Also, there is missing a piece that could transform a puppet like facts from discovery into discovery like facts. And because it's change of type of `FactName`, the migration will be easy to write. However, only for non-puppet environments. So it will be better to do migration for that, or a task?

### DIFF
--- a/app/models/discovery_fact_name.rb
+++ b/app/models/discovery_fact_name.rb
@@ -1,0 +1,9 @@
+class DiscoveryFactName < FactName
+  def origin
+    'Discovery'
+  end
+
+  def icon_path
+    "icons16x16/stub/darkred-d"
+  end
+end
--- a/app/services/discovery_fact_importer.rb
+++ b/app/services/discovery_fact_importer.rb
@@ -1,0 +1,5 @@
+class DiscoveryFactImporter < StructuredFactImporter
+  def fact_name_class
+    DiscoveryFactName
+  end
+end
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -21,10 +21,10 @@ class ForemanDiscovery::HostConverter
   def self.set_build_clean_facts(host)
     # fact cleaning
     if Setting['discovery_clean_facts']
-      # clean all facts except those starting with "discovery_"
+      # clean all facts except those from Discovery
       host.define_singleton_method(:clear_facts) do
-          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.name like 'discovery_%'").pluck("fact_values.id")
-          FactValue.where("host_id = #{host.id} and id not in (?)", keep_ids).delete_all
+        keep_ids = FactValue.where(host_id: host.id, fact_names: { type: 'DiscoveryFactName' }).where("fact_names.name like 'discovery_%'").joins(:fact_name).pluck("fact_values.id")
+        FactValue.where.not(id: keep_ids).delete_all
       end
     else
       # clean no facts (default behavior)
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -216,6 +216,8 @@ module ForemanDiscovery
       # Controller extensions
       ::HostsController.send :include, ForemanDiscovery::Concerns::HostsControllerExtensions
       ::Api::V2::FactValuesController.send :include, Api::V2::FactValuesControllerExtensions
+
+      Foreman::Plugin.fact_importer_registry.register(:foreman_discovery, DiscoveryFactImporter)
     end
 
     rake_tasks do


app/models/discovery_fact_name.rb
```diff
@@ -0,0 +1,9 @@
+class DiscoveryFactName < FactName
+  def origin
+    'Discovery'
+  end
+
+  def icon_path
+    "icons16x16/stub/darkred-d"
```
domitea, line 7: For recognition purposes I used a icon used for Darwin based operating system recognition. I'm not sure if it's good to stay with that or use/create another one.

ares: I think that's perfectly ok, we don't have a specific discovery icon

app/services/discovery_fact_importer.rb
```diff
@@ -0,0 +1,9 @@
+class DiscoveryFactImporter < StructuredFactImporter
+  def self.authorized_smart_proxy_features
+    'Discovery'.freeze
```
ares: Will this work or is this even necessary? The proxy running on FDI is typically not registered in Foreman, but I have no idea how authorization works in this case of facts upload. Perhaps we're skipping it entirely?

domitea: I just copied a `PuppetFactImporter`, so that's the reason, why this method has been created. I tried that without this method and `DiscoveryFactImporter` working properly. So I deleted this method and updated PR.

domitea:
@ares Tests are not failing, so I think it's ready for merge.

app/models/discovery_fact_name.rb
rabajaj0509: Although, can we use the `gear` icon, just to keep it consistent with the welcome page?

rabajaj0509: Although, can we use the gear icon, just to keep it consistent with the welcome page?

domitea: Well, I would like to use a some icon with D in the center (like Puppet one). I'm not sure what kind of icon do you think but I really don't care what kind icon it will be. I only want to have a D in there because it signals that fact came from Discovery.

What do you think? @rabajaj0509 @ares

lzap: I am okay with D or Foreman icon. Not sure about gears, that looks weird compared to other facts on a single page.

lzap (changes requested):
It works if you start with a clean FactName table. But once you already have fact names with type "PuppetFactName" and apply this patch, after new host uploads facts I am unable to fetch them from the database for some reason:

http://localhost:3000/fact_values?search=name+%3D+architecture&page=1

This no longer shows any results, there must be some (default) scope that limits it.

lzap:
Please test this scenario, maybe add a test:

* Add a FactName with PuppetFactName type
* Import a host via discovery
* Search facts

We really must not break this.

domitea:
This bad behavior can be archived with the `develop` branch too. So I think that my PR doesn't broke that. Yes, It should be fixed but It's out of scope of this PR.

lzap:
Okay, I would like to do one more change. This is a great opportunity to update a behavior - method `set_build_clean_facts` is overriden to prevent deletion of facts which starts with `discovery_`. Can you modify the behaviour so it actually prevents deleting of all facts stored as Discovery origin.

Also fix the tests and include a case when a discovered fact is named without the prefix.

domitea:
Checking if only one test is failed...

app/services/foreman_discovery/host_converter.rb
```diff
@@ -21,9 +21,9 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
   def self.set_build_clean_facts(host)
     # fact cleaning
     if Setting['discovery_clean_facts']
-      # clean all facts except those starting with "discovery_"
+      # clean all facts except those from Discovery
       host.define_singleton_method(:clear_facts) do
-          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.name like 'discovery_%'").pluck("fact_values.id")
+          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.type like 'DiscoveryFactName'").pluck("fact_values.id")
           FactValue.where("host_id = #{host.id} and id not in (?)", keep_ids).delete_all
```
timogoebel: This could be refactored to use pure rails notation.
```ruby
FactValue.where.not(id: keep_ids)
```
Even better would be to use subqueries.

domitea: Thanks for the idea! Select is refactored.

domitea:
@lzap I've updated the tests. However, since this PR we're not rely on the name of fact, the fact can have any name, right?

app/services/foreman_discovery/host_converter.rb
```diff
@@ -21,10 +21,9 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
   def self.set_build_clean_facts(host)
     # fact cleaning
     if Setting['discovery_clean_facts']
-      # clean all facts except those starting with "discovery_"
+      # clean all facts except those from Discovery
       host.define_singleton_method(:clear_facts) do
-          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.name like 'discovery_%'").pluck("fact_values.id")
-          FactValue.where("host_id = #{host.id} and id not in (?)", keep_ids).delete_all
+          FactValue.where.not(host_id: FactValue.where(host_id: host.id, fact_name: { type: 'DiscoveryFactName' }).joins(:fact_name).pluck(:id)).delete_all
```
timogoebel: You should be able to omit the pluck here, see

https://github.com/rubocop-hq/rubocop-rails/issues/246

rabajaj0509: Nit: Since there are so many methods involved, it will be difficult to debug in the future if anything went wrong. Lets break it to getting the `id` in one and then deleting it in the next. What do you think @domitea?

domitea: Thanks @timogoebel for tip what to omit.

I understand, that huge one liner can be difficult to debug @rabajaj0509. So I split that into two lines.

rabajaj0509:
Tests are still failing @domitea

rabajaj0509 (approved):
LGTM! :+1: 

Let's wait for @lzap's final approval :)

ares:
@lzap any more comments?

test/unit/host_discovered_test.rb
```diff
@@ -321,21 +321,6 @@ class HostDiscoveredTest < ActiveSupport::TestCase
...
-      'discovery_keep_me' => "content",
-      })
-    host = discover_host_from_facts(raw)
-    host.save
-    managed = ::ForemanDiscovery::HostConverter.to_managed(host)
-    managed.clear_facts
-    assert_nil managed.facts_hash['delete_me']
-    assert_equal "content", managed.facts_hash['discovery_keep_me']
-  end
-
```
lzap: Please do not delete this test but rewrite it so the behavior is actually tested. What you want to do is to create some facts of a different type manually (create `FactValue` and `FactName` via FactoryBot) and test if these were actually removed.

lzap (changes requested):
Sorry for the delay, the code is okay, however we should not remove a test for the deletion.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -21,10 +21,10 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
...
-      # clean all facts except those starting with "discovery_"
+      # clean all facts except those from Discovery
       host.define_singleton_method(:clear_facts) do
-          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.name like 'discovery_%'").pluck("fact_values.id")
-          FactValue.where("host_id = #{host.id} and id not in (?)", keep_ids).delete_all
+        keep_ids = FactValue.where(host_id: host.id, fact_names: { type: 'DiscoveryFactName' }).joins(:fact_name)
+        FactValue.where.not(id: keep_ids).delete_all
       end
     else
       # clean no facts (default behavior)
```
lzap, line 30: Apologies I misled you here, actually purpose of discovery_clean_facts is to clean facts starting with discovery_ which are some facts which can be useful during provisioning. Fact type is actually irrelevant, can you put this hunk back as it was? Of course the code must search DiscoveryFactName but also starting with discovery_.

Sorry my bad, I told you to do this change but now I see it is incorrect.

domitea:
I've updated a PR with your comment, @lzap. I hope I merged that functionality well...

app/services/foreman_discovery/host_converter.rb
lzap: Nono, what I mean is: delete all facts which are type of DiscoveryFactName AND they start with "discovery_" prefix.

lzap (changes requested):
I am not helping much, am I? :)

domitea:
@lzap I will took the the quote from The Mandalorian: "This is the way" and I only add ", the hard way" :D

I hope that the deleting of facts is finally correct.

app/services/foreman_discovery/host_converter.rb
```diff
@@ -21,10 +21,10 @@ def self.to_managed(original_host, set_managed = true, set_build = true, added_a
   def self.set_build_clean_facts(host)
     # fact cleaning
     if Setting['discovery_clean_facts']
-      # clean all facts except those starting with "discovery_"
+      # clean all facts except those from Discovery
       host.define_singleton_method(:clear_facts) do
-          keep_ids = FactValue.where("host_id = #{host.id}").joins(:fact_name).where("fact_names.name like 'discovery_%'").pluck("fact_values.id")
-          FactValue.where("host_id = #{host.id} and id not in (?)", keep_ids).delete_all
+        keep_ids = FactValue.where(host_id: host.id, fact_names: { type: 'DiscoveryFactName' }).where("fact_names.name like 'discovery_%'").joins(:fact_name)
+        FactValue.where.not(id: keep_ids).delete_all
```
lzap, line 27: Have you removed the `pluck` call on purpose? It is way better in this case (faster, less memory used).

test/unit/host_discovered_test.rb
```diff
@@ -324,15 +324,15 @@ class HostDiscoveredTest < ActiveSupport::TestCase
...
+    fact_name = FactoryBot.create(:fact_name, name: 'PuppetFact', type: 'PuppetFactName')
+    FactoryBot.create(:fact_value, fact_name: fact_name, value: 'delete_me', host: host)
     host.save
     managed = ::ForemanDiscovery::HostConverter.to_managed(host)
+    assert_equal host.facts_hash['PuppetFact'], 'delete_me'
     managed.clear_facts
-    assert_nil managed.facts_hash['delete_me']
+    assert_nil host.facts_hash['PuppetFact']
     assert_equal "content", managed.facts_hash['discovery_keep_me']
   end
```
lzap: Can you revert this test completely to the original state? If I am not mistaken, no change should be required in this case so it should be passing. This way we will be sure we are not changing any behavior.

lzap (changes requested):
Kindof :-)

lzap (approved):
It works fine, can you do Upgrade note:

Discovered hosts have a new kind of fact (discovery) so facts will no longer appear on the Monitor - Facts page. To search for discovery hosts, use the search box on the Hosts - Discovered page.

We usually do that in theforeman.org github repository, example: https://github.com/theforeman/theforeman.org/pull/1813

Thanks!

PR #525, opened 2021-02-03
fixes #19404 - Introducing notification about failed discovery

domitea:
Today the user has no way how to be informed about failed discovery of new hosts in Foreman. This PR tries to solve that by creating notification about failed discovery and also what error occurs.

### DIFF
--- a/app/controllers/api/v2/discovered_hosts_controller.rb
+++ b/app/controllers/api/v2/discovered_hosts_controller.rb
@@ -122,6 +122,7 @@ module Api
         end
         process_response state
       rescue Exception => e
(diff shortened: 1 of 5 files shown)


lzap (approved):
Works fine. One nitpick: If a discovered host fails with message X then all failures with message Y are ignored. I don't think that solution to that is doable with the current notifications design.

lzap:
You're missing a test.

ares:
Just to clarify: the test should probably just try the endpoint with invalid facts (e.g. empty set) and ensuring new notification instance was added, we don't need to check the notification stack functionality.

domitea:
@lzap @ares Test added.

lzap (approved):
Thank you for your contribution!

PR #526, opened 2021-02-10
Add conditions to allow DHCP or Static IP during Kexec

cschlue:
Previously was forcing static provisioning template, now if host boots with DHCP the proper template should be provided.

### DIFF
--- a/app/views/foreman_discovery/debian_kexec.erb
+++ b/app/views/foreman_discovery/debian_kexec.erb
@@ -12,12 +12,15 @@ environments. The template must generate JSON format with the following items
 "kernel", "initram", "append" and "extra". The kexec command is composed in
 the following way:
 
-kexec --force --debug --append=$append --initrd=$initram $extra $kernel
+kexec --force --reset-vga --append=$append --initrd=$initram $extra $kernel
 
 Please read kexec(8) man page for more information about semantics.
-Extra options like --reset-vga can be set via "extra" array.
 -%>
 <%
+  iface = @host.provision_interface
+  subnet4 = iface.subnet
+  subnet6 = iface.subnet6
+ 
   mac = @host.facts['discovery_bootif']
   bootif = '00-' + mac.gsub(':', '-') if mac
   ip_cidr = @host.facts['discovery_ip_cidr']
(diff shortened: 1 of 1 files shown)


lzap:
Hello, thanks for the patch.

For the kexec workflow I wrongly assumed there is no DHCP, because why would you not configure PXE when there is a DHCP service. But I understand that there might be environments just like that.

Now, we can't simply remove static flag, it needs to be there so you are going the right direction, but `@static` variable is a special workaround, it's only set when this is passed in via `?static=1` URL argument. This will work in bootdisk workflow for kickstart, but not in this context.

The only way of making this more flexible is to find out DHCP boot mode for provisioning subnet. So can you rewrite the patch and utilize this code (we already have in other template in core):

```ruby
  if subnet4 && !subnet4.dhcp_boot_mode?
    foreman_url = foreman_url('provision', static: '1')
  elsif subnet6 && !subnet6.dhcp_boot_mode?
    foreman_url = foreman_url('provision', static6: '1')
  else
    foreman_url = foreman_url('provision')
  end
```

Please do this for both Debian and Red Hat kexec templates. Also when in DHCP mode, do not send those "ip=" arguments, just do `ip=dhcp` for Red Hat. Do the same for Debian.

The only concern I have is that users with subnets set to DHCP mode will have a regression after ugprade. The best we can do is to give them the upgrade note (TM). Nobody reads them tho.

cschlue (commented):
Should the netcfg flags be included in options for dhcp? Should they be set as "dhcp"?

cschlue:
Does `Kexec append[]` require `interface`, `netcfg/get_hostname` and `BOOTIF` to be defined when using DHCP?

cschlue:
Hello, can someone review my latest commits and provide some input on the questions posted?

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -12,12 +12,15 @@ environments. The template must generate JSON format with the following items
 "kernel", "initram", "append" and "extra". The kexec command is composed in
 the following way:
 
-kexec --force --debug --append=$append --initrd=$initram $extra $kernel
+kexec --force --reset-vga --append=$append --initrd=$initram $extra $kernel
 
 Please read kexec(8) man page for more information about semantics.
-Extra options like --reset-vga can be set via "extra" array.
```
lzap, line 18: This is no longer true, we've had `--reset-vga` for few releases but we had to remove this by default because it was causing some problems on some systems. So it's really without `--reset-vga` by default, if you want it then add it to this template.

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -29,12 +32,31 @@ Extra options like --reset-vga can be set via "extra" array.
   options << @host.facts['append']
   options << "domain=#{@host.domain}"
   options << 'console-setup/ask_detect=false console-setup/layout=USA console-setup/variant=USA keyboard-configuration/layoutcode=us localechooser/translation/warn-light=true localechooser/translation/warn-severe=true'
-  options << "locale=#{host_param('lang') || 'en_US'}"
+  options << "locale=#{@host.params['lang'] || 'en_US'}"
```
lzap, line 35: Please use the `host_param` macro, this has been deprecated already. Are you on some older version?

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -29,12 +32,31 @@ Extra options like --reset-vga can be set via "extra" array.
...
+    foreman_url = foreman_url('provision')
+    options << "netcfg/disable_dhcp=false"
+  end
 -%>
 {
   "kernel": "<%= @kernel_uri %>",
   "initram": "<%= @initrd_uri %>",
-  "append": "url=<%= @static ? foreman_url('provision') + "&static=yes" : foreman_url('provision') %> interface=<%= mac %> netcfg/get_ipaddress=<%= ip %> netcfg/get_netmask=<%= mask %> netcfg/get_gateway=<%= gw %> netcfg/get_nameservers=<%= dns %> netcfg/disable_dhcp=<%= @static ? true : false %> netcfg/get_hostname=<%= @host.name %> BOOTIF=<%= bootif %> <%= options.compact.join(' ') %>",
+  "append": "url=<%= foreman_url %> interface=<%= mac %> netcfg/get_hostname=<%= @host.name %> BOOTIF=<%= bootif %> <%= options.compact.join(' ') %>",
   "extra": []
```
lzap, line 61: Yeah this looks nice.

app/views/foreman_discovery/debian_kexec.erb
```diff
@@ -29,12 +32,31 @@ Extra options like --reset-vga can be set via "extra" array.
...
+  elsif subnet6 && !subnet6.dhcp_boot_mode?
+    foreman_url = foreman_url('provision', static6: '1')
+    options << "netcfg/disable_dhcp=true"
+    options << "netcfg/get_ipaddress=#{ip}"
+    options << "netcfg/get_netmask=#{mask}"
+    options << "netcfg/get_gateway=#{gw}"
+    options << "netcfg/get_nameservers=#{dns}"
+  else
+    foreman_url = foreman_url('provision')
+    options << "netcfg/disable_dhcp=false"
```
lzap, line 54: Well I am not sure if this is correct. It looks like the installer will not recognize these options, only disable_dhcp according to docs:

https://www.debian.org/releases/jessie/amd64/ch05s03.html.en

These are preseed options:

https://www.debian.org/releases/stretch/amd64/apbs04.html.en#preseed-network

Have you tested this?

lzap:
> Should the netcfg flags be included in options for dhcp? Should they be set as "dhcp"?

Honestly, I can't tell. I know Red Hat very well. I cannot confirm this in the docs, some blogposts do recommend what you have here tho.

I can implement the same change for Red Hat and test it if you don't plan to update RH template, but please test this on Debian before we accept the patch.

> Does Kexec append[] require interface, netcfg/get_hostname and BOOTIF to be defined when using DHCP?

Kexec or append does not require anything, they just pass this to the linux kernel which is downloaded from your repository. In Red Hat case (Anaconda) it requires BOOTIF to be present which is the interface it picks up for configuration. For Debian I assume you need to pass this information as well.

For Debian I don't know how to do this honestly, there is an option in preseed to choose interface via its interface name, but that's not very good - interfaces do get often different names depending on configuration, drivers and Foreman does not actually make interface name mandatory. You can only rely on MAC address which is.

```
# netcfg will choose an interface that has link if possible. This makes it
# skip displaying a list if there is more than one interface.
d-i netcfg/choose_interface select auto

# To pick a particular interface instead:
#d-i netcfg/choose_interface select eth1
```

stejskalleos:
I'm closing the PR since the last update was on Feb 25, 2021

PR #528, opened 2021-02-19
fixes #19448 - Improving notification about new host

domitea:
Today the notification informs about one or more discovered hosts. However, this can be little confusing. So this PR improving notification to work on per host basis. That means that every discovered get own notification about discovery.

### DIFF
--- a/app/services/foreman_discovery/ui_notifications/new_host.rb
+++ b/app/services/foreman_discovery/ui_notifications/new_host.rb
@@ -8,7 +8,7 @@ module ForemanDiscovery
       private
 
       def create
-        add_notification if update_notifications.zero?
+        add_notification
       end
 
       def update_notifications
@@ -19,6 +19,7 @@ module ForemanDiscovery
         Notification.create!(
           initiator: initiator,
           audience: ::Notification::AUDIENCE_SUBJECT,
+          message: _("Host %s has been dicovered") % subject.name,
           subject: subject,
           notification_blueprint: blueprint
         )
--- a/test/unit/ui_notifications/destroy_host_test.rb
+++ b/test/unit/ui_notifications/destroy_host_test.rb
@@ -21,13 +21,6 @@ class DestroyHostNotificationTest < ActiveSupport::TestCase
       ForemanDiscovery::UINotifications::NewHost.deliver!(host)
     end
     assert_equal 1, blueprint.notifications.count
-    assert_no_difference('blueprint.notifications.count') do
-      host = FactoryBot.create(:discovered_host)
-      ForemanDiscovery::UINotifications::NewHost.deliver!(host)
-    end
-    assert_no_difference('blueprint.notifications.count') do
-      Host::Discovered.all.last.destroy
-    end
     Host::Discovered.destroy_all
     assert_equal 0, blueprint.notifications.count
   end
@@ -55,9 +48,9 @@ class DestroyHostNotificationTest < ActiveSupport::TestCase
     ForemanDiscovery::UINotifications::NewHost.deliver!(host1)
     host2 = FactoryBot.create(:discovered_host)
     ForemanDiscovery::UINotifications::NewHost.deliver!(host2)
-    assert_equal 1, blueprint.notifications.count
+    assert_equal 2, blueprint.notifications.count
     new_host = ::ForemanDiscovery::HostConverter.to_managed(host1, false, false)
     assert new_host.save!
-    assert_equal 1, blueprint.notifications.count
+    assert_equal 2, blueprint.notifications.count
   end
 end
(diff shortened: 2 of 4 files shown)


test/unit/ui_notifications/destroy_host_test.rb
```diff
@@ -55,7 +51,7 @@ class DestroyHostNotificationTest < ActiveSupport::TestCase
     ForemanDiscovery::UINotifications::NewHost.deliver!(host1)
     host2 = FactoryBot.create(:discovered_host)
     ForemanDiscovery::UINotifications::NewHost.deliver!(host2)
-    assert_equal 1, blueprint.notifications.count
+    assert_equal 2, blueprint.notifications.count
```
rabajaj0509, line 51: I haven't tested the code, but by the looks, i have 2 questions:
a) Does this indicate when I create one discovered host I get a notification, then if I create one more discovered host, I get 2 notifications, making it a total of 3 notifications for 2 hosts? (I believe I am wrong here?)
b) If I have 100 hosts discovered, then I will have 100 notifications. Is that what we want?

domitea: You should get a two notifications for two hosts. So yes, you will get a 100 notifications for 100 hosts and that is what we want, we're discussed it with @ares.

rabajaj0509: Maybe there is way to append the list of discovered hosts and add those to the message would be more feasible rather than filling up the administrators notifications, what do you think about this?

domitea: I think that would require a more improvement of existing notifications if you want to add a list to one notification. Also, I think that it would be current functionality but by different code. 

However, this notification is under "Hosts" category at notifications center. So that is similar you wrote. Maybe we can create different category for created for easy cleanup from center. This also avoid to unexpectedly mark as read/clear all different notifications that is tied to host.     

What do you think? @rabajaj0509 @ares

lzap: I think the implementation of notifications for discovery was never designed well, the misbehavior explained in the RM issue is definitely something that needs to be fixed.

If you ask me, I would expect the following as a discovery user:

* For each discovered host I would like to see a single notification.
* I would like to be able to click on "Read all notifications of the same type" to mark all the similar notifications.
* I would like to be able to click on "Disable all notifications of this type" to stop getting those notifications.

This is what I feel would make this feature useful. I don't know what notifications are capable in core, changes might be needed in order to achieve that.

domitea: Thanks @lzap for great comment, you're wrote what I had in mind when I did this PR.

The first comment is done, there is a code for one notification for one host in this PR. 
The last two comments are doable too but with little drawback. Notifications are in "hosts" category of notifications. That means you can mark all as read etc. for hosts category. It's also one negative there that you can miss some another notifications in hosts category in the flood of creation notifications. 

I know that current implementation of notifications are far from perfect. I don't want to use this PR for complete rewriting of them because it's out of scope and also requires some discussion how to do it in proper way. However I don't want to stay with this implementation of PR because I see that that could be really confusing. I think that it could be better if the notifications of created hosts has own category. You can mark them and do you like in a group way and other notifications for hosts will not be affected.

What do you think?

ares: I think moving this under a separate category is the best option for now and would resolve some of the concerns raised here. If @lzap agrees, let's modify this PR in that way.

domitea:
I've looked at the failing test and `HostgroupJSTest` is failing due to `Connection refused`. So I think these this failures are not relative to this PR. Could you @lzap take a look?

test/unit/ui_notifications/destroy_host_test.rb
lzap: Sure let's do whatever is reasonable.

app/services/foreman_discovery/ui_notifications/new_host.rb
```diff
@@ -19,6 +19,7 @@ def add_notification
         Notification.create!(
           initiator: initiator,
           audience: ::Notification::AUDIENCE_SUBJECT,
+          message: _("Host #{subject.name} has been dicovered"),
```
lzap: Do not use interpolation in i18n messages, please rewrite as noted in: https://projects.theforeman.org/projects/foreman/wiki/Translating

lzap:
Last nitpick, works file locally, tests are also green. We are good to go thanks.

domitea:
@lzap, I've updated the code with your comments. I've made separate group for notifications. The test are red due to `Connection Refused` error. So I think It's not a problem for merge.

lzap:
It works fine thanks.

PR #530, opened 2021-02-22
Fixes #31935 - notify errors for enforced discovery taxonomy

rabajaj0509:
Use:
```ruby
errors.add(:base, '') or errors.add(:conflict, '')
```

Else, the error will only be logged and there will not be any notification for the failure of the creation of discovery rules.

discoveryRuleError

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -49,11 +49,11 @@ class DiscoveryRule < ApplicationRecord
     return if hostgroup.nil?
     unless (ms = hostgroup.organizations - organizations).empty?
       names = ms.collect(&:name).to_sentence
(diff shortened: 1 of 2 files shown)


rabajaj0509:
Curent format of the error looks like: 
```ruby
@messages=
   {:organizations=>
     ["Host group organizations aaron-romane, Default Organization, jewel-piedra, and rocky-
bladen must also be associated to the discovery rule"],
    :base=>[],
    :conflict=>[]}>
```
And, in core, application.rb, we try to parse only the `:base` or `:conflict`:
```ruby
logger.error "Failed to save: #{hash[:object].errors.full_messages.join(', ')}" if hash[:object].respond_to?(:errors)
hash[:error_msg] ||= [hash[:object].errors[:base] + hash[:object].errors[:conflict].map { |e| _("Conflict - %s") % e }].flatten
```

lzap (approved):
Yeah thank you.

PR #531, opened 2021-02-23
Fixes #31945 - add taxonomy to discovery rules index table

rabajaj0509:
The PR is a part of fixing the _**priority**_ in Discovery Rules as discussed in https://github.com/theforeman/foreman_discovery/pull/450:
* [x] Drop priority uniqueness test: https://github.com/theforeman/foreman_discovery/pull/532
* [ ] Fix the order of discovery rule processing (and on index page as well): priority first, name second
* [x] Add organizations/locations to the index page: https://github.com/theforeman/foreman_discovery/pull/531
* [x] Priority suggestion must be suggested from current organization: https://github.com/theforeman/foreman_discovery/pull/533

PR #532, opened 2021-02-26
Fixes #31975 - remove uniqueness validation from discovery rules

rabajaj0509:
The PR is a part of fixing the _**priority**_ in Discovery Rules as discussed in https://github.com/theforeman/foreman_discovery/pull/450:
* [x] Drop priority uniqueness test : https://github.com/theforeman/foreman_discovery/pull/532
* [ ] Fix the order of discovery rule processing (and on index page as well): priority first, name second
* [x] Add organizations/locations to the index page: https://github.com/theforeman/foreman_discovery/pull/531
* [x] Priority suggestion must be suggested from current organization: https://github.com/theforeman/foreman_discovery/pull/533
<hr />

This PR should only be merged after the taxonomy scope is defined for the rules and priority suggestion is fixed.

### DIFF
--- a/app/models/discovery_rule.rb
+++ b/app/models/discovery_rule.rb
@@ -17,7 +17,6 @@ class DiscoveryRule < ApplicationRecord
   validates :max_count, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates :priority, :presence => true, :numericality => { :only_integer => true, :greater_than_or_equal_to => 0, :less_than => 2**31 }
   validates_lengths_from_database
-  validates_uniqueness_of :priority
   before_validation :default_int_attributes
   before_validation :enforce_taxonomy
 


rabajaj0509:
I think we can fix the order of discovery rule processing (and on index page as well: priority first, name second) in this PR itself.

db/migrate/20210226102122_remove_priority_score_index_from_discovery_rules.rb
```diff
@@ -0,0 +1,6 @@
+class RemovePriorityScoreIndexFromDiscoveryRules < ActiveRecord::Migration[6.0]
+  def change
+    remove_index :discovery_rules, :priority
+    add_index :discovery_rules, [:priority, :name]
+  end
+end
```
lzap: Let's just keep the index on priority column and not introduce a new one, just drop this migration from the patch completely.

lzap (changes requested):
Sorry for the delay, this is correct.

However we still need to have an index on priority column, we do use it in the `find_discovery_rule` method and it can be useful for the UI index page query. SQL server will take advantage of such index when searching with WHERE conditions on priority and name (as long as priority is first).

There is no need to create composed index for priority + name as there will not be many records with the same priority. Probably just few, up to dozens, that's not very useful index.

lzap (approved):
Thank you.

PR #534, opened 2021-03-18
fixes #21041 - unify usage of datetime macros

domitea:
This PR unifies usage of datetime macros from core version 1.16


rabajaj0509:
Thanks @domitea

PR #536, opened 2021-04-19
Fixes #32341 - parametrize bootif hardware type

lzap:
RHEL 8.3 no longer accepts 00- in BOOTIF kernel command line argument and therefore the correct interface is not picked. We have fixed bootdisk workflows however missed the discovery kexec workflow.

For the whole story see the BZ and the other associated BZ. But in short, we must send 01- instead of 00- by default.

Please test rendering, I am not entirely sure if `host_param` is actually available in kexec rendering context. It actualy might fail, therefore I would need to rewrite the patch and use 01 as a constant.


rabajaj0509:
@lzap looking into the BZs and will test the PR. Let me get back to you with my findings.

lzap:
Thanks get better!

PR #537, opened 2021-04-22
Fixes #18931 - Prefill architecture when creating managed host

domitea:
The creation of managed host from discovered host has a lot of fields
to fill. This tries to solve this problem by derive possible attributes
of new managed host by using known facts.

### DIFF
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -13,6 +13,8 @@ class ForemanDiscovery::HostConverter
       host.managed = set_managed
       host.primary_interface.managed = set_managed
     end
+
+    prefill_attributes_from_facts(host)
     # set build only and only on final save (facts are deleted)
     set_build_clean_facts(host) if set_build
     host
@@ -70,5 +72,14 @@ class ForemanDiscovery::HostConverter
     else
       Rails.logger.debug "Discovered host subnets #{[host.subnet, host.subnet6]} match hostgroup subnets"
     end
+  end
+
+  def self.prefill_attributes_from_facts(host)
+    facts = host.facts_hash
+
+    architecture_type = facts['architecture']
+
+    architecture = Architecture.find_by(name: architecture_type)
+    host.architecture = architecture
   end
 end


timogoebel:
Is this a functionality that should be implemented via the import hook abstraction we have? Just wondering.

lzap (commented):
You cannot do that, the method `to_managed` is called from several different context. The one that you are interested in is when creating a host for the UI form and in this context it is correct.

Also as Timo mentions, we actually have a mechanism called ImportHook in the codebase that was created exactly for this purpose (e.g. there is a taxonomy assignment). Note this is called when a host is actually imported, therefore your new importer will set architecture during provision phase, then the architecture will be correctly presented in the host edit form, or can be overriden by a hostgroup.

However it is also used when actually saving the host after it is submitted. Also it is used when auto-provisioning is performed. In both these context this would overwrite the architecture set by the user or hostgroup

lzap:
I agree that if you type "model" into discovered hosts index page, no results are found because model is not being set and this is confusing. The fix should be relatively small, either in finding why model is not set in core's PuppetFactParser or setting it explicitly via FactImporter in discovery. Model attribute is part of Base host, therefore there is no problem with that.

However the original issue requirement to set architecture is a bit more work because architecture is not attribute of Base host but only Managed host. Therefore we would need to move it into Base host class, fix all issues introduced with this refactoring and figure out if to move other attributes associated along with it (e.g. operating system). Since you cannot type in "architecture" into discovery host search field, I do believe this is not worth the effort.

Therefore I suggest to implement setting model however avoid setting architecture due to technical limitations and development cost. We can reconsider this if we ever get rid of STI, which we already started (host facets). @ares

domitea:
I'll make it even more confusing because I'm for closing of this PR and also the continuous one.

Since the architecture is of the table because limitations, I focused on the correct model assignation. The problem is that it's working right now. Even the `Host::Discovered` has assigned a hardware model since the fact import (we're overlooked with @lzap). We can try it in rails console and even the discovered page has column model, that can be searched right now. 
All of this means that when the discovered host is flipped to managed host, the model relation is used for proper form setup.

And that means that I would implement a feature that is already working. So I'm for closing this PRs (also https://github.com/theforeman/foreman_discovery/pull/538)

What do you think, @ares or @lzap or others?

ares:
sounds reasonable, if it works already, let's close PRs and also redmine issues as resolved

domitea:
Great! Closing.

lzap:
Thank you for investigating this, appreciated.

PR #538, opened 2021-04-22
Fixes #18931 - Prefill hardware model when creating managed host

domitea:
The creation of managed host from discovered host has a lot of fields
to fill. This tries to solve this problem by derive possible attributes
of new managed host by using known facts.

This PR continues from https://github.com/theforeman/foreman_discovery/pull/537 because it use a same method for gathering data for prefill values.

### DIFF
--- a/app/services/foreman_discovery/host_converter.rb
+++ b/app/services/foreman_discovery/host_converter.rb
@@ -13,6 +13,8 @@ class ForemanDiscovery::HostConverter
       host.managed = set_managed
       host.primary_interface.managed = set_managed
     end
+
+    prefill_attributes_from_facts(host)
     # set build only and only on final save (facts are deleted)
     set_build_clean_facts(host) if set_build
     host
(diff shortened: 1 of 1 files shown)


lzap:
The same as with the architecture, lets use the Importer framework.

lzap:
I agree that if you type "model" into discovered hosts index page, no results are found because model is not being set and this is confusing. The fix should be relatively small, either in finding why model is not set in core's PuppetFactParser or setting it explicitly via FactImporter in discovery. Model attribute is part of Base host, therefore there is no problem with that.

However the original issue requirement to set architecture is a bit more work because architecture is not attribute of Base host but only Managed host. Therefore we would need to move it into Base host class, fix all issues introduced with this refactoring and figure out if to move other attributes associated along with it (e.g. operating system). Since you cannot type in "architecture" into discovery host search field, I do believe this is not worth the effort.

Therefore I suggest to implement setting model however avoid setting architecture due to technical limitations and development cost. We can reconsider this if we ever get rid of STI, which we already started (host facets). @ares

domitea:
Closing PR according to discussion in https://github.com/theforeman/foreman_discovery/pull/537

lzap:
Thank you for investigating this.

PR #539, opened 2021-04-22
Fixes #14595 - Removing macros that are in core

domitea:
Removing allowance of macros and variables because they are now live in
core.


lzap:
TY merged.

PR #541, opened 2021-06-15
Fixes #32812 - try IPv6 for reboot/kexec

lzap:
Introducing also a new setting so users can override if they prefer IPv6 or IPv4.

### DIFF
--- a/app/models/host/discovered.rb
+++ b/app/models/host/discovered.rb
@@ -133,26 +133,39 @@ class Host::Discovered < ::Host::Base
     subnet.present? && subnet.discovery.present?
   end
 
+  def ip4or6
+    if Setting[:discovery_prefer_ipv6]
+      IPAddr.new(self.ip6 || self.ip)
+    else
+      IPAddr.new(self.ip || self.ip6)
+    end
+  end
+
   def proxy_url(node_ip)
-    proxied? ? subnet.discovery.url + "/discovery/#{node_ip}" : "https://#{node_ip}:8443"
+    wrapped_ip = node_ip.ipv6? ? "[#{node_ip}]" : node_ip
+    proxied? ? subnet.discovery.url + "/discovery/#{node_ip}" : "https://#{wrapped_ip}:8443"
   end
 
   def refresh_facts
-    facts = ::ForemanDiscovery::NodeAPI::Inventory.new(:url => proxy_url(self.ip)).facter
+    facts = ::ForemanDiscovery::NodeAPI::Inventory.new(:url => proxy_url(ip4or6)).facter
     self.class.import_host facts
     ::ForemanDiscovery::HostFactImporter.new(self).import_facts facts
   rescue => e
     ::Foreman::Logging.exception("Unable to get facts from proxy", e)
-    raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url(self.ip), :error => e)
+    raise ::Foreman::WrappedException.new(e, N_("Could not get facts from proxy %{url}: %{error}"), :url => proxy_url(ip4or6), :error => e)
   end
 
-  def reboot(old_ip = nil, new_ip = nil)
+  def reboot(old_ip = nil, new_ip = nil, old_ip6 = nil, new_ip6 = nil)
     # perform the action against the original lease as well as the new reservation
-    ips = [old_ip, new_ip, self.ip].compact.uniq
+    if Setting[:discovery_prefer_ipv6]
+      ips = [old_ip6, new_ip6, self.ip6, old_ip, new_ip, self.ip].compact.uniq
+    else
+      ips = [old_ip, new_ip, self.ip, old_ip6, new_ip6, self.ip6].compact.uniq
+    end
     logger.debug "Performing reboot calls against #{ips.to_sentence}, facts left #{facts.count}"
     ips.each do |next_ip|
       begin
-        node_url = proxy_url(next_ip)
+        node_url = proxy_url(IPAddr.new(next_ip))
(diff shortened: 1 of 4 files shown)


app/models/host/discovered.rb
```diff
@@ -133,22 +133,35 @@ def proxied?
...
+    if Setting[:discovery_prefer_ipv6]
+      self.ip6 || self.ip
+    else
+      self.ip || self.ip6
+    end
+  end
+
   def proxy_url(node_ip)
-    proxied? ? subnet.discovery.url + "/discovery/#{node_ip}" : "https://#{node_ip}:8443"
+    wrapped_ip = node_ip =~ Resolv::IPv6::Regex ? "[#{node_ip}]" : node_ip
```
timogoebel: Why don't you return a IPAddr object in `ip4or6` in the hope that this is not needed anymore.

lzap: Just like that? (Amended)

lzap:
Do not merge, we are still testing this.

lzap:
That is weird, I do not understand why this isn't working. If you have a clue @timogoebel maybe different view would help I am stuck with this:

```
irb(main):016:0> resource = ::ForemanDiscovery::NodeAPI::Power.service(:url => "https://[2a00:4bc0:1010:1102:0:b19:7175:babe]:8443")
=> #<ForemanDiscovery::NodeAPI::PowerService:0x0000000012d5f0b8 @args={:url=>"https://[2a00:4bc0:1010:1102:0:b19:7175:babe]:8443"}, @connect_params={:headers=>{:accept=>:json}, :timeout=>20, :verify_ssl=>0}>
irb(main):017:0> resource.reboot
Traceback (most recent call last):
        5: from lib/tasks/console.rake:5:in `block in <top (required)>'
        4: from (irb):17
        3: from foreman_discovery (16.3.1) app/services/foreman_discovery/node_api/power_service.rb:8:in `reboot'
        2: from foreman_discovery (16.3.1) app/services/foreman_discovery/node_api/node_resource.rb:103:in `put'
        1: from lib/foreman/http_proxy/net_http_extension.rb:5:in `request'
URI::InvalidURIError (bad URI(is not URI?): 2a00:4bc0:1010:1102:0:b19:7175:babe)
```

Why it reports just the IPv6 as a bad URI? Indeed that is not how an URI should look like, but the code passes this into request correctly I think.

lzap:
Looks like a bug in lib/foreman/http_proxy/net_http_extension.rb!

lzap:
Patch for core: https://github.com/theforeman/foreman/pull/8606

app/models/host/discovered.rb
timogoebel: Yep!

app/models/host/discovered.rb
```diff
@@ -133,26 +133,39 @@ def proxied?
...
+    if Setting[:discovery_prefer_ipv6]
+      IPAddr.new(self.ip6 || self.ip)
+    else
+      IPAddr.new(self.ip || self.ip6)
+    end
+  end
+
   def proxy_url(node_ip)
-    proxied? ? subnet.discovery.url + "/discovery/#{node_ip}" : "https://#{node_ip}:8443"
+    wrapped_ip = node_ip.ipv6? ? "[#{node_ip}]" : node_ip
```
timogoebel, line 145: While looking at this again, I believe there must be a better way (my idea was `IPAddr`) to wrap the IP ourselves.

lzap:
And this was confirmed to work.

app/models/host/discovered.rb
lzap: So you don't like the if statement and want to figure out something better? Problem is, URL and IPAddr do not play nice together. There is no IPAddr.hostname field that would return either wrapped or unwrapped version.

```
>> URI::HTTP.build(host: IPAddr.new("1.2.3.4"))
Traceback (most recent call last):
       12: from /home/lzap/.rbenv/versions/2.7.2/bin/irb:23:in `<main>'
       11: from /home/lzap/.rbenv/versions/2.7.2/bin/irb:23:in `load'
       10: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        9: from (irb):6
        8: from (irb):7:in `rescue in irb_binding'
        7: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/http.rb:62:in `build'
        6: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:137:in `build'
        5: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:137:in `new'
        4: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:191:in `initialize'
        3: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:663:in `hostname='
        2: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:663:in `!~'
        1: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/uri/generic.rb:663:in `=~'
TypeError (no implicit conversion of IPAddr into String)
```

timogoebel: Ok, I guess this was just me being nit-picky.

lzap: You are right and it is ugly code, well, I did my homework and research this. There is nothing in both IPAdd and URI classes (common, generic, HTTP(s)) that would make this easier. :-(

lzap:
This is ready for merge.

timogoebel (approved):

PR #542, opened 2021-06-15
fix missing delimiter dash in the resulting mac address string

rplevka:
this should fix the resulting string, which would otherwise end up in a format `01XX-XX-XX...`
Thanks @ezr-ondrej for extra pair of eyes.


ezr-ondrej:
ok to test

ezr-ondrej (approved):
LGTM, @lzap care to take a look?

lzap:
Ah sorry I missed this one and filed https://github.com/theforeman/foreman_discovery/pull/543 but since my tests are passing and these not (probably a random error) and I have a better commit message, let's merge mine if you don't mind?

Thanks for the report, analysis and fix Roman.

PR #543, opened 2021-06-16
Refs #32341 - added missing bootif dash

lzap:
I'm seeing the following in the kernel args during kexec-ing:

ksdevice=bootif BOOTIF=0156-6f-0d-72-0a-9d

Is the missing delimmiting dash between first two bytes intended?

My host is unable to fetch the kickstart from the server so I'm trying to understand whether this is the issue or it's just my env being configured incorrectly.

https://bugzilla.redhat.com/show_bug.cgi?id=1950929


timogoebel:
This looks like a dup of #542?!

lzap:
Thanks I closed the other one since tests are green here and not there.

timogoebel:
Thanks!

PR #546, opened 2021-09-13
Changed mac based naming to utilize discovery_hostname_fact_array

jmazzie-micron:
We wanted to be able to use the MAC address of our ipmi interface for MAC based naming, if the system had this interface. This modification utilizes the discovery_hostname_fact_array for MAC naming, which allows a user to specify a different interface for consistent naming if a server gets moved or the boot interface cable is switched. It defaults to bootif_mac, which was hard coded for MAC based naming before.


lzap:
Apologies for such a huge delay! Please write a test for this case and we are good. Nice catch.

ok to test

lzap (approved):
Sorry for the delays, thanks!

PR #548, opened 2021-09-21
Fixes #33531 - deep clone taxonomies

ares:


lzap:
Oh come on, I just made that patch locally :-)

I think this was copy-pasted from operating system controller which comes first when grepping. Unfortunately, it has no taxonomy. Merging!

lzap:
For the record, I made gem-only release 17.0.2 with only this patch, it is so minor that we will only push this into Satellite since there is a BZ with failed QA.

PR #549, opened 2021-09-24
Fixes #33559 - Update the call for registration of fact parser

domitea:
There is a Parser registry in core and that makes old style of registration at fact parser outdated. This PR updates the registration.

Thanks @lzap for noticing it (https://github.com/theforeman/foreman/pull/8762#issuecomment-926362672)! You're faster that me.


lzap:
JS irrelevant test failure, thanks.

PR #551, opened 2021-09-30
Fixes #33599 - Update call fact parser and remove importer

domitea:
There is a effort to unify fact importers in core (https://github.com/theforeman/foreman/pull/8781) but discovery doesn't reflect that. This PR removes Discovery fact importer because it's not needed. It handles importer in core.

Note: Please merge it after core PR (https://github.com/theforeman/foreman/pull/8781)! Otherwise It's going to break Discovery fact handling (becuase I don't register a `DiscoveryFactImporter`)


domitea:
The tests are failing because core PR is not merged.

lzap:
I dropped some comments in the core PR.

PR #553, opened 2021-10-13
Refs #33531 - fix deep cloning of rules

lzap:
I somehow tested this and missed that it actually did not create a clone but edited the original. This fixes it. I also tested it works with taxonomy, in order to change taxonomy you need to test with a hostgroup with more than one loc/org otherwise edit form will not allow you to remove taxonomy from a rule during cloning/editing (you would loose access to it I think that is the reason).


ares:
There is a reason why managers should not wrote the code :-) APJ

ares (approved):

ares:
Thanks, merging!

PR #555, opened 2021-10-21
Add Puppet params with puppet plugin

ezr-ondrej:
Needs https://github.com/theforeman/foreman_puppet/pull/211


ezr-ondrej:
@lzap could you test this? Puppet side is done in `foreman_puppet 1.0.4`

ezr-ondrej:
@lzap ping?

lzap:
There is no error, however, when I set environment it is not saved. The provisioned host has `nil` for environment.

ezr-ondrej:
I believe it works now

lzap:
Works with customization form, without and also with env override. THANK YOU!

PR #556, opened 2021-10-22
Fixes #33770 - create notification only for new hosts

lzap:
SSIA


lero (commented):
it fixes the issue, thank you!

tbrisker (approved):
Thanks @lzap !

PR #557, opened 2021-11-15
Fixes #33917 - skip invalid IP address

lzap:
Some hosts errors out with IP address exception which we are unable to reproduce. To me it looks like some bug in Ruby and IPv6. Symptoms:

error (IPAddr::InvalidAddressError): invalid address


lzap:
Helped the customer to workaround the problem. Merged.

PR #564, opened 2022-03-25
Fixes #34677,#34678 - adds settings DSL, allow empty taxonomy

lzap:
SSIA

PR #565, opened 2022-03-25
Fixes #34678 - allow emptying org and loc settings

ezr-ondrej:
UI selection did not contain an blank value which caused the UI to missbehave and also confused API show.
Adding blank values to the collection should fix the issue.


lzap:
Incorporated as a separate commit into: https://github.com/theforeman/foreman_discovery/pull/564

PR #566, opened 2022-03-25
Fixes #34678 - allow emptying org and loc settings [CP 19.0]

ezr-ondrej:
UI selection did not contain an blank value which caused the UI to missbehave and also confused API show.
Adding blank values to the collection should fix the issue.

### DIFF
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -14,8 +14,8 @@ class Setting::Discovered < ::Setting
 
   def self.default_settings
     [
-      self.set('discovery_location', N_("The default location to place discovered hosts in"), "", N_("Discovery location"), nil, { :collection => Proc.new {Hash[Location.all.map{|loc| [loc[:title], loc[:title]]}]} }),
-      self.set('discovery_organization', N_("The default organization to place discovered hosts in"), "", N_("Discovery organization"), nil, { :collection => Proc.new {Hash[Organization.all.map{|org| [org[:title], org[:title]]}]} }),
+      self.set('discovery_location', N_("The default location to place discovered hosts in"), "", N_("Discovery location"), nil, { :collection => Proc.new {Hash[[['', '']].concat Location.all.map{|loc| [loc[:title], loc[:title]]}]} }),
+      self.set('discovery_organization', N_("The default organization to place discovered hosts in"), "", N_("Discovery organization"), nil, { :collection => Proc.new {Hash[[['', '']].concat Organization.all.map{|org| [org[:title], org[:title]]}]} }),
       self.set('discovery_fact', N_("Fact name to use for primary interface detection"), "discovery_bootif", N_("Interface fact")),
       self.set('discovery_auto_bond', N_("Automatic bond interface (if another interface is detected on the same VLAN via LLDP)"), false, N_("Create bond interfaces")),
       self.set('discovery_clean_facts', N_("Clean all reported facts during provisioning (except discovery facts)"), false, N_("Clean all facts")),


lzap:
`NameError: uninitialized constant ActiveRecord::Base::EnsureNotUsedBy`

De fak?!

lzap:
Looks like some refactoring in core: https://github.com/theforeman/foreman/commit/604dc716ef06935a4947a46129bc0ec9f94de0e6

But that is supposed to be only in 3.2 not in 3.1. Maybe CI job is mis-configured, weird.

ezr-ondrej:
oh yeah, that is unrelated, the fix was released with foreman 3.2 IMHO. so unless we want to release this for Foreman 3.2, that error is fine

lzap:
Problem is, I don't feel comfortable merging stuff into stable branch without tests passing. 

Can you also fix this in a separate commit (in this PR)? I think you just need to include/extend/prepend/wetf that concern:

```
[lzap@nuc foreman_discovery]$ ag EnsureNotUsed
app/models/concerns/discovery_taxonomy_extensions.rb
5:    before_destroy EnsureNotUsedBy.new(:discovered_hosts)
```

ezr-ondrej:
The issue is that it is not backwards compatible, so tests would pass, but in the older releases it would fail.
I've tried to do it in a backward compatible manner.

lzap:
Oh I see, thanks a bunch. I am wondering if we should do the same in discovery develop too?

ezr-ondrej:
> I am wondering if we should do the same in discovery develop too?

I don't think so, unless we want to release it for older Foreman versions

lzap:
Ok I believe we have Jenkins config old so this fails. I do not expect this CP to fail in 19, settings that is a recent feature.

https://github.com/theforeman/jenkins-jobs/pull/201

Let's re-test once this is merged.

ezr-ondrej:
[test foreman_discovery]

lzap:
Hmmm "/home/jenkins/workspace/foreman_discovery-pull-request/database/postgresql/label/fast/ruby/2.7/foreman/app/services/setting_registry.rb:115:in `each': DEPRECATION WARNING: You are using a deprecated behavior, it will be removed in version 3.4, subclassing Setting is deprecated 'Setting::Discovered' should be migrated to setting DSL see https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#settings for details (called from each at /home/jenkins/workspace/foreman_discovery-pull-request/database/postgresql/label/fast/ruby/2.7/foreman/app/services/setting_registry.rb:115) (ActiveSupport::DeprecationException)"

How is this possible? This is supposed to be Foreman 3.1 and not develop.

ezr-ondrej:
Maybe https://github.com/theforeman/jenkins-jobs/pull/201 needs jenkins restart to get propagated?

lzap:
So to review, settings was deprecated in 3.2: https://projects.theforeman.org/issues/33781

That ensure constant move also in 3.2: https://projects.theforeman.org/issues/33986

So please remove the extra commit and lets retrigger tests once again, it probably takes some time for Jenkins to apply the changes.

ezr-ondrej:
Yeah, that's why the ugly if, in case we would release it with 3.2 accidently, but that should not happen, right?
Dropped the commit, hopefully jenkins got the memo what Foreman to test against already? :)

lzap:
Still on develop, try once again? Checking out Revision 7965c802862c07aae663001f3622c3081fa36ae3 (origin/develop)

lzap:
TIL we only do PR tests against develop, here is what I am going to do since I do not have a working 19-develop Foreman dev setup locally:

* I will merge
* I will pray
* just kidding, will trigger a weekly test and paste a link here

lzap (approved):
Let's go.

lzap:
https://ci.theforeman.org/job/test_plugin_foreman_discovery_19.0-stable/2/

ezr-ondrej:
:green_apple: :tada:

lzap:
Successfully registered gem: foreman_discovery (19.0.3)

PR #585, opened 2022-09-07
Replace Foreman to_bool with rails implementation

Dyrkon:
Replaced Foreman::Cast.to_bool with implementation from rails.

### DIFF
--- a/app/models/setting/discovered.rb
+++ b/app/models/setting/discovered.rb
@@ -9,7 +9,7 @@ class Setting::Discovered
   end
 
   def self.discovery_lock?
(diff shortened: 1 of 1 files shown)


ekohl:
ok to test

evgeni:
in https://github.com/theforeman/foreman_discovery/pull/589 I just went and dropped that conversion alltogether, as the setting is already defined as `:boolean` and I expect the settings layer in Foreman to do the right thing for me

stejskalleos:
Closing the PR, we are not going to implement it

PR #589, opened 2022-11-02
make hostname and fact_column arrays

evgeni:


evgeni:
@ezr-ondrej would you mind looking at this? is this the right way to migrate settings to a different "type"?

ekohl (approved):
This looks sensible to me.

ehelms:
@stejskalleos Could you double check over this if ready to merge?

stejskalleos (approved):
:green_apple: **LGTM**, works as expected

* `discovery_hostname` & `discovery_fact_column` settings are applied correctly
* Migration works well too, although having rollback option would be nice
* Tests & Rubocop are passing

stejskalleos:
Thanks @everybody!

PR #598, opened 2023-05-15
Fix tests for array settings migrations

stejskalleos:


stejskalleos:
[test foreman_discovery]

PR #604, opened 2023-06-08
Fixes #36491 - Use better fact search query

adamruzicka:
Requires:
- [ ] https://github.com/theforeman/foreman/pull/9724


lero:
hi, any news on when this will be merged? thank you!

PR #605, opened 2023-06-20
don't depend on theforeman/stories

evgeni:
this was dropped from core in https://github.com/theforeman/foreman/commit/5e7343ecdad7632a94efa26fcba84889e66b06b4


evgeni:
:eyes: @Ron-Lavi @stejskalleos

stejskalleos (approved):
:green_apple: **LGTM**

PR #606, opened 2023-07-17
Fix extra/discover-host

stejskalleos:
How to test:
```
cd extra
./discover-host --url foreman.example.com
```
Go to hosts > Discovered hosts and see sample host there


nofaralfasi (approved):
Fixed the issue. Thanks @stejskalleos!

PR #607, opened 2023-09-19
Rubocop

stejskalleos:


stejskalleos:
@nofaralfasi review please

Gemfile
```diff
@@ -2,3 +2,4 @@ source "http://rubygems.org"
 
 gemspec
 
+gem 'theforeman-rubocop', '~> 0.1'
```
ekohl: semver says all .y versions in the 0.y series can be breaking so they suggest to pin to:
```suggestion
gem 'theforeman-rubocop', '~> 0.1.0'
```

.rubocop.yml
ekohl, line 1: If you really want to benefit from theforeman-rubocop you should also inherit its rules: https://github.com/theforeman/theforeman-rubocop#usage

ekohl (commented):
I've asked @archanaserver to look at updating theforeman-rubocop to a more recent version. Enabling this in CI is a good first step.

stejskalleos:
Ruby tests are changing probably due to some changes in the `foreman:develop`, need to investigate further. But rubocop should be set correctly now

ekohl (approved):
What's your plan with the TODO file?

stejskalleos:
> What's your plan with the TODO file?

1. Fix failing Ruby tests
2. Fix failing foreman_discovery job
3. Apply safe autocorrect suggestions
4. Apply the rest of the autocorrect suggestions
5. ...
6. profit

PR #609, opened 2023-11-13
Fixes BZ#2238952 - Selecting host group change form URL

stejskalleos:
Fixed issue:
* Go to discovered hosts
* Select provision > Customize
* Select hostgroup
* Create host

### DIFF
--- a/app/helpers/discovered_hosts_helper.rb
+++ b/app/helpers/discovered_hosts_helper.rb
@@ -94,8 +94,15 @@ module DiscoveredHostsHelper
   end
 
   def host_path(host)
-    return super unless controller_name == 'discovered_hosts'
-    discovered_host_path(host)
+    return discovered_host_path(host) if controller_name == 'discovered_hosts'
+    # This is a hack to fix the hostgroup selection on the discovered host edit page
+    # The hostgroup onChange action is replacing the form url from `/discovered_hosts/:id`
+    # to `/hosts/:id``, which is not correct (in this case) and it breaks the form submission
+    return discovered_host_path(host) if controller_name == 'hosts' &&
+                                         action_name == 'process_hostgroup' &&
+                                         Host::Discovered.find_by(id: host[:id])
+
+    super
   end
 
   def discovery_doc_version


ShimShtein (commented):
While this could work, I think we should try and fix the root cause to f this problem. I think it comes form the fact that `hosts_path` override does not take into account hostgroup replacement use case.

ShimShtein (changes requested):
Looks quite good, kudos for documentation! a couple of readability comments:

I would reverse the conditions of the method though:
```
return discovered_host if controller == 'discovered_hosts'
return discovered_host if controller == 'hosts' && action == 'hostgroup' && Host::Discovered.exists?(id)
super
```
I think it's a bit more readable.
Also I would use `exists?` instead of `find_by` , it's much less memory consuming:
https://api.rubyonrails.org/v6.1.3/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F

app/helpers/discovered_hosts_helper.rb
```diff
@@ -94,8 +94,15 @@ def provision_button(host, authorization_options)
...
 
   def host_path(host)
-    return super unless controller_name == 'discovered_hosts'
-    discovered_host_path(host)
+    return discovered_host_path(host) if controller_name == 'discovered_hosts'
+    # This is a hack to fix the hostgroup selection on the discovered host edit page
+    # The hostgroup onChange action is replacing the form url from `/discovered_hosts/:id`
+    # to `/hosts/:id``, which is not correct (in this case) and it breaks the form submission
+    return discovered_host_path(host) if controller_name == 'hosts' &&
+                                         action_name == 'process_hostgroup' &&
```
ShimShtein, line 102: While I'm OK with specifying this particular method as it's the one that causes the bug, I wonder if we shouldn't include all the AJAX methods from the [controller](https://github.com/theforeman/foreman/blob/2774da229db941d1f0b55c92ccab8935951397a4/app/controllers/hosts_controller.rb#L15)

stejskalleos: Maybe, but I'm worried about the impact it can have across the plugins. This fix is right now important for downstream release, can we go with it as it is and later on think about the refactoring? Dont want to block QAs who are waiting for this fix.

ShimShtein: Sounds good for me. Let's start with this, and deal with the fallout later.

nofaralfasi (approved):
I didn't test it because I don't have the required setup, but the changes look good to me.

ShimShtein (approved):
Didn't test it, but ACK from me.
Just make sure the tests are green before the merge!

stejskalleos:
The CI is broken, but not because of the changes in this PR, it's been red for a while.
The packit build is broken as well, I can't send the rpm to QEs for confirmation before the merge.
The safe conditions are pretty strict, I tested it, but let's get this in and see if QAs will be happy with it.

@ShimShtein @nofaralfasi

PR #611, opened 2023-11-28
i18n - pulling from tx

ekohl:


locale/Makefile
```diff
@@ -6,19 +6,19 @@
...
 POFILES = $(shell find . -name '$(DOMAIN).po')
 MOFILES = $(patsubst %.po,%.mo,$(POFILES))
 POXFILES = $(patsubst %.po,%.pox,$(POFILES))
 EDITFILES = $(patsubst %.po,%.edit.po,$(POFILES))
+JSFILES = $(shell find ../app/assets/javascripts/*/locale -name '$(DOMAIN).js')
 
 %.mo: %.po
 	mkdir -p $(shell dirname $@)/LC_MESSAGES
 	msgfmt -o $(shell dirname $@)/LC_MESSAGES/$(MOFILE) $<
-	touch $(shell dirname $@)/LC_MESSAGES/$(MOFILE)
```
ekohl, line 21: I've seen that this can be useful. msgfmt doesn't change the file (and keep the timestamp if the content is the same). That means the .po file can be newer than the .mo file.

Can be a good change to https://github.com/theforeman/foreman_plugin_template/blob/master/locale/Makefile and let other repos also benefit from it.

PR #612, opened 2024-01-12
Test on Ruby 3

ofedoren:
Please, don't merge just yet. Foreman on Ruby 3.0 is still in progress.

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -78,7 +78,7 @@ module ForemanDiscovery
               full_name: N_("Hostname facts"),
               description: N_("List of facts to use for the hostname (first wins)")
 
-            validates "discovery_hostname", presence: true
+            validates "discovery_hostname", { presence: true }
 
             setting "discovery_auto",
               type: :boolean,
@@ -98,7 +98,7 @@ module ForemanDiscovery
               full_name: N_("Hostname prefix"),
               description: N_("The default prefix to use for the host name, must start with a letter")
 
-            validates "discovery_prefix", presence: true
+            validates "discovery_prefix", { presence: true }
 
             setting "discovery_fact_column",
               type: :array,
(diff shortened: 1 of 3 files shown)


lib/foreman_discovery/engine.rb
```diff
@@ -78,7 +78,7 @@ class Engine < ::Rails::Engine
               full_name: N_("Hostname facts"),
               description: N_("List of facts to use for the hostname (first wins)")
 
-            validates "discovery_hostname", presence: true
```
ekohl, line 81: It surprises me this is needed. Can't we look at the `validates` method to see if `presence: true` can be made valid?

ofedoren: Only if we change its signature: https://github.com/theforeman/foreman/blob/develop/app/registries/foreman/setting_manager.rb#L58

I mean, currently `validates` expects 2 arguments. Due to new Ruby kwargs, they are not considered as an implicit `Hash` argument.

As a solution I'd suggest to finally use kwargs as Ruby wants or we could make second (`validations`) parameter to be optional (`validations = {}`).

ekohl: It's supposed to map to [ActiveModel::Validations::ClassMethods.validates](https://github.com/rails/rails/blob/758e6e1a22eb106f798799ddfce149b91b632f26/activemodel/lib/active_model/validations/validates.rb#L8-L128) and that uses `def validates(*attributes)`. I'm not sure that is a great API.

So I think adjusting the signature is the right thing. Just not 100% sure what it should be. Perhaps:
```ruby
def validates(name, **validations)
```

And then special case if `:proc` is in validations.

ofedoren: That could be, but since the old signature accepts `proc` as an argument (not nested under `:proc` key), we would still need to update places which use that. I think we would need to adjust plugins anyway if we change the signature.

ekohl: Perhaps `def validates(name, proc = nil, **validations)` would work?

ekohl: Let's see if https://github.com/theforeman/foreman_discovery/pull/622 works.

ekohl:
I think https://github.com/theforeman/foreman_discovery/pull/622 replaces this.

ofedoren:
Closing in favor of https://github.com/theforeman/foreman_discovery/pull/622

PR #613, opened 2024-01-16
Rewrite workflow using reusable actions

archanaserver:


.github/workflows/ruby.yml
```diff
@@ -1,77 +1,27 @@
 name: Ruby tests
+
 on:
   pull_request:
+
 env:
   BUNDLE_WITHOUT: journald:development:console:libvirt
   RAILS_ENV: test
   DATABASE_URL: postgresql://postgres:@localhost/test
   DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
```
ekohl: We don't have these in our common action. Are they still needed?

.github/workflows/ruby.yml
```diff
@@ -1,77 +1,27 @@
...
-        run: |
-          bundle exec rake db:create
-          bundle exec rake db:migrate
-          bundle exec rake db:test:prepare
-      - name: Prepare webpack
-        run: bundle exec rake webpack:compile
-      - name: Run plugin tests
-        run: |
-          bundle exec rake test:discovery -v --trace
-          bundle exec rake test TEST="test/unit/foreman/access_permissions_test.rb" -v --trace
```
ekohl, line 70: Also a case of https://github.com/theforeman/actions/pull/16. Not a blocker, but keeping track.

stejskalleos:
I see it's using Ruby 2.7 and Node 14, I was expecting Ruby 3.0 there as well. Or it is planned for later?

ekohl:
> I see it's using Ruby 2.7 and Node 14, I was expecting Ruby 3.0 there as well. Or it is planned for later?

Planned for later. It reads `matrix.json` from the target version (defaults to develop, so https://github.com/theforeman/foreman/blob/develop/.github/matrix.json). Once we modify that to include 3.0, that will be tested.

You can take this as a base, similar to https://github.com/theforeman/foreman_remote_execution/pull/864

.github/workflows/ruby.yml
archanaserver: No, we don't have to. I just looked into the foreman plugin test.

.github/workflows/ruby.yml
```diff
@@ -1,77 +1,21 @@
...
+
 concurrency:
   group: ${{ github.head_ref }}
   cancel-in-progress: true
+
 jobs:
   rubocop:
     uses: theforeman/actions/.github/workflows/rubocop.yml@v0
     with:
       command: bundle exec rubocop --parallel --format github
```
stejskalleos, line 14: Is the `command` needed? Checking [the Readme](https://github.com/stejskalleos/foreman_actions/tree/v0#rubocop), it should work without it.

ekohl: The difference is that one uses the Rake task, but this has 2 benefits: parallel execution (faster) and it adds any failures as annotations in GitHub. You can add the formatter in the rake task (autodetecting the GHA environment), but there's no way for the parallel part to be there.

So you can argue about benefits in either direction. As the maintainer of the repository I'll put it with you.

For what it's worth, I've been considering moving the GHA to this exact command as a default, without detecting a Rakefile. It's been a long time since rubocop has introduced the standalone command and by now we should probably just force any repositories to update.

ekohl:
Packit is failing and https://github.com/theforeman/foreman_discovery/pull/615 should resolve that.

ekohl (approved):

PR #618, opened 2024-01-23
Fix menu item integration test

ekohl:
Since Foreman 2.0 there has been a UserMenu service that provides the entire user menu. At some point the UI also changed to retrieve this dynamically rather than include it in the actual HTML. This rewrites the test to use the service, rather than testing the base template indirectly.


ekohl:
And I also added a commit that I think fixes the Jenkins tests, just because I want to see everything green now.

stejskalleos (approved):
:green_apple: Awesome, the CI is green again!
Thanks @ekohl !

PR #620, opened 2024-02-06
Remove deprecated webpacked_plugins* methods

stejskalleos:


lib/foreman_discovery/engine.rb
```diff
@@ -78,7 +78,7 @@ class Engine < ::Rails::Engine
               full_name: N_("Hostname facts"),
               description: N_("List of facts to use for the hostname (first wins)")
 
-            validates "discovery_hostname", presence: true
+            # validates "discovery_hostname", presence: true
```
ekohl, line 81: We had a discussion about this in https://github.com/theforeman/foreman_discovery/pull/612#discussion_r1450829762. For now I wouldn't change it.

ekohl (commented):
Does this break compatibility with older Foreman versions? If so, we should raise the minimum compatible version.

lib/foreman_discovery/engine.rb
ekohl: https://github.com/theforeman/foreman_discovery/pull/622 has an alternative approach.

ekohl:
https://github.com/theforeman/foreman_discovery/pull/622 made this redundant.

PR #624, opened 2024-02-16
require Foreman 3.10+ now that we do not use webpacked_plugins_*_for

evgeni:
Fixes: 43352a831d4641545a8f38de2f0770464bb8a2c5


stejskalleos (approved):

stejskalleos:
Thanks @evgeni

PR #632, opened 2024-02-20
Fixes #37186 - webpack 5 adjustments

MariaAga:
to ensure correct loading order we need to wrap js scripts into content for.
.input.counter_spinner doesnt exist for awhile now

### DIFF
--- a/app/views/discovered_hosts/show.html.erb
+++ b/app/views/discovered_hosts/show.html.erb
@@ -96,28 +96,30 @@
   </div>
 </div>
 
-<script type="text/javascript" charset="utf-8">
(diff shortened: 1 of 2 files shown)


ShimShtein:
LGTM, Can we mark it as ready and merge it?

MariaAga:
was on draft so I could find references for `counter_spinner` not existing but I dont have time right now

ShimShtein:
Didn't find any references to counter_spinner in Foreman. Seems safe to remove. 
Besides that, the pr works as expected

nofaralfasi (approved):

PR #634, opened 2024-03-05
Fix rails constants warning

sbernhard:
This fixes:
lib/foreman_discovery/engine.rb:302: warning: already initialized constant ForemanDiscovery::Engine::MANAGER lib/foreman_discovery/engine.rb:271: warning: previous definition of MANAGER was here

### DIFF
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -266,7 +266,7 @@ module ForemanDiscovery
           # discovered_rules
           :view_discovery_rules,
         ]
-        MANAGER = READER + [
+        manager_permissions = READER + [
           # core permissions
           :create_hosts,
           :edit_hosts,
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I don't think this is correct and the actual fix is the `app.reloader.to_prepare` changes in https://github.com/theforeman/foreman_discovery/commit/7bc87fbc6022e25e09b16778cc7bcc5b2e0d35a2#diff-4b654d6bf0704dbe13565638885158149d062fb4cffe86d2b57d290450f1ee9bR25. @ofedoren do you agree?

ofedoren:
> @ofedoren do you agree?

I guess? If it were me I would get rid of those constants unless they are used somewhere else. Even though REX plugin does the same thing and it already has `app.reloader.to_prepare`, so it works, I'd remove them. There is no much point to use constants here...

ekohl:
Oh, good point. They really are "constants" because they're not frozen.

PR #636, opened 2024-03-07
Fixes #34845 - Support non-RSA private keys

ekohl:
OpenSSL can figure out which key is used and this allows supporting non-RSA keys, like elliptic curve.

I didn't test this, just trusting the instructions in https://projects.theforeman.org/issues/34842.


stejskalleos (approved):
:green_apple: LGTM, should be working, [according to docs ](https://ruby.github.io/openssl/OpenSSL.html#module-OpenSSL-label-Loading+a+Key)

PR #640, opened 2024-08-19
i18n - pulling from tx

stejskalleos:


ekohl (approved):
Technically it doesn't change any strings, but it doesn't hurt to update the line references in case translators want to see where it happens.

PR #644, opened 2024-11-08
Fixes #37997 - Support Rails 7.0

ofedoren:


ekohl (approved):
This feels compatible with Rails 6.1, so let's get it merged without it?

ofedoren:
Removed the commit for testing with opened PR, ready for merge.

test/integration/discovered_hosts_test.rb
```diff
@@ -42,7 +42,9 @@ class DiscoveredHostsTest < IntegrationTestWithJavascript
       page.find_link('Reboot').click
       wait_for_ajax
       assert page.has_text?('The following hosts are about to be changed')
-      page.find_button('Submit').click
+      wait_for do
```
stejskalleos, line 45: What's the reason to put this particular click action into the `wait_for` block?
Other tests have it as well; should we update them as well?

ofedoren: For some reason some integration tests fail either on timeouts or (like in this case) the app seems to not respond 
(with either data or page) in a certain time. Not sure if it's frontend or backend related, but waiting for a bit more helps :/

stejskalleos: Oh yes, the daily joys of maintaining integration tests.
If it helps, +1 from me.

stejskalleos (approved):

PR #650, opened 2025-01-14
IPv6 in API & tests

stejskalleos:


app/controllers/api/v2/discovered_hosts_controller.rb
```diff
@@ -62,6 +63,7 @@ def create
         param :medium_id, String, :desc => N_("required if not imaged based provisioning and host is managed and value is not inherited from host group")
         param :ptable_id, :number, :desc => N_("required if host is managed and custom partition has not been defined")
         param :subnet_id, :number, :desc => N_("required if host is managed and value is not inherited from host group")
+        param :subnet6_id, :number, :desc => N_("required if host is managed, does not have IPv4 IP / Subnet, or the value is not inherited from the host group")
```
nofaralfasi, line 66: What does it mean ` does not have IPv4 IP / Subnet`?

stejskalleos: The host must pass one of these conditions:

* Have IPv4 address or IPv4 subnet
* Have IPv6 address or IPv6 subnet
* Have IPv4 & IPv6 addresses
* Have IPv4 & IPv6 subnets

Calling for @Lennonka for text review / suggestions.

Lennonka:
```suggestion
        param :subnet6_id, :number, :desc => N_("required if host is managed, has neither IPv4 IP nor IPv4 subnet, or the value is not inherited from the host group")
```
Hopefully this is clear :)

stejskalleos:
... and it's :green_apple:

test/functional/api/v2/discovered_hosts_controller_test.rb
```diff
@@ -91,6 +107,24 @@ def test_provision_host
...
+        id: host.id,
+        discovered_host: {
+          hostgroup_id: hostgroup.id,
+          build: true
+        }
+      }
+
+    assert_response :success
+    actual = Host.unscoped.find(host.id)
+    assert actual.build?
```
nofaralfasi, line 125: Shouldn't we check whether the IPv6 fields were set correctly here?

nofaralfasi (approved):

PR #646, opened 2025-01-09
Fixes #SAT-29266 - Add ipv6 field to the discovered hosts UI

stejskalleos:

PR #652, opened 2025-01-16
Support subnet6 facts in FactToCategoryResolver

stejskalleos:
Follow up to https://github.com/theforeman/foreman_discovery/pull/646

### DIFF
--- a/app/controllers/discovered_hosts_controller.rb
+++ b/app/controllers/discovered_hosts_controller.rb
@@ -335,13 +335,6 @@ class DiscoveredHostsController < ::ApplicationController
     Bullet.enable = true if defined? Bullet
   end
 
-  def add_custom_facts
-    unless @host.primary_interface.subnet.nil?
-      discovery_subnet = "#{@host.primary_interface.subnet.name} (#{@host.primary_interface.subnet.network})"
-      assign_fact_to_category("discovery_subnet", discovery_subnet)
-    end
-  end
-
   def process_warning(hash = {})
     warning hash[:warning_msg]
   end
--- a/app/services/foreman_discovery/fact_to_category_resolver.rb
+++ b/app/services/foreman_discovery/fact_to_category_resolver.rb
@@ -51,10 +51,7 @@ module ForemanDiscovery
         assign_fact_to_category(key, value)
       end
 
-      return if host.primary_interface.subnet.nil?
-
-      discovery_subnet = "#{host.primary_interface.subnet.name} (#{host.primary_interface.subnet.network})"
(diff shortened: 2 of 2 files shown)


app/services/foreman_discovery/fact_to_category_resolver.rb
```diff
@@ -50,10 +50,7 @@ def assign_facts(host)
         assign_fact_to_category(key, value)
       end
 
-      return if host.primary_interface.subnet.nil?
-
-      discovery_subnet = "#{host.primary_interface.subnet.name} (#{host.primary_interface.subnet.network})"
-      assign_fact_to_category("discovery_subnet", discovery_subnet)
+      subnets_facts(host.primary_interface)
```
stejskalleos, line 54: To test this, you need to first create the subnet in the Foreman, and then you create a discovery host with:
```
# From foreman_discovery dir:
./extra/discover-host --json test/facts/only-ipv6.json -x -p eth0
```

app/controllers/discovered_hosts_controller.rb
```diff
@@ -335,13 +335,6 @@ def skip_bullet
     Bullet.enable = true if defined? Bullet
   end
 
-  def add_custom_facts
```
stejskalleos, line 338: I haven't found any usage of this, so I removed it.

nofaralfasi:
> Follow up to #646

Do we need to apply the changes from #646 to test this PR?

app/services/foreman_discovery/fact_to_category_resolver.rb
nofaralfasi: Is a simple IPv6 subnet sufficient, or do we need to configure specific fields?

nofaralfasi (commented):
For the remaining discovered hosts (generated using the `./extra/discover-host` command, without passing any arguments), navigating to their details page results in the following error:  
`NoMethodError (undefined method 'network' for nil:NilClass)`.

stejskalleos:
@nofaralfasi Should be fixed now.

app/services/foreman_discovery/fact_to_category_resolver.rb
nofaralfasi: @stejskalleos

stejskalleos: Yes, you need to create a simple subnet6 network before discovering the host.

stejskalleos:
... and it's :green_apple:

nofaralfasi (approved):

nofaralfasi:
Thanks @stejskalleos :rocket:

PR #657, opened 2025-02-19
i18n - pulling from tx

ekohl:


ekohl:
Looks like some strings are missing because the `.pot` file wasn't updated. This will fail the 100% translated check in downstream. This PR at least updates the `.pot` file so we can start another round of translations.

stejskalleos:
so should we merge it as it is and pull fix with the next PR?

ekohl:
If you merge this, the `.pot` file is correct and we can actually start the next round of translations. So please merge it, but don't feel pressured to make sure it's released right now.

stejskalleos (approved):

PR #659, opened 2025-04-01
Bump version to 26.0.0

nofaralfasi:


evgeni (approved):

nofaralfasi:
Any idea why we're getting this error and whether it's something we need to resolve on our side?
`npm error 403 403 Forbidden - GET https://registry.npmjs.org/decamelize/-/decamelize-1.2.0.tgz`

PR #661, opened 2025-07-28
Fixes #38620 - Validate permission filters for view_discovered_hosts

stejskalleos:
Users can assign a search filter to a discovered host that is valid only for the Host model, but not for the Host::Discovered model. For example, the 'hostgroup' field is valid for Host, but not for Host::Discovered.

### DIFF
--- a/app/lib/discovery_filter_validator.rb
+++ b/app/lib/discovery_filter_validator.rb
@@ -1,0 +1,19 @@
+class DiscoveryFilterValidator < ActiveModel::Validator
+  # Users can assign a search filter to a discovered host that is valid
+  # only for the Host model, but not for the Host::Discovered model.
+  # For example, the 'hostgroup' field is valid for Host, but not for Host::Discovered.
+  def validate(record)
+    return if record.search.blank?
+
+    has_permission = record.permissions.find { |perm| perm.name == 'view_discovered_hosts' }
+
+    return unless has_permission
+
+    begin
+      Host::Discovered.search_for(record.search)
+    rescue ScopedSearch::QueryNotSupported => _e
+      msg = N_("Search filter is invalid: Field 'search' is not recognized for Discovered hosts.")
+      record.errors.add(:search, msg)
+    end
+  end
+end
--- a/app/models/concerns/discovery_filter_extensions.rb
+++ b/app/models/concerns/discovery_filter_extensions.rb
@@ -1,0 +1,7 @@
+module DiscoveryFilterExtensions
+  extend ActiveSupport::Concern
+
+  included do
+    validates_with DiscoveryFilterValidator
+  end
+end
--- a/lib/foreman_discovery/engine.rb
+++ b/lib/foreman_discovery/engine.rb
@@ -351,6 +351,7 @@ module ForemanDiscovery
       ::Host::Managed.send :include, Host::ManagedExtensions
       ::Hostgroup.send :include, HostgroupExtensions
       ::Nic::Managed.send :include, Nic::ManagedExtensions
+      ::Filter.send :include, DiscoveryFilterExtensions
 
       # Controller extensions
       ::HostsController.send :include, ForemanDiscovery::Concerns::HostsControllerExtensions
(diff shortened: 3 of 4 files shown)


app/lib/discovery_filter_validator.rb
```diff
@@ -0,0 +1,19 @@
...
+    return if record.search.blank?
+
+    has_permission = record.permissions.find { |perm| perm.name == 'view_discovered_hosts' }
+
+    return unless has_permission
+
+    begin
+      Host::Discovered.search_for(record.search)
+    rescue ScopedSearch::QueryNotSupported => _e
+      msg = N_("Search filter is invalid: Field 'search' is not recognized for Discovered hosts.")
```
stejskalleos, line 15: @Lennonka text review please. Looking at it, the message right now doesn't make sense, the field `search` is recognized for Discovered hosts, but its value is wrong. So yeah, this needs to be updated.

Lennonka: Can we tell the user specifically what in the value is invalid?

Lennonka: For example:
```suggestion
      msg = N_("Search query is invalid: 'hostgroup' is not recognized in search for discovered hosts.")
```
Not sure if I understand it correctly.

stejskalleos:
@ShimShtein 

I've been investigating how to fix the issue in the Foreman core, rather than in the plugin.

The root of the problem lies here:
```ruby
# app/controllers/application_controller.rb
rescue_from ScopedSearch::QueryNotSupported, :with => :invalid_search_query
```
and in
```ruby
# app/controllers/concerns/foreman/controller/auto_complete_search.rb
def invalid_search_query(e)
  error (_("Invalid search query: %s") % e)
  redirect_back(fallback_location: public_send("#{controller_name}_path"))
end
```

## Possible fixes

**Show error and continue with render**
```ruby
def invalid_search_query(e)
  error (_("Invalid search query: %s") % e)
  render action: action_name, status: :bad_request
end
```

This works, but the problem is that the views for the controller actions don't account for the possibility that some variables might be nil. This would require massive rework across the whole Foreman ecosystem.
So pass.

**Render error**
```ruby
def invalid_search_query(e)
  render "common/422", :status => :bad_request
end
```
This works, but there is a problem of missing layout; it's an empty blank page with the error message. It improves the user workflow, but I couldn't get it to work with the layouts.

**Disabling the redirect**
```
# rescue_from ActiveRecord::RecordNotFound, :with => :not_found
```
In production, you'll get an error without any explanation, which won't work.

And I ran out of ideas. WDYT?

ShimShtein:
Opened a discussion here: https://community.theforeman.org/t/invalid-filters-in-user-permissons/44050
Specifically for our case, we should return proper `DiscoveredHost` type from https://github.com/theforeman/foreman/blob/ec892303ea58ffa7619f817e090ef365458b5cd7/app/controllers/permissions_controller.rb#L5 to make sure it is selected properly as a resource type in the form.

ShimShtein:
Created: https://community.theforeman.org/t/invalid-filters-in-user-permissons/44050

Here we should solve the second part: we need to make sure https://github.com/theforeman/foreman/blob/ec892303ea58ffa7619f817e090ef365458b5cd7/app/controllers/permissions_controller.rb#L5 returns the `DiscoveredHost` class, so the scoped search and everything will validate against that model.

ShimShtein:
@stejskalleos did you try to implement the discovered host type as the base object?

stejskalleos:
> @stejskalleos did you try to implement the discovered host type as the base object?

I don't know if I have the strength and will to make such a refactoring.
IMHO, the priorities are somewhere else right now.

Thinking about this PR with a fresh Monday mind, the [changes in the core](https://github.com/theforeman/foreman/commit/b49b89ad425c09ee35a9665ceff337ad3fcb96ad) are enough and this PR can be closed.

ShimShtein:
This should be relatively straight forward and will help users to use only the fields that are relevant to discovered hosts. On the other hand I won't argue with other priorities.

PR #663, opened 2025-09-03
Set up trusted publishers workflow

stejskalleos:


evgeni (approved):
Will you do the setup on the rubygems.org side, or shall I?

stejskalleos:
> Will you do the setup on the rubygems.org side, or shall I?

I don't know what needs to be done there.

evgeni:
https://guides.rubygems.org/trusted-publishing/adding-a-publisher/

But I'll do it, no worries

PR #668, opened 2025-09-24
Apply host group attrs after the host is set to managed

stejskalleos:


test/functional/discovered_hosts_controller_test.rb
```diff
@@ -122,18 +122,6 @@ def test_edit_form_submit_parameters
...
-    end
-    assert_select '#host_operatingsystem_id [selected]' do |e|
-      assert_equal hostgroup.operatingsystem_id.to_s, e.first[:value]
-    end
-    assert_select '#host_medium_id [selected]' do |e|
-      assert_equal hostgroup.medium_id.to_s, e.first[:value]
-    end
-    assert_select '#host_ptable_id [selected]' do |e|
-      assert_equal hostgroup.ptable_id.to_s, e.first[:value]
-    end
```
evgeni, line 136: does that mean we can't test these as functional tests, only as integration? because of the new UI?

stejskalleos: Yes, without JS you won't trigger exactly this method: https://github.com/theforeman/foreman/pull/10699/files

However, fortunately, it's not a significant loss, as it's already tested in the test/integration/discovered_hosts_test.rb

evgeni: Given what we see in #669, I tend to say this is not the right approach and we should further investiage a real fix.

stejskalleos: yes, agree.

evgeni (approved):

PR #669, opened 2025-09-25
try older foreman for CI

evgeni:


evgeni:
```
f356ff61487d13a6707145093e2ed3cc0fa14e6f is the first bad commit
commit f356ff61487d13a6707145093e2ed3cc0fa14e6f (HEAD)
Author: Bernhard Suttner <suttner@atix.de>
Date:   Tue Jul 16 18:10:19 2024 +0200

    Fixes #37602 - Fix broken host/hostgroup media association
    
    Hostgroup with assigned Installation Media is used.
    A host will be registered using HostRegistartion. During this process, the
    Hostgroup was used and is assigned to the host. The host is using the
    subscription-manager to register. This will create a new OS in foreman if
    the OS doesn't currently exist. This new OS doesn’t have the installation
    media assigned (because its a new OS).
    
    Results:
    
    - The new created Operatingsystem like AlmaLinux9 doesn't have
      a Installation Media.
    - The used Hostgroup has different Operatingsystems and is using
      a Installation Media which is valid for the Operatingsystem is used in
      this Hostgroup.
    
    The POST request during the HostRegistration process which assigns the
    Hostgroup to the Host and should render the host_init_config Template
    will fail with:
    
    The system has been registered with ID: 21312a8f-1b6e-4302-ae9e-1a64b643c6f8
    The registered system name is: myra-bergh.foreman.local
    ERROR: Validation failed: Medium must belong to host's operating system
    Notice:
    
    This commit makes sure that the installation medium and some other
    managed-host-related fields are only inherited for managed hosts.

 app/models/host/managed.rb |  9 +++++++--
 test/models/host_test.rb   | 35 ++++++++++++++++++++++++++++++-----
 2 files changed, 37 insertions(+), 7 deletions(-)
```

```diff
diff --git app/models/host/managed.rb app/models/host/managed.rb
index ed8988ba9..eadda33c2 100644
--- app/models/host/managed.rb
+++ app/models/host/managed.rb
@@ -558,10 +558,15 @@ autopart"', desc: 'to render the content of host partition table'
   end
 
   def inherited_attributes
-    inherited_attrs = %w{domain_id operatingsystem_id architecture_id compute_resource_id}
+    inherited_attrs = %w{domain_id operatingsystem_id compute_resource_id}
     inherited_attrs << "subnet_id" unless compute_provides?(:ip)
     inherited_attrs << "subnet6_id" unless compute_provides?(:ip6)
-    inherited_attrs.concat(%w{medium_id ptable_id pxe_loader}) unless image_build?
+
+    if managed?
+      inherited_attrs << "architecture_id"
+      inherited_attrs.concat(%w{medium_id ptable_id pxe_loader}) unless image_build?
+    end
+
     inherited_attrs
   end
```

@stejskalleos I bet this is related! :-)

stejskalleos:
cc @sbernhard

sbernhard:
> cc @sbernhard

Looks like this PR broke something but I don't know the details. Can you share some background information, please?

stejskalleos:
These two tests in foreman_discovery are failing because of the changes in the core.
https://github.com/theforeman/foreman_discovery/pull/668

sbernhard:
> These two tests in foreman_discovery are failing because of the changes in the core. #668

hmm, the overall question is then, if a host is unmanged, is it expected that the architecture is set / inherited?

These references looks like, architecture is not needed to be set for unmanaged host:
https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L109
https://github.com/theforeman/foreman/blob/develop/app/models/host/managed.rb#L317

evgeni:
Discovery expects that, yes. Whether core does (or should): no idea.
https://github.com/theforeman/foreman_discovery/blob/d0b58e411b82b9ef0d0d99888cff0b4d9285e3ae/app/controllers/api/v2/discovered_hosts_controller.rb#L58

@stejskalleos can/should discovery force the host to be "managed"? Or is there a reason we don't do this?

evgeni:
#668 now looks like a sensible patch, fixing the issue, so I'll close here :)

PR #672, opened 2025-10-07
Bump version to 26.1.0

stejskalleos:


lib/foreman_discovery/version.rb
```diff
@@ -1,3 +1,3 @@
 module ForemanDiscovery
-  VERSION = "26.0.2"
+  VERSION = "26.0.3"
```
evgeni: I don't think "remove grub1" is a patch? :)

stejskalleos: :thinking: Hmmm, good point.

evgeni (approved):

PR #673, opened 2025-10-07
i18n - pulling from tx

adamlazik1:


app/assets/javascripts/foreman_discovery/locale/ca/foreman_discovery.js
```diff
@@ -3,7 +3,7 @@
   "locale_data": {
     "foreman_discovery": {
       "": {
-        "Project-Id-Version": "foreman_discovery 26.0.1",
+        "Project-Id-Version": "foreman_discovery 26.0.2",
```
stejskalleos, line 6: Yesterday, we released 26.1.0 version in 05415c8

stejskalleos (changes requested):

app/assets/javascripts/foreman_discovery/locale/ca/foreman_discovery.js
stejskalleos: After a discussion, this should not be a problem; the metadata for translations is the important part.

ekohl: Technically this is more metadata for translators so it doesn't really hurt. So up to you to merge it or update it.

stejskalleos (approved):

PR #676, opened 2026-01-08
Fix 'Template render check' for discovery form

stejskalleos:
Requires: https://github.com/theforeman/foreman/pull/10813

### DIFF
--- a/app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
@@ -2,7 +2,7 @@ module ForemanDiscovery
   module Concerns
     module HostsControllerExtensions
       extend ActiveSupport::Concern
-      included do
+      prepended do
         before_action :set_discovered_params
       end
 
@@ -12,6 +12,12 @@ module ForemanDiscovery
       def set_discovered_params
         return if params[:discovered_host].nil?
         params[:host] ||= params.delete(:discovered_host)
+      end
+
+      def host_for_template_used
+        super
+
+        @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false, host_params) if @host.is_a?(Host::Discovered)
       end
     end
   end
--- a/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
+++ b/test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
@@ -18,6 +18,38 @@ module ForemanDiscovery
           assert_match(/"#{os.id}"/, response.body)
         end
       end
+
+      describe 'template_used' do
+        include FactImporterIsolation
+        allow_transactions_for_any_importer
+
+        setup do
+          @facts = {
+            "interfaces" => "lo,eth0",
+            "ipaddress" => "192.168.100.42",
+            "ipaddress_eth0" => "192.168.100.42",
+            "macaddress_eth0" => "AA:BB:CC:DD:EE:FF",
+            "discovery_bootif" => "AA:BB:CC:DD:EE:FF",
+            "physicalprocessorcount" => "42",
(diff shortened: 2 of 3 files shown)


test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
```diff
@@ -18,6 +18,39 @@ class HostsControllerTest < ActionController::TestCase
...
+            "ipaddress" => "192.168.100.42",
+            "ipaddress_eth0" => "192.168.100.42",
+            "macaddress_eth0" => "AA:BB:CC:DD:EE:FF",
+            "discovery_bootif" => "AA:BB:CC:DD:EE:FF",
+            "physicalprocessorcount" => "42",
+            "discovery_version" => "3.0.0",
+          }
+        end
+
+        test 'returns the discovered host if the host is a discovered host' do
```
evgeni: Why is this test passing without https://github.com/theforeman/foreman/pull/10813?

app/controllers/foreman_discovery/concerns/hosts_controller_extensions.rb
```diff
@@ -13,6 +13,17 @@ def set_discovered_params
...
+      def host_for_template_used
+        @host = if params[:id]
+                  temp_host = Host::Base.readonly.find(params[:id])
+
+                  return temp_host unless temp_host.is_a?(Host::Discovered)
+                  ::ForemanDiscovery::HostConverter.to_managed(temp_host, true, false, host_params)
+                else
+                  Host.new(host_params)
+                end
+      end
```
evgeni, line 21: This completely overrides `host_for_template_used` from core, right? So if we ever update it there, a user who has discovery installed won't benefit from the update, unless we also update it here. (And we'll forget).

Given this concern is prepended, we can't just call `super` here, right?
```ruby
def host_for_template_used
  @host = super
  @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false, host_params) if @host.is_a?(Host::Discovered)
end
```

The alternative would be to split `host_for_template_used` into a shallow `@host = find_host_for_template_used` and put the actual search into a dedicated `find_host_for_template_used` in core, then overrides can become
```ruby
def host_for_template_used
  @host = find_host_for_template_used
  @host = ::ForemanDiscovery::HostConverter.to_managed(@host, true, false, host_params) if @host.is_a?(Host::Discovered)
end
```

Or am I over-complicating things?

.rubocop_todo.yml
evgeni, line 1: this doesn't belong into this PR

stejskalleos (commented):
Updated, now the core method is not overwritten & test is checking the `@host` class

evgeni:
@stejskalleos can you please make the tests pass?

test/functional/foreman_discovery/concerns/hosts_controller_extensions_test.rb
```diff
@@ -18,6 +18,40 @@ class HostsControllerTest < ActionController::TestCase
...
+          }
+        end
+
+        test 'converts the discovered host to a managed host' do
+          # rubocop:disable Naming/VariableNumber
+          host = discover_host_from_facts(@facts)
+          params = {
+            id: host.id,
+            host: {
+              architecture_id: architectures(:x86_64).id,
```
evgeni:
```suggestion
              architecture_id: architectures(:x86_64).id,  # rubocop:disable Naming/VariableNumber
```

then you don't need to disable that cop on the whole test and it's more clear which is the offending line

evgeni:
Apart from the tests this works well!

evgeni:
Ah yes, of course, the tests need to run with the core change…

Makes sense.

evgeni (approved):
don't forget to drop the `foreman_version` from the GHA workflow before hitting that nice green button ;)

PR #688, opened 2026-07-08
Fixes #39508 - Update EmptyStatePattern mock from pf3 to pf5

pondrejk:
Updating the icon import in mock to PF5 and updating tests from Enzyme

PR #689, opened 2026-07-27
Bump version to 26.1.4

stejskalleos:


arvind4501 (approved):

arvind4501:
Just to clarify, https://github.com/theforeman/foreman_discovery/pull/688/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 since package.json is updated, should it be patch release or minor?

stejskalleos:
> should it be patch release or minor

I was thinking about it and found it too big for a patch and too small for a minor :D So I chose the patch.

